PR #3, opened 2019-05-14
Ported BlockRotator.java

DeadlyMC:
Converted the file back to .java from .javax and fixed compile errors


src/main/java/carpet/helpers/BlockRotator.java
```diff
@@ -0,0 +1,255 @@
+package carpet.helpers;
+
+import carpet.CarpetSettings;
+import net.minecraft.block.*;
```
gnembon: can you turn off combining of imports in your IDE?

src/main/java/carpet/helpers/BlockRotator.java
```diff
@@ -0,0 +1,255 @@
...
+import net.minecraft.block.enums.ComparatorMode;
+import net.minecraft.block.enums.SlabType;
+import net.minecraft.entity.Entity;
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.item.BlockItem;
+import net.minecraft.item.ItemPlacementContext;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.BlockRotation;
+import net.minecraft.util.Hand;
+import net.minecraft.util.math.*;
```
gnembon: Same here, Its not that crucial for library classes but its still better

src/main/java/carpet/helpers/BlockRotator.java
gnembon: nvm, I will do that myself at some point

PR #23, opened 2019-06-01
Added placementRotationFix mixins

DeadlyMC:


src/main/java/carpet/utils/IEntity.java
```diff
@@ -0,0 +1,6 @@
+package carpet.utils;
```
gnembon: Could you please put this one as carpet.fakes.EntityInterface. I am trying to avoid mixing useless interfaces for the sake of mixins and useful utils.

PR #26, opened 2019-06-05
Added movableTE mixins

DeadlyMC:
Code by @2No2Name 
Mixins by @DeadlyMC and @2No2Name

### DIFF
--- a/src/main/java/carpet/mixins/PistonBlock_movableTEMixin.java
+++ b/src/main/java/carpet/mixins/PistonBlock_movableTEMixin.java
@@ -54,7 +54,7 @@ public abstract class PistonBlock_movableTEMixin extends FacingBlock
     {
         if (!block.hasBlockEntity())
         {
-            return !true;
+            return false;
         }
         else
         {
--- a/src/main/java/carpet/mixins/PistonBlock_movableTEMixin.java
+++ b/src/main/java/carpet/mixins/PistonBlock_movableTEMixin.java
@@ -1,0 +1,123 @@
+package carpet.mixins;
+
+import carpet.CarpetSettings;
+import carpet.fakes.PistonBlockEntityInterface;
+import com.google.common.collect.Lists;
+import net.minecraft.block.*;
+import net.minecraft.block.entity.BlockEntity;
+import net.minecraft.block.piston.PistonHandler;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.Direction;
+import net.minecraft.world.World;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.Redirect;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+import java.util.List;
+import java.util.Set;
+
+@Mixin(PistonBlock.class)
+public abstract class PistonBlock_movableTEMixin extends FacingBlock
+{
+    protected PistonBlock_movableTEMixin(Settings block$Settings_1)
+    {
+        super(block$Settings_1);
+    }
+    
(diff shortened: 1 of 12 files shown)


src/main/java/carpet/mixins/PistonBlock_movableTEMixin.java
```diff
@@ -0,0 +1,123 @@
...
+                       block != Blocks.END_GATEWAY && block != Blocks.END_PORTAL && block != Blocks.MOVING_PISTON  &&
+                       block != Blocks.SPAWNER;
+    }
+    
+    @Redirect(method = "isMovable", at = @At(value = "INVOKE", target = "Lnet/minecraft/block/Block;hasBlockEntity()Z"))
+    private static boolean ifHasBlockEntity(Block block)
+    {
+        if (!block.hasBlockEntity())
+        {
+            return !true;
```
gnembon: !true? you mean false...

src/main/java/carpet/mixins/PistonBlock_movableTEMixin.java
```diff
@@ -0,0 +1,123 @@
...
+                    //hopefully this call won't have any side effects in the future, such as dropping all the BlockEntity's items
+                    //we want to place this same(!) BlockEntity object into the world later when the movement stops again
+                    world_1.removeBlockEntity(blockpos);
+                    blockEntity.markDirty();
+                }
+            }
+        }
+    }
+    
+    @Inject(method = "move", at = @At(value = "INVOKE", shift = At.Shift.BEFORE,
```
gnembon: why this is inject, not redirect. You are essentially doubling on BE's, one is vanilla and the second is this one. The vanilla setBlockEntity will be called right after right?

src/main/java/carpet/mixins/PistonBlock_movableTEMixin.java
```diff
@@ -0,0 +1,123 @@
...
+            int int_3, BlockPos blockPos_4, BlockState blockState_1)
+    {
+        BlockEntity blockEntityPiston = PistonExtensionBlock.createBlockEntityPiston((BlockState) list_2.get(int_3),
+                direction_1, boolean_1, false);
+        if (CarpetSettings.b_movableTileEntities)
+            ((PistonBlockEntityInterface) blockEntityPiston).setCarriedBlockEntity(list1_BlockEntities.get().get(int_3));
+        world_1.setBlockEntity(blockPos_4, blockEntityPiston);
+    }
+    
+    @Redirect(method = "move", at = @At(value = "INVOKE",
```
gnembon: OK, so you cancel vanilla call here. I assume that only the trick to being able to capture locals while doing a redirect, correct?

src/main/java/carpet/mixins/PistonBlock_movableTEMixin.java
```diff
@@ -0,0 +1,123 @@
...
+            target = "Lnet/minecraft/world/World;setBlockEntity(Lnet/minecraft/util/math/BlockPos;" +
+                             "Lnet/minecraft/block/entity/BlockEntity;)V",
+            ordinal = 0))
+    private void dontDoAnything(World world, BlockPos blockPos_6, BlockEntity blockEntityPiston)
+    {
+    }
+    
+    @Redirect(method = "move", at = @At(value = "INVOKE",
```
gnembon: weird flex but OK. its all because redirect doesn't allow to capture locals, right?

src/main/java/carpet/mixins/PistonBlock_movableTEMixin.java
DeadlyMC: This bit was by 2No... i did ask him the same...he said he did so for better readability...and i do agree with him

src/main/java/carpet/mixins/PistonBlock_movableTEMixin.java
DeadlyMC: You can't capture local variables with @Redirects..the very next mixin cancels the same call

src/main/java/carpet/mixins/PistonBlock_movableTEMixin.java
DeadlyMC: Yes, unfortunately

src/main/java/carpet/mixins/PistonBlock_movableTEMixin.java
DeadlyMC: Yeah

PR #30, opened 2019-06-06
Completed persistentParrots port

DeadlyMC:


src/main/java/carpet/mixins/PlayerEntity_parrotMixin.java
```diff
@@ -0,0 +1,95 @@
...
+        if (!this.world.isClient && ((parrots_will_drop && this.fallDistance > 0.5F) || this.isInWater() || (parrots_will_drop && this.hasVehicle())) || this.abilities.flying)
+        {
+            this.dropShoulderEntities();
+        }
+    }
+    
+    @Redirect(method = "damage", at = @At(value = "INVOKE",
+              target = "Lnet/minecraft/entity/player/PlayerEntity;dropShoulderEntities()V"))
+    private void cancelDropShoulderEntities2(PlayerEntity playerEntity)
+    {
```
gnembon: why not to just use one @Redirect in this case, and just redirect dropShoulderEntities.

src/main/java/carpet/mixins/PlayerEntity_parrotMixin.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+    
+    @Inject(method = "tickMovement", at = @At(value = "INVOKE", shift = At.Shift.AFTER, ordinal = 1,
+            target = "Lnet/minecraft/entity/player/PlayerEntity;updateShoulderEntity(Lnet/minecraft/nbt/CompoundTag;)V"))
+    private void onTickMovement(CallbackInfo ci)
+    {
+        boolean parrots_will_drop = !CarpetSettings.getBool("persistentParrots") || this.abilities.invulnerable;
+        if (!this.world.isClient && ((parrots_will_drop && this.fallDistance > 0.5F) || this.isInWater() || (parrots_will_drop && this.hasVehicle())) || this.abilities.flying)
+        {
+            this.dropShoulderEntities();
```
gnembon: also in this case, one redirect on dropShoulderEntities with a few extra checks should make a simpler mixin

src/main/java/carpet/mixins/PlayerEntity_parrotMixin.java
DeadlyMC: I can't make use of **float_1** with the @ Redirect

gnembon: I didn't notice that. That's correct. But the first one could be replaced with one redirect I guess.

PR #96, opened 2019-10-23
Ported TNT rules

Earthcomputer:
This PR ports the following carpet rules from 1.12:
- optimizeTNT
- tntRandomRange
- hardcodeTNTangle
- mergeTNT

### DIFF
--- a/src/main/java/carpet/settings/CarpetSettings.java
+++ b/src/main/java/carpet/settings/CarpetSettings.java
@@ -17,13 +17,7 @@ import net.minecraft.world.dimension.DimensionType;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
-import static carpet.settings.RuleCategory.BUGFIX;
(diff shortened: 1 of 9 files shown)


src/main/java/carpet/settings/CarpetSettings.java
```diff
@@ -89,9 +83,53 @@
...
+                return newValue;
+            if (!optimizedTNT)
+                return null;
+            return newValue >= 0 ? newValue : null;
+        }
+
+        @Override
+        public String description() {
+            if (!optimizedTNT)
+                return "Requires optimizedTNT to be enabled";
```
gnembon: Why preventing it from setting first and then optimizedTNT.
Honestly I would break them down into two validators, cause in the rule description they could figure out as two entries. And one validator can cancel another out.

Earthcomputer: Done

PR #136, opened 2020-02-03
Added vital apps, and updated math.sc

Ghoulboy78:
I've included detailed descriptions below


gnembon:
I would claim that these are not vital apps. block counting is the closest, but would need a proper command support and be more polished.  Item counter is just a demo and duplicates functionality of hopper counters, and chunkdisplay can be destructive if you don't leave it properly. Chunk display makes sense when carpet client if finished, and probably will be added there, because displaying it with particles could be quite laggy.

Regarding math ops - sure, but they also need to be selected - there is a bunch of ops there that are not that useful and can be done organically (like hex repr)

Ghoulboy78:
Ah well, i suppose you have a point

PR #140, opened 2020-02-08
Add a modifiable delay for portal when in survival

### LINKED ISSUE #139: Feature request: The opposite of portalCreativeDelay, which allows nether portals to insta-teleport players in survival mode.
The name could be like portalSurvivalDelay which defaults to true.

ImUrX:
Fixes #139 and well nothing more

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -545,4 +545,21 @@ public class CarpetSettings
             category = CREATIVE
     )
     public static boolean extremeBehaviours = false;
+
+    @Rule(
+            desc = "Modify the ticks for Nether portal in survival",
+            options = {"1", "40", "80"},
+            category = SURVIVAL,
+            strict = false,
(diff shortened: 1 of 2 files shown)


gnembon:
First of all - it makes sense to rethink these settings - requirement of holding an obsidian block or some other weird things feels dated and having just two integer type settings for creative delay and survival delay makes sense.

But first - please don't PR to 1.16 branch - only to master.

so to summarize:
 - PR to master
 - rewrite `portalCreativeDelay to use an integer defaulting to 1
 - add `portalSurvivalDelay` type int defaulting to 80
 - rather than redirect constant (common place for mods to be incompatible with each other), add one mixin, simple injection that cancels when either of these are not default and returns appropriate delay depending on the gamemode (using isSpectator and isCreative, that would leave you both survival like modes)

ImUrX:
Ok, will close this issue because 1.16 is different from master, btw shouldnt i use isInvulnerable like vanilla minecraft does instead of isSpectator and isCreative?

PR #142, opened 2020-02-08
make the nether portal delays better

### LINKED ISSUE #139: Feature request: The opposite of portalCreativeDelay, which allows nether portals to insta-teleport players in survival mode.

ImUrX:
Fixes #139 and overwrites the default method of getMaxNetherPortalTime to something similar of the vanilla one but instead of returning constant numbers, it returns carpet rules.

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -50,11 +50,30 @@ public class CarpetSettings
     public static boolean superSecretSetting = false;
 
     @Rule(
-            desc = "Portals won't let a creative player go through instantly",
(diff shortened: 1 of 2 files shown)


gnembon:
@inject at head: 
if portalCreativeDelay != 1 && invulnerable: setReturnValue(portalCreativeDelay); 
elif (portalSurvivalDelay != 80 && !invulnerable: setReturValue(portalSurvivalDelay);

this way if neither is set, the injection is a simple passthrough allowing potentially anybody else to do with that call what they want.

ImUrX:
Done

gnembon:
fixes #139

PR #179, opened 2020-03-19
Added scoreboard manipulation in its current state to language docs, though I'm expecting to have to modify it when gnembon adds proper support for scoreboard manipulation in v1.7

Ghoulboy78:
I said it all

### DIFF
--- a/docs/scarpet/Full.md
+++ b/docs/scarpet/Full.md
@@ -2963,6 +2963,33 @@ Feature list:
 *   `coral`: random coral structure. Require water to spawn.
 *   `sea_pickle`
 *   `boulder`: A rocky, mossy formation from a giant taiga biome. Doesn't update client properly, needs relogging.
+
+## Scoreboards
+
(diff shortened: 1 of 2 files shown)


gnembon:
Lol, is that to force me to add scoreboard API?

Ghoulboy78:
more or less
also cos people can still work with the limited tools. It's like only having two pieces of wood to make a fire, so you need to get some pig entrails to serve as the string for a bow so you can make it the old fashioned way.
HEY! They should totally add pig entrails to the game so we can make better bows, lyres, and sacrifice them to get diamonds!

Ghoulboy78:
(do you think I scared him off with that comment?)

gnembon:
this should suffice
https://github.com/gnembon/fabric-carpet/commit/ba96b1b0343cb5a44db5c5608394dd3e38ce2270

Ghoulboy78:
I LOVE U @gnembon ! U R DA BEST!

PR #206, opened 2020-04-07
Fix incorrectly formatted title in Full.md

Stypox:


gnembon:
hmm, since the Full is autogenerated using mergedoc.bat I would suggest just add a newline in the doc file that is right before it.

PR #213, opened 2020-04-12
More infomation in TNT logging

X-Green:
With gametime and order-in-gametick of explosions shown

### DIFF
--- a/src/main/java/carpet/logging/logHelpers/TNTLogHelper.java
+++ b/src/main/java/carpet/logging/logHelpers/TNTLogHelper.java
@@ -9,6 +9,8 @@ public class TNTLogHelper
 {
     public boolean initialized;
     private double primedX, primedY, primedZ;
+    private static long lastGametime = 0;
+    private static int tntCount = 0;
     private Vec3d primedAngle;
     /**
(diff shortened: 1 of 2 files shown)


gnembon:
That's a good idea to add tick time. Isn't the order information little redundant? (messages will be showing up in order anyways from 0 to <whatever>

X-Green:
hmmm...
Maybe orders should be hidden when there's only single explosion in one tick?
Or should all of them be hidden in "biref" mode, thus shown in "full" mode?

src/main/java/carpet/logging/logHelpers/TNTLogHelper.java
```diff
@@ -24,18 +26,30 @@ public void onPrimed(double x, double y, double z, Vec3d motion)
...
+            tntCount = 0;
+            lastGametime = gametime;
+        }
+        tntCount++;
         LoggerRegistry.getLogger("tnt").log( (option) -> {
             switch (option)
             {
                 case "brief":
                     return new BaseText[]{Messenger.c(
+                            "g  #" + String.valueOf(tntCount),
```
gnembon, line 41: yeah, if you could just remove this one line from brief that would be better imo

X-Green:
Yeah

PR #224, opened 2020-04-15
Added 'sphere' iterator

Ghoulboy78:
I hope like hell it works

### DIFF
--- a/src/main/java/carpet/script/CarpetExpression.java
+++ b/src/main/java/carpet/script/CarpetExpression.java
@@ -2019,6 +2019,72 @@ public class CarpetExpression
             return (c_, t_) -> new NumericValue(finalSCount);
         });
 
+        this.expr.addLazyFunction("sphere", -1, (c, t, lv) ->
+        {
+            int lvsise = lv.size();
+            if (lvsise != 5)
+                throw new InternalExpressionException("'sphere' takes a triple of coords, a radius, and the expression");
+            int cx = (int) NumericValue.asNumber(lv.get(0).evalValue(c)).getLong();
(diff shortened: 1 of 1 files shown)


Ghoulboy78:
you see, you said in discord that it would be nice if peeps submitted prs cos ur busy cos of some mysterious reason, so I just copy pastad code from `scan` command like literally underneath, then just edited so it worked. I hope it does now

gnembon:
well, its not a sphere, its a ball. Second - its not really more efficient than running scan and adding that condition in the actual scarpet code. What would be nice is efficient sphere generator, similar to the one used in `/draw sphere` or other world edits. I don't think specific API call for balls is needed.

Having said that -there is one for diamond, mostly because of different method of iterating over an area. which makes it different. Lemme think about that. There is plenty shapes that would need to be covered and pollution built-ins space is not a good idea.

Ghoulboy78:
um, ok. Ill try and add a diamond to the /draw command. You see, I was simply inspired by your comment to try and help, and sort of did things without thinking through.

Ghoulboy78:
shall I close?

Ghoulboy78:
also, @gnembon , wdym by

> What would be nice is efficient sphere generator, similar to the one used in /draw sphere or other world edits.

Ghoulboy78:
couldn't you just you just use `/script fill (or outline) cx cy cz x1 y1 z1 x2 y2 z2 "x*x+y*y+z*z<=r*r" blue_stained_glass`, and then the same for other shape but instead with `/script run for(diamond(x,y,z,w,h),set(pos(_),'blue_stained_glass'))`
Unless you want a specific call for that in `/draw` command?

Ghoulboy78:
actually this is useless

PR #232, opened 2020-04-18
Added wacky shapes to draw command

Ghoulboy78:
Can you tell me how test to see whether it works?

nvm, i figured out

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -299,7 +299,7 @@ public class CarpetSettings
     )
     public static String commandPerimeterInfo = "true";
 
-    @Rule(desc = "Enables /draw commands", extra = "... allows for drawing simple shapes or other shapes which are sorta difficult to do normally", category = COMMAND)
+    @Rule(desc = "Enables /draw commands", extra = {"... allows for drawing simple shapes or","other shapes which are sorta difficult to do normally"}, category = COMMAND)
     public static String commandDraw = "true";
 
     @Rule(
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -299,7 +299,7 @@ public class CarpetSettings
     )
     public static String commandPerimeterInfo = "true";
 
-    @Rule(desc = "Enables /draw commands", extra = "... allows for drawing simple shapes", category = COMMAND)
+    @Rule(desc = "Enables /draw commands", extra = "... allows for drawing simple shapes or other shapes which are sorta difficult to do normally", category = COMMAND)
     public static String commandDraw = "true";
 
     @Rule(
--- a/src/main/java/carpet/commands/DrawCommand.java
+++ b/src/main/java/carpet/commands/DrawCommand.java
@@ -274,7 +274,7 @@ public class DrawCommand {
         int affected = 0;
         int offset = 0;
         if(pointup){
-            offset=height;
+            offset=height+1;
         }
         BlockPos.Mutable mbpos = new BlockPos.Mutable(pos);
 
@@ -286,16 +286,15 @@ public class DrawCommand {
 
         for (int x = -radius; x <= radius; ++x) {
 
-            for (int y = -offset; y <= height+1-offset; ++y) {
+            for (int y = 2-offset; y <= height-offset+1; ++y) {
 
                 for (int z = -radius; z <= radius; ++z) {
 
-                    if (x*x+z*z>(radius/height*y)*(radius/height*y)) {
+                    if (!(x*x+z*z<=(radius/height*y)*(radius/height*y))) {
                         continue;
                     }
 
                     mbpos.set(pos.getX() + x, pos.getY() + y + offset, pos.getZ() + z);
-
                     if (replacement == null || replacement.test(new CachedBlockPosition(world, mbpos, true))) {
                         BlockEntity tileentity = world.getBlockEntity(mbpos);
 
--- a/src/main/java/carpet/commands/DrawCommand.java
+++ b/src/main/java/carpet/commands/DrawCommand.java
@@ -165,7 +165,13 @@ public class DrawCommand {
     }
 
     private static int drawDiamond(ServerCommandSource source, BlockPos pos, int radius, BlockStateArgument block,
-            Predicate<CachedBlockPosition> replacement) {
+    Predicate<CachedBlockPosition> replacement) {
+        return (int) drawDiamond(source, pos, radius, block, replacement, false);
+    }
+
+    private static double drawDiamond(ServerCommandSource source, BlockPos pos, int radius,
+            BlockStateArgument block,
+            Predicate<CachedBlockPosition> replacement, boolean solid) {
         int affected = 0;
 
         BlockPos.Mutable mbpos = new BlockPos.Mutable(pos);
--- a/src/main/java/carpet/commands/DrawCommand.java
+++ b/src/main/java/carpet/commands/DrawCommand.java
@@ -159,12 +159,14 @@ public class DrawCommand {
     }
 
     private static int drawSphere(ServerCommandSource source, BlockPos pos, int radius, BlockStateArgument block,
-            Predicate<CachedBlockPosition> replacement) {
+            Predicate<CachedBlockPosition> replacement)
+    {
         return drawSphere(source, pos, (double) radius, (double) radius, (double) radius, block, replacement, false);
     }
 
     private static int drawSphere(ServerCommandSource source, BlockPos pos, double radiusX, double radiusY,
-            double radiusZ, BlockStateArgument block, Predicate<CachedBlockPosition> replacement, boolean solid) {
+            double radiusZ, BlockStateArgument block, Predicate<CachedBlockPosition> replacement, boolean solid)
+    {
         int affected = 0;
         ServerWorld world = source.getWorld();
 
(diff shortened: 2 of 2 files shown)


Ghoulboy78:
OMFG IT WORKS AND I MANAGED TO ACTUALLY TEST IT AND IT WORKS!!!!!!
I know I'm hyping a little, but this is a first!

Ghoulboy78:
I think thats a good name for the pr cos cones and diamonds are kinda hard to calculate.

I will work on adding cones in all three axes, but later

Ghoulboy78:
Could you please reply here telling me what other strange shapes to add?

src/main/java/carpet/commands/DrawCommand.java
```diff
@@ -109,30 +205,25 @@ private static int drawCircle(ServerCommandSource source, BlockPos pos, double r
...
                     if (!solid) {
-                        if (lengthSq(nextXn, yn, zn) <= 1 && lengthSq(xn, nextYn, zn) <= 1 && lengthSq(xn, yn, nextZn) <= 1) {
+                        if (lengthSq(nextXn, yn, zn) <= 1 && lengthSq(xn, nextYn, zn) <= 1
+                                && lengthSq(xn, yn, nextZn) <= 1) {
                             continue;
                         }
                     }
 
                     CarpetSettings.impendingFillSkipUpdates = !CarpetSettings.fillUpdates;
-                    for (int xmod = -1; xmod < 2; xmod+= 2)
```
gnembon, line 118: Please undo your egyptian braces

src/main/java/carpet/commands/DrawCommand.java
```diff
@@ -144,21 +235,269 @@ private static int drawCircle(ServerCommandSource source, BlockPos pos, double r
...
+
+        CarpetSettings.impendingFillSkipUpdates = !CarpetSettings.fillUpdates;
+
+        for (int x = -radius; x <= radius; ++x) {
+
+            for (int y = -radius; y <= radius; ++y) {
+
+                for (int z = -radius; z <= radius; ++z) {
+
+                    if (manhattan(0, 0, 0, x, y, z) != radius) {
```
gnembon, line 281: See, this is super slow. For 100x100x100 you have to analyze 1M blocks, and only place couple hundreds. It only makes sense to have these custom commands if they are more efficient than simple `/script fill <center> <center-radius> <center+radius> "abs(x)+abs(y)+abs(z)==<radius>" <block> <replace air?>`
Make sure you are only walking the blocks you need to place, like in the sphere drawer above.

src/main/java/carpet/commands/DrawCommand.java
```diff
@@ -144,21 +235,269 @@ private static int drawCircle(ServerCommandSource source, BlockPos pos, double r
...
+
+        CarpetSettings.impendingFillSkipUpdates = !CarpetSettings.fillUpdates;
+
+        for (int x = -radius; x <= radius; ++x) {
+
+            for (int y = -offset; y <= height-offset-1; ++y) {
+
+                for (int z = -radius; z <= radius; ++z) {
+
+                    if (!(x*x+z*z<=(radius/height*y)*(radius/height*y))) {
```
gnembon, line 344: Same for cone - this is not the way to iterate over blocks in a cone.

src/main/java/carpet/commands/DrawCommand.java
```diff
@@ -144,21 +235,269 @@ private static int drawCircle(ServerCommandSource source, BlockPos pos, double r
...
+
+        return 1;
+    }
+
+    private static int drawConex(ServerCommandSource source, BlockPos pos, int radius, int height, boolean pointup, BlockStateArgument block,
+    Predicate<CachedBlockPosition> replacement) {
+        return (int) drawConex(source, pos, radius, height, block, replacement, pointup, false);
+    }
+
+    private static int drawConex(ServerCommandSource source, BlockPos pos, int radius, int height,BlockStateArgument block,
```
gnembon, line 385: The Y, X and Z cone should use the same code, just parametrized.

gnembon:
Scarpet's diamond iterator is efficient and it wasn't trivial to code, especially as an iterator, so you can look it up. Its not yet how I wanted it to be - should always start from the center and build up outwards, but at least is not checking any arbitrary block in the cube area.

src/main/java/carpet/commands/DrawCommand.java
Ghoulboy78: so shall I remove?

src/main/java/carpet/commands/DrawCommand.java
Ghoulboy78: then how?

src/main/java/carpet/commands/DrawCommand.java
Ghoulboy78: could you please be more specific?

src/main/java/carpet/commands/DrawCommand.java
Ghoulboy78: shall I remove?

Ghoulboy78: also, this is a lot more efficient than the diamond iterator, as unlike the diamond which only fills in 11% of iterated blocks, the cone fills 26%. then again you could fill in the diamond and that gives 18% efficeincy (Here i calculate efficiency by the number of blocks filled divided by the number of blocks iterated over, as that gives a pretty good estimate of how many blocks are being 'wasted', so to speak)

Ghoulboy78:
btw, I'm sorry about this whole covfefe of my bad coding, I'm not a master coder like you, I just wanted to be able to contribute

src/main/java/carpet/commands/DrawCommand.java
Ghoulboy78: actually im working on improving it a ton

src/main/java/carpet/commands/DrawCommand.java
Ghoulboy78: and the new diamond iterator only ever iterates on blocks it needs to iterate on, so by this calculation, it is 100% efficient

Ghoulboy78: also, I'll have to rework the cone thing, as it currently only properly works if the base and height are the same, cos otherwise they dont return integers. I think it's as simple as just rounding, but ill try nontheless

Ghoulboy78:
BUT, I can learn, which is what I did

Ghoulboy78:
The new diamond is (according to a simple scarpet app I cooked up in 3 secs) about 300% faster that the old one. If you want I can show you the app so you can see how I did it

Ghoulboy78:
I added a new `blockset` method cos it made the code less messy, and then the future contributor doesn't need to copy pasta code from one part to another, just has to copy that one-liner. I can't believe I didn't think of this earlier. Do you think it's a good idea, gnembon?

Ghoulboy78:
also, I apologize for renaming your method, but your sphere method called drawCircle was interfering with my circle method called drawCircle, so I renamed all instances of your sphere method to drawSphere

Ghoulboy78:
Now both methods are more efficient

src/main/java/carpet/commands/DrawCommand.java
Ghoulboy78: Done

Ghoulboy78:
BTW, gnembon, why can't a command accept a string input? It just wont do it, which is why I had to make 3 branches in the command input, one for each axis. While on the one hand it means that I can have suggestions for the axes instead of just taking an input and returning an error message which is a crappier way of doing it it, it still does make the code longer.

For example, how did you get the string input for the customMOTD input?

src/main/java/carpet/commands/DrawCommand.java
Ghoulboy78: done: see fe0eccc

src/main/java/carpet/commands/DrawCommand.java
Ghoulboy78: done: see 7783f60

Ghoulboy78:
what do you think of that?

src/main/java/carpet/commands/DrawCommand.java
Ghoulboy78: ok, i googled egyptian braces. This programming jargon is so weird compared to science lab stuff.
I'll remove the egyptian braces that I added into you methods

src/main/java/carpet/CarpetSettings.java
```diff
@@ -299,7 +299,7 @@ public String description() {
     )
     public static String commandPerimeterInfo = "true";
 
-    @Rule(desc = "Enables /draw commands", extra = "... allows for drawing simple shapes", category = COMMAND)
+    @Rule(desc = "Enables /draw commands", extra = "... allows for drawing simple shapes or other shapes which are sorta difficult to do normally", category = COMMAND)
```
gnembon, line 302: please break up extra into two lines - otherwise it wraps weirdly in a UI

src/main/java/carpet/commands/DrawCommand.java
```diff
@@ -32,38 +36,233 @@ public static void register(CommandDispatcher<ServerCommandSource> dispatcher)
...
+                                )
+                            )
+                        )
+                        .then(literal("square")
+                            .then(argument("center", BlockPosArgumentType.blockPos())
+                                .then(argument("radius", IntegerArgumentType.integer(1))
+                                    .then(argument("height",IntegerArgumentType.integer(1))
+                                        .then(argument("pointing up?",BoolArgumentType.bool())
+                                            .then(literal("x")
+                                                .then(argument("block", BlockStateArgumentType.blockState())
```
gnembon, line 171: Might be useful to extract the <block> <replace <block>> subtree on its own branch so you dont need to repeat it 1000 times.

src/main/java/carpet/commands/DrawCommand.java
```diff
@@ -109,29 +308,29 @@ private static int drawCircle(ServerCommandSource source, BlockPos pos, double r
...
                         {
                             for (int zmod = -1; zmod < 2; zmod += 2)
                             {
-                                mbpos.set(pos.getX()+xmod*x, pos.getY()+ymod*y, pos.getZ()+zmod*z);
-                                if (replacement == null || replacement.test(
-                                        new CachedBlockPosition( world, mbpos, true)))
-                                {
+                                mbpos.set(pos.getX() + xmod * x, pos.getY() + ymod * y, pos.getZ() + zmod * z);
+                                if (replacement == null
+                                        || replacement.test(new CachedBlockPosition(world, mbpos, true))) {
```
gnembon, line 326: egyptian

gnembon:
Ok, looks fine. If you can fix the command structure that would be great, if not - its also fine - just lemme know. 
drawing by 3d printing is far from optimal, but would do for now.

src/main/java/carpet/CarpetSettings.java
Ghoulboy78: Done

src/main/java/carpet/commands/DrawCommand.java
Ghoulboy78: ok lemme do dat 🧠

Ghoulboy78: ofc 🧠

src/main/java/carpet/commands/DrawCommand.java
Ghoulboy78: what do you have against them, eh?

Ghoulboy78:
wdym by 3d printing? It's pretty efficient (if ur talking about the diamond) and also the cone (if ur talking about that)

src/main/java/carpet/commands/DrawCommand.java
Ghoulboy78: And now for something completely different

Ghoulboy78:
now to test if it actually works...

Ghoulboy78:
btw, gnembon, you just can't input string, liek it wont read inputted strings properly, so I can't do anything like that. Im testing now to see if osmething I saw in /spawn rates command will work

Ghoulboy78:
aaand, it.... fails... again

Ghoulboy78:
The problem is that now it freezes whenever I run it

Ghoulboy78:
ok, now it won't work, but idk why

Ghoulboy78:
ok I figured it out. Basically, @gnembon , it will read the string only if you use StringArgumentType.string, not StringArgumentType.word. ALso, you have to use switch statement to figure out what is what. Y tho?

Ghoulboy78:
(@gnembon , that wasn't a rhetorical question, why can't you read strings that way?)

Issue #237, opened 2020-04-20
[Suggestion] You should be able to do modify(player_name,'gamemode','gamemode_name')


Ghoulboy78:
and same for the `gamemode_id`. Its there for `query(player,'gamemode')`, so I don't see why it shouldn't be a thing. I may even submit pr to fix it myself.

Ghoulboy78:
actually, I just tried for 1 hour, and found it is like soo hard. I'll have to try again tomorrow, with a fresh brain

gnembon:
Ok, will let you do it. You can look for the hint in the /c /s commands.

Ghoulboy78:
thanks for the hint man

Ghoulboy78:
123rd closed issue!

PR #241, opened 2020-04-21
added gamemode and jumping modifications

Ghoulboy78:
however in sp you cant make yourself jump, but you can spawn a player and make him jump.
Also, you can't read inputted string, so I had to go with gamemode_id

### DIFF
--- a/src/main/java/carpet/script/value/EntityValue.java
+++ b/src/main/java/carpet/script/value/EntityValue.java
@@ -897,18 +897,11 @@ public class EntityValue extends Value
                 throw new InternalExpressionException("Cannot change the gamemode of non-player entities");
             }
             
-            if(NumericValue.asNumber(v).getLong()==0){
-                ((ServerPlayerEntity) e).setGameMode(GameMode.SURVIVAL);
-                return;
-            }else if(NumericValue.asNumber(v).getLong()==1){
-                ((ServerPlayerEntity) e).setGameMode(GameMode.CREATIVE);
-                return;
-            }else if(NumericValue.asNumber(v).getLong()==3){
-                ((ServerPlayerEntity) e).setGameMode(GameMode.SPECTATOR);
-                return;
-            }else if(NumericValue.asNumber(v).getLong()==2){
-                ((ServerPlayerEntity) e).setGameMode(GameMode.ADVENTURE);
-                return;
+            switch((int)NumericValue.asNumber(v).getLong()){
+                case 0:((ServerPlayerEntity) e).setGameMode(GameMode.SURVIVAL);return;
+                case 1:((ServerPlayerEntity) e).setGameMode(GameMode.CREATIVE);return;
+                case 2:((ServerPlayerEntity) e).setGameMode(GameMode.ADVENTURE);return;
+                case 3:((ServerPlayerEntity) e).setGameMode(GameMode.SPECTATOR);return;
             }
             throw new InternalExpressionException(NumericValue.asNumber(v).getLong()+" is an invalid gamemode id");
         });
--- a/src/main/java/carpet/script/value/EntityValue.java
+++ b/src/main/java/carpet/script/value/EntityValue.java
@@ -10,6 +10,7 @@ import carpet.script.EntityEventsGroup;
 import carpet.script.exception.InternalExpressionException;
 import com.mojang.brigadier.StringReader;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
+import net.minecraft.entity.effect.StatusEffects;
 import net.minecraft.network.packet.s2c.play.EntityPassengersSetS2CPacket;
 import net.minecraft.network.packet.s2c.play.EntityPositionS2CPacket;
 import net.minecraft.network.packet.s2c.play.PlayerPositionLookS2CPacket;
@@ -875,7 +876,7 @@ public class EntityValue extends Value
                     {
                         le.removeStatusEffect(effect);
                         return;
-                    } 
+                    }
                     int amplifier = 0;
                     if (lv.size() > 2)
                         amplifier = (int)NumericValue.asNumber(lv.get(2)).getLong();
@@ -896,7 +897,7 @@ public class EntityValue extends Value
             if(!(e instanceof ServerPlayerEntity)){
                 return;
             }
-            
+
             switch(v.getString().toLowerCase()){
                 case "survival":((ServerPlayerEntity) e).setGameMode(GameMode.SURVIVAL);return;
                 case "creative":((ServerPlayerEntity) e).setGameMode(GameMode.CREATIVE);return;
@@ -917,6 +918,27 @@ public class EntityValue extends Value
             if(!(e instanceof LivingEntity)) return;
 
             ((LivingEntity) e).setJumping(v.getBoolean());
+        });
+
+        put("jump",(e,v)->{//Most *REDACTED* up code in the world
+
+            float m = e.world.getBlockState(new BlockPos(e)).getBlock().getJumpVelocityMultiplier();
+            float g = e.world.getBlockState(new BlockPos(e.getX(), e.getBoundingBox().y1 - 0.5000001D, e.getZ())).getBlock().getJumpVelocityMultiplier();
+            float JumpVelocityMultiplier= (double)m == 1.0D ? g : m;
+
+            float f = (0.42F * JumpVelocityMultiplier);
+            if (((LivingEntity)e).hasStatusEffect(StatusEffects.JUMP_BOOST)) {
+                f += 0.1F * (float)(((LivingEntity)e).getStatusEffect(StatusEffects.JUMP_BOOST).getAmplifier() + 1);
+            }
+
+            Vec3d vec3d = e.getVelocity();
+            e.setVelocity(vec3d.x, (double)f, vec3d.z);
+            if (e.isSprinting()) {
+                float u = e.yaw * 0.017453292F;
+                e.setVelocity(e.getVelocity().add((double)(-MathHelper.sin(g) * 0.2F), 0.0D, (double)(MathHelper.cos(u) * 0.2F)));
+            }
+
+            e.velocityDirty = true;
         });
 
         // gamemode         [check]
--- a/src/main/java/carpet/script/value/EntityValue.java
+++ b/src/main/java/carpet/script/value/EntityValue.java
@@ -899,19 +899,18 @@ public class EntityValue extends Value
             }
 
             switch(v.getString().toLowerCase()){
-                case "survival":((ServerPlayerEntity) e).setGameMode(GameMode.SURVIVAL);return;
-                case "creative":((ServerPlayerEntity) e).setGameMode(GameMode.CREATIVE);return;
-                case "adventure":((ServerPlayerEntity) e).setGameMode(GameMode.ADVENTURE);return;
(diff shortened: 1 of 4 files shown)


Ghoulboy78:
It works, and I can now close #237

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -891,21 +893,51 @@ else if (v instanceof ListValue)
...
+                return;
+            }
+            if(v.readInteger()==2){
+                ((ServerPlayerEntity) e).setGameMode(GameMode.ADVENTURE);
+                return;
+            }
+            throw new InternalExpressionException("s"+v.getString()+"s is an invalid gamemode type");
+        });
+
+        put("jump",(e,v)->{
```
gnembon, line 919: all living entities can jump

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -4,6 +4,7 @@
 import carpet.fakes.ItemEntityInterface;
 import carpet.fakes.MobEntityInterface;
 import carpet.helpers.Tracer;
+import carpet.helpers.EntityPlayerActionPack;
```
gnembon, line 7: why that?

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -891,21 +893,51 @@ else if (v instanceof ListValue)
...
-        // "invulnerable"
-        // "dimension"
-        // "item"
-        // "count",
-        // "age",
-        // "effect_"name
-        // "hold"
-        // "hold_offhand"
-        // "jump"
+        put("gamemode_id", (e,v)->{
```
gnembon, line 896: rename to 'gamemode' and handle both string as well as numerical values. I would suggest checking variable type and if its a number use the numerical representation, otherwise use getString() and check agains names of gamemodes (case insenstitive)

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -891,21 +893,51 @@ else if (v instanceof ListValue)
...
+            if(!(e instanceof ServerPlayerEntity)){
+                throw new InternalExpressionException("Cannot change the gamemode of non-player entities");
+            }
+            if(v.readInteger()==0){
```
gnembon, line 900: Also - readInterger is for scanf-like applications, check type first and then use NumericValue.asNumber(v).getLong()

also use switch statement or if - else if

gnembon:
Thanks for looking into it :)

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: yeah, i tried that at first, but it had a ton of trouble accepting the string I typed in. I had a similar problem when trying to do the cone in my other pr, and finding that I couldnt input 'x','y' or 'z', tho I can't tell why

Ghoulboy78:
no prob man!

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: umm, i forgot to remove it, cos I was using e.EntityPlayerActionPack.Jump()

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: tho i suppose that is a different trouble

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: done

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: done

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: it wont let me unless I typecast to PlayerEntity. Maybe I'm using the wrong jump method but It just wont work

Ghoulboy78: when I typecast to LivingEntity, for which there is a jump method, it tells me `The method jump() from the type LivingEntity is not visible`

Ghoulboy78: `protected void jump() {
  }`

Ghoulboy78: in LivingEntity.class

Ghoulboy78: I dont see how to do it without making EntityValue extend LivingEntity, which cant happen

Ghoulboy78: DOne, but no it needs to be changed to 'jumping'

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: @gnembon it works only if I use switch statement for some reason

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -891,21 +892,49 @@ else if (v instanceof ListValue)
...
+            switch((int)NumericValue.asNumber(v).getLong()){
+                case 0:((ServerPlayerEntity) e).setGameMode(GameMode.SURVIVAL);return;
+                case 1:((ServerPlayerEntity) e).setGameMode(GameMode.CREATIVE);return;
+                case 2:((ServerPlayerEntity) e).setGameMode(GameMode.ADVENTURE);return;
+                case 3:((ServerPlayerEntity) e).setGameMode(GameMode.SPECTATOR);return;
+            }
+            throw new InternalExpressionException(NumericValue.asNumber(v).getLong()+" is an invalid gamemode id");
+        });
+
+        put("jumping",(e,v)->{
```
gnembon, line 916: I would still add jump(), just get access to that method through LivingEntityInterface

gnembon: as put("jump", ....)

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -891,21 +892,49 @@ else if (v instanceof ListValue)
...
-        // "item"
-        // "count",
-        // "age",
-        // "effect_"name
-        // "hold"
-        // "hold_offhand"
-        // "jump"
+        put("gamemode", (e,v)->{
+            if(!(e instanceof ServerPlayerEntity)){
+                throw new InternalExpressionException("Cannot change the gamemode of non-player entities");
```
gnembon, line 897: API should not throw exceptions if the arguments types and counts are fine, but are not applicable - should return null. Exeptions is when fromat of arguments is incorrect.

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -891,21 +892,49 @@ else if (v instanceof ListValue)
...
+            }
+            
+            switch(v.getString().toLowerCase()){
+                case "survival":((ServerPlayerEntity) e).setGameMode(GameMode.SURVIVAL);return;
+                case "creative":((ServerPlayerEntity) e).setGameMode(GameMode.CREATIVE);return;
+                case "adventure":((ServerPlayerEntity) e).setGameMode(GameMode.ADVENTURE);return;
+                case "spectator":((ServerPlayerEntity) e).setGameMode(GameMode.SPECTATOR);return;
+            }
+
+            switch((int)NumericValue.asNumber(v).getLong()){
```
gnembon, line 907: Check value type first and then use numbers, otherwise check string.

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: done

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: idk how to do

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: what is LivingEntityInterface?

Ghoulboy78: I did it, but I had to do it by basically copy pasting code from the jump() mthod in LivingEntity.java, then copying abit of code form other places which was also protected and private, and finally got something that may or may not work

Ghoulboy78: at least it works...

src/main/java/carpet/script/value/EntityValue.java
gnembon: check instanceof of the 'v'

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -891,21 +893,68 @@ else if (v instanceof ListValue)
...
+            }
+        });
+
+        put("jumping",(e,v)->{
+            if(!(e instanceof LivingEntity)) return;
+
+            ((LivingEntity) e).setJumping(v.getBoolean());
+        });
+
+        put("jump",(e,v)->{//Most *REDACTED* up code in the world
```
gnembon, line 922: I agree - should have made the jump() call public and just call it here.

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: ah tru

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: nah, that would have made it too simple

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: wai,t it wont let me do if(v instanceof String)
It's an instance of a carpet.script.value.Value

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: In fact, I added it to MobAi.java (best place I could think to put it in, also cos you may need it when (and if) you r gonna add pathfinding to scarpet) I added it to carpet.utils, not carpet.script.utils, cos I thought that it felt more appropriate in MobAi as opposed to WorldTools

src/main/java/carpet/utils/MobAI.java
```diff
@@ -93,4 +98,24 @@ public static TrackingType byName(String aspect)
         }
     }
 
+    public static void jump(LivingEntity e){//Most *REDACTED* up code in the world
```
gnembon, line 101: oh, boi - use vanilla code, just get access to that via an interface...

Ghoulboy78: wdym by use and interface...

gnembon: one example: https://github.com/gnembon/fabric-carpet/blob/94ffd5b9900653f1a2994bf051046790a87e5168/src/main/java/carpet/helpers/XPcombine.java#L40

Ghoulboy78: idk how to do

gnembon: ok, will merge as is, and I will fix the accessors

PR #250, opened 2020-04-24
Allow to query mob category

Ghoulboy78:
It's like 2 lines, but soo useful


src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -479,6 +479,8 @@ else if (entities)
                 return new NBTSerializableValue(nbttagcompound);
             return new NBTSerializableValue(nbttagcompound).get(a);
         });
+
+        put("category",(e,a)->{return new StringValue(e.getType().getCategory().toString().toLowerCase());});
```
gnembon, line 483:
```suggestion
        put("category",(e,a)->{return new StringValue(e.getType().getCategory().toString().toLowerCase(Locale.ROOT));});
```

Ghoulboy78:
yo, what is Locale.root?

PR #267, opened 2020-05-11
Add Language Support For Carpet Settings

X-Green:
use /set-carpet-language command to change ur language.


Ghoulboy78:
Could you also add en_us.json, so we that don't speack chinese can use it?

X-Green:
> Could you also add en_us.json, so we that don't speack chinese can use it?

As the original language , coded in the source, is en_us, it will keep the default language if you don't use the command -- meaning no changes are made for English players.

X-Green:
> Could you also add en_us.json, so we that don't speack chinese can use it?

Also, Spanish file is being made

X-Green:
Well, I changed the logic so that language's now changed by command ：
/carpet carpetLanguage xxx
And it will update your language when listAllSettings() function is called.

src/main/java/carpet/settings/SettingsManager.java
```diff
@@ -20,23 +21,9 @@
...
 import org.apache.logging.log4j.util.TriConsumer;
 
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.FileReader;
-import java.io.FileWriter;
-import java.io.IOException;
-import java.io.PrintStream;
+import java.io.*;
```
gnembon, line 24: Please expand these, this is very useful for vc and control of minecraft updates.

gnembon:
Thanks for starting that aspect. I like the approach, would probably shorten some names, but that can be adjusted later. 

One thing that's problematic, is renaming the rules as well. Those are itentifiers, used in configuration etc. plus translations don't affect values etc. It would be quite weird to type `/carpet xxxxxx true` where xxxx is not in english, and carpet and true, are in english. Please exclude rule names from translations.

src/main/java/carpet/settings/SettingsManager.java
Ghoulboy78: what is vc? Isn't it version control?

Ghoulboy78:
and translate only what the rule actually does?
so normally if you hover over , say, flippingCactus, it will say:
'Right clicking with cactus will flip block'.
So then if I put, say, galician, it will still say flippinCactus, but hovering over it will say:
'clic esquerdo co cacto vai virar bloque', and Italian will say
'Premere il tastro destro con un cactus girerà il blocco'?

src/main/java/carpet/settings/SettingsManager.java
gnembon: yes

gnembon:
like if there is a translation, display original name and maybe translation in the brackets. The thing is you would not now which rule to set by name if that was muddled up
so instead of 
 - 传送门窒息修复 [true] [false]
you would have
 - portalSuffocationFix (传送门窒息修复) [true] [false]

X-Green:
image
This may be its fix?

gnembon:
yes, as long as actual commands in the back can still be handled with default logic and values. Will merge tomorrow in case there are some other things to iron out, since there might be a deploy tomorrow.

gnembon:
I assume that would be the beginning and once first pass it merged in there will be more translations, for more things, right? I can be adding to the en_us.json as a model.

X-Green:
Yes. An en_us file with UI-Keys only is currently available in the same dir. It would  be nice to have a model to make other files for translation.

X-Green:
I came up with another idea that the language setting should be done by /log rather than /carpet, because mostly translations are for specified players. /log would be better than /carpet in such case.

Ghoulboy78:
but then you have to reapply every time you enter the server, cos you can't set your own defaultLoger then you're not op

X-Green:
🤔
Thus, the best way might be a new command like "/set-carpet-lang"?

Ghoulboy78:
or maybe it automagically detects what language u have, and if it can it will change to that lang, and if not it will do american

gnembon:
its fine as it is. commands are server side and handling them per client might be challenging especially when reporting errors where player context is unknown (via log error) - you would want to get them translated too.

PR #286, opened 2020-05-18
Added hunger, saturation and exhaustion modification and querying

### LINKED ISSUE #230: Suggestion: Query and modify player's hunger
It would be nice if Scarpet can query and modify a player's 
* [x] food level (`query(e, 'hunger')` and `modify(e, 'hunger', value)`)
* [x] saturation level (`query(e, 'saturation')` and `modify(e, 'saturation', value)`)
* [ ] exhaustion level (`query(e, 'exhaustion')` and `modify(e, 'exhaustion', value)`)

This is so Scarpet apps can add abilities that directly cost food points rather than having to go through the vanilla food system or the Hunger effect. Currently, there is no way to directly change a player's food level as in-game NBT editing doesn't work on players.

Ghoulboy78:
But I can't for the life of me query exhaustion. You're gonna have to do your magic with Interfaces.
Fixes #230

### DIFF
--- a/src/main/java/carpet/script/value/EntityValue.java
+++ b/src/main/java/carpet/script/value/EntityValue.java
@@ -284,6 +284,7 @@ public class EntityValue extends Value
         put("sprinting", (e, a) -> e.isSprinting()?Value.TRUE:Value.FALSE);
         put("swimming", (e, a) -> e.isSwimming()?Value.TRUE:Value.FALSE);
         put("hunger", (e, a) -> new NumericValue(((PlayerEntity) e).getHungerManager().getFoodLevel()));
+        put("saturation", (e, a) -> new NumericValue(((PlayerEntity) e).getHungerManager().getSaturationLevel()));
         put("jumping", (e, a) -> {
             if (e instanceof LivingEntity)
             {
@@ -916,6 +917,10 @@ public class EntityValue extends Value
         put("fire",(e,v)-> e.setFireTicks((int)NumericValue.asNumber(v).getLong()));
 
         put("hunger", (e, v)-> ((PlayerEntity) e).getHungerManager().setFoodLevel((int)NumericValue.asNumber(v).getLong()));
+
+        put("exhaustion", (e, v)-> ((PlayerEntity) e).getHungerManager().addExhaustion((int)NumericValue.asNumber(v).getLong()));
+
+        put("saturation", (e, v)->((PlayerEntity) e).getHungerManager().setSaturationLevelClient((float)NumericValue.asNumber(v).getLong()));
 
         // gamemode         [check]
         // spectate         [check]
--- a/src/main/java/carpet/script/value/EntityValue.java
+++ b/src/main/java/carpet/script/value/EntityValue.java
@@ -1,17 +1,21 @@
 package carpet.script.value;
 
+import carpet.CarpetSettings;
 import carpet.fakes.EntityInterface;
 import carpet.fakes.ItemEntityInterface;
 import carpet.fakes.LivingEntityInterface;
 import carpet.fakes.MobEntityInterface;
+import carpet.fakes.HungerManagerInterface;
 import carpet.helpers.Tracer;
 import carpet.patches.EntityPlayerMPFake;
 import carpet.script.CarpetContext;
 import carpet.script.EntityEventsGroup;
 import carpet.script.argument.Vector3Argument;
 import carpet.script.exception.InternalExpressionException;
+import carpet.utils.Messenger;
 import com.mojang.brigadier.StringReader;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
+import net.minecraft.entity.player.HungerManager;
 import net.minecraft.network.packet.s2c.play.EntityPassengersSetS2CPacket;
 import net.minecraft.network.packet.s2c.play.EntityPositionS2CPacket;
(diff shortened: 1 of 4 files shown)


src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -906,21 +908,19 @@ else if (v instanceof ListValue)
             }
         });
 
-        put("silent",(e,v)->{
-           e.setSilent(v.getBoolean());
-        });
+        put("silent",(e,v)-> e.setSilent(v.getBoolean()));
```
James103, line 911: Why did you compress
```
        put("silent",(e,v)->{
           e.setSilent(v.getBoolean());
        });
```
to `put("silent",(e,v)-> e.setSilent(v.getBoolean()));`? The original is more readable.
(same for next 3 occurrences)

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -906,21 +908,19 @@ else if (v instanceof ListValue)
...
-            e.setInvulnerable(v.getBoolean());
-        });
+        put("invulnerable",(e,v)-> e.setInvulnerable(v.getBoolean()));
 
-        put("fire",(e,v)->{
-            e.setFireTicks((int)NumericValue.asNumber(v).getLong());
-        });
+        put("fire",(e,v)-> e.setFireTicks((int)NumericValue.asNumber(v).getLong()));
+
+        put("hunger", (e, v)-> ((PlayerEntity) e).getHungerManager().setFoodLevel((int)NumericValue.asNumber(v).getLong()));
```
James103, line 919: As above, this should be expanded to something like this (including removing the extra parenthesis):
```suggestion
        put("hunger", (e, v)-> {
            ((PlayerEntity) e).getHungerManager().setFoodLevel((int)NumericValue.asNumber(v).getLong());
        }
```
(same for next 2 occurrences)

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: not really. I like it better this way. It's also more in keeping with gnembon's other on-liners

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: yah, no

Ghoulboy78: I don't think it's necessary, and it bloats file unnecessarily, and I think it's better to look at this way.

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -283,6 +283,8 @@ public Value get(String what, Value arg)
         put("sneaking", (e, a) -> e.isSneaking()?Value.TRUE:Value.FALSE);
         put("sprinting", (e, a) -> e.isSprinting()?Value.TRUE:Value.FALSE);
         put("swimming", (e, a) -> e.isSwimming()?Value.TRUE:Value.FALSE);
+        put("hunger", (e, a) -> new NumericValue(((PlayerEntity) e).getHungerManager().getFoodLevel()));
```
gnembon, line 286: you have to make sure it won't crash if e is not a player - returning null, not npe

gnembon:
what's your problem with exhaustion?

src/main/java/carpet/script/value/EntityValue.java
gnembon: I am fine either way.

best way - not changing anything. This code lives in two mc versions at a time (1.16) so better not change / refactor unnecessarily code since it gets messy when merging with 1.16 branch

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: ah ok, I hadn't tested that

Ghoulboy78:
image
exhaustion has private access

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: so what, should I change or no change?

gnembon:
I will let you figure out the access through interfaces - its crucial for mixin modding to learn it. Check how serverplayerentity interface is done - a similar thing is needed for playerentity for that. Not sure what exhaustion corresponds to though - food level is chicken drumsticks, saturation is the hidden chicken drumsticks and not sure what saturation does and why its a value.

Ghoulboy78:
do you mean exhaustion? I think it's how you lose saturation, for taking damage. I think so cos I could trace back the var exhaustion to HungerManager.java __and__ to DamageSource.java

Ghoulboy78:
@gnembon Ok, I will definitely figure that stuff out, if you think it's important. But I'm afraid that will take a week or so, because I am currently in the middle of my examinations, so I will be basically unavailable for the entirety of the next week, so don't expect anything until basically June

gnembon:
don't worry

Ghoulboy78:
IT WORKS! I mean, i dunno what the f exhaustion does, but still!!!

Ghoulboy78:
@gnembon I am a mixin modder now!

gnembon:
congrats

Ghoulboy78:
but wait, that means that this pr is complete, no?

PR #289, opened 2020-05-23
Update math.scl

Ghoulboy78:
So we can import the necessary functions. I may have forgotten the other files, but I'm too bloody tired to check rn


gnembon:
well distance should not be call distance but euclidean, since its unclear. Also lcm and hcf are not really needed for anybody. If there is some work done on libraries, I would think only on things that would be applicable to more people.

Ghoulboy78:
But I do think there are some funcs that would be useful to have in a semi-lib, for the ease of importing

Ghoulboy78:
actually, this is just cluttering up the pr's, as gnembon ain't gonna accept anyways

PR #292, opened 2020-05-25
Removing your balls

Ghoulboy78:
Cos the way you implemented them (in the draw command, what did you **think** I meant?) may be ambiguous, especially since some balls are hollow (you don't wanna be kicking a round piece of solid leather across a field)
And cos asking for the sphere to be filled makes, imo, more sense

That title tho.

### DIFF
--- a/src/main/java/carpet/commands/DrawCommand.java
+++ b/src/main/java/carpet/commands/DrawCommand.java
@@ -5,6 +5,7 @@ import carpet.settings.SettingsManager;
 import carpet.utils.Messenger;
 import com.google.common.collect.Lists;
 import com.mojang.brigadier.CommandDispatcher;
+import com.mojang.brigadier.arguments.BoolArgumentType;
 import com.mojang.brigadier.builder.LiteralArgumentBuilder;
 import com.mojang.brigadier.arguments.IntegerArgumentType;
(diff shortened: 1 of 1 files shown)


Ghoulboy78:
oh, and btw, here is a scarpet app I cooked up in 3 secs, which does diamonds with recursion, the way you wanted them to be done:
```
diam(radius,block, filled)->(
    pos=pos(block(pos(player())));//so it rounds
    __diamond(pos,radius,pos,radius,block, filled)
);

_manhattan(pos1,pos2)->return(abs(pos1:0-pos2:0)+abs(pos1:1-pos2:1)+abs(pos1:2-pos2:2));

__diamond(init_pos,init_radius,pos,rc,block, filled)->(
    if(rc<0,return());//cos if not crash
    for(neighbours(pos),
        if((filled)||(_manhattan(pos(_),init_pos)==init_radius+1),set(pos(_),block));
        __diamond(init_pos,init_radius,pos(_),rc-1,block, filled)
    )
)
```

Ghoulboy78:
oh, and btw, this app is running slower than actually the '3d printing' which is 'not ideal' which I was doing earlier, though it is admittedly easier to understand.

gnembon:
https://en.wikipedia.org/wiki/Ball_(mathematics)

PR #298, opened 2020-05-29
Add mobcap to scarpet

Ghoulboy78:
running mobcap() will return mobcaps for the dimension,
Somehow IDEA screwed up the imports. Sorry.


src/main/java/carpet/script/CarpetExpression.java
```diff
@@ -99,11 +92,7 @@
 import net.minecraft.structure.StructurePiece;
 import net.minecraft.structure.StructureStart;
 import net.minecraft.text.LiteralText;
-import net.minecraft.util.ActionResult;
```
gnembon, line 102: Yeah, fix those, and change in your IDE not to do that again.

Ghoulboy78: how to fix the IDE tho?

Ghoulboy78:
Now it returns a list, but the code is kinda messy. It basicaly gives list: `l('MONSTER','3','CREATURE','4')`,blabla
I dunno how to fix that or if I even wanna fix it. Idk man, ur call.

gnembon:
will trash it for now.

Reason is that in 1.16 - its considerably easier to get that information from the game, and 1.16 is right around the corner. Will add that to 1.16 branch instead.

Ghoulboy78:
yeah, true, but you can't get it scarpet can u?

Ghoulboy78:
ah no wait I got ur point. ok

gnembon:
your solution is the same as running `hostile_counts = for(entity_list('*'), _ ~'category'=='monster')`

Ghoulboy78:
oh ofc. That wud have been easier.

PR #299, opened 2020-05-30
Adding Explosion Logger

X-Green:
It will show:
1.gametime of explosion
2.position
3.if any block will be affected
4.pushed entities by explosion
the entities will be showned by its type, position as well as delta speed.
entities with same type, position, and deltaSpeed in a single explosion will be merged into one entry.

### DIFF
--- a/src/main/java/carpet/mixins/ExplosionMixin.java
+++ b/src/main/java/carpet/mixins/ExplosionMixin.java
@@ -2,9 +2,13 @@ package carpet.mixins;
 
 import carpet.helpers.OptimizedExplosion;
 import carpet.CarpetSettings;
+import carpet.logging.LoggerRegistry;
+import carpet.logging.logHelpers.ExplosionLogHelper;
 import net.minecraft.block.BlockState;
 import net.minecraft.block.Blocks;
+import net.minecraft.entity.Entity;
 import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.Vec3d;
 import net.minecraft.world.World;
 import net.minecraft.world.explosion.Explosion;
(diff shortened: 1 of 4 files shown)


src/main/java/carpet/mixins/ExplosionMixin.java
```diff
@@ -58,4 +70,24 @@ private BlockState noBlockCalcsWithNoBLockDamage(World world, BlockPos pos)
...
+    {
+        if (LoggerRegistry.__explosions && ! world.isClient)
+        {
+            eLogger = new ExplosionLogHelper(entity, x, y, z, power, createFire, blockDestructionType);
+        }
+    }
+
+    @Redirect(method = "collectBlocksAndDamageEntities",
+            at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/Entity;setVelocity(Lnet/minecraft/util/math/Vec3d;)V"))
+    private void setVelocityAndUpdateLogging(Entity entity, Vec3d velocity)
```
gnembon, line 86: This might not work if optimized explosions are in plus it may collide with lithium explosions. You should add these trackers to carpet optimized code for explosions as well as test with lithium making sure there is no mixin issues with mod compat.

X-Green: I tested and found it capable with lithium but not with carpet tnt optimization. I will have it fixed as soon as possible.

gnembon: cool np

X-Green:
Done. XD

gnembon:
can you drop me a message on discord - trying to locate an email or any contact to you.

PR #316, opened 2020-06-07
Added serverTime() func, fixes #314

### LINKED ISSUE #314: [suggestion] serverTime() Function
Per Gnembon's suggestion in Discord I am making this bug tracker to suggest that a serverTime() function be added to return actual server time. This could be very useful for timekeeping even across server resets and outages.

Ghoulboy78:
_No description provided_

### DIFF
--- a/src/main/java/carpet/script/CarpetExpression.java
+++ b/src/main/java/carpet/script/CarpetExpression.java
@@ -2967,6 +2967,12 @@ public class CarpetExpression
             return (cc, tt) -> ret;
         });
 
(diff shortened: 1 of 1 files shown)


Ghoulboy78:
fixes #314

Ghoulboy78:
Fixes #314

gnembon:
first of all - all functions are snake case, not camel case
second of all no docs added
thrid - its not doing what its supposed to - here you are picking the last server tick lengths

PR #350, opened 2020-06-25
Added united tick display in logging & Normalized variables in LoggerRegistry.java

X-Green:
It would be better to let variables rank in alphabetic orders...

### DIFF
--- a/src/main/java/carpet/logging/Logger.java
+++ b/src/main/java/carpet/logging/Logger.java
@@ -231,4 +231,9 @@ public class Logger
         if (options != null && Arrays.asList(options).contains(arg)) return arg;
         return null;
     }
+
+    public Map<String, String> getSubscribedOnlinePlayers()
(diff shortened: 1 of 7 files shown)


James103:
Your commit message say "Add files via upload" repeated 3 times. Could you make the commit messages more descriptive than just that, so other people know what's actually going on?

gnembon:
Can you explain how that would be useful? We are here talking about printing tick information to chat which can be easily achieved with a scarpet app, or even a vanilla datapack, right?

X-Green:
> Can you explain how that would be useful? We are here talking about printing tick information to chat which can be easily achieved with a scarpet app, or even a vanilla datapack, right?

Hmm...
That's only a change to make sure that the ticknumber display of different loggers won't be repeated and different in pattern.
As for self-achieving chat information, I'm now working on it.

PR #366, opened 2020-07-04
Allow "/player look" to target locations in the world with "/player look at"

hkva:
Hi. This pull request adds the ability for fake players to look at a location in the world with `/player <name> look at <x> <y> <z>` as an alternative to supplying cardinal directions or angles. This is useful if you ever want to have a fake player do something that requires precise aim like activating a button.

Here's a quick demo of the change in action: https://youtu.be/njLHSUuTt44


TCasseBloc:
Are entities also supported like in the `facing entity <selector>` argument in /tp ?

PR #370, opened 2020-07-07
HUD logger hook for carpet extensions

Fallen-Breath:
It's currently not avaliable for carpet extension to add its own HUD loggers without mixin into fabric carpet.
What I did is add a extra method`onUpdateHUD` in `CarpetExtension` Interface and invoke it after every fabric-carpet HUD loggers have finished their jobs and before the logging message is sent to players.

I'm not 100% sure if these 2 lines of code work since I don't know how to test that qwq

One more feature suggestion: give HUD loggers a priority number about who will be on the top and who will be at the bottom. That might be useful but idk xd


gnembon:
will merge that but gonna change it to a listener model

PR #321, opened 2020-06-10
Add support for triples of positions in volume

Ghoulboy78:
Fixes #222

PR #375, opened 2020-07-12
Firigion's suggestion(s)

Ghoulboy78:
The suggestions that Firigion got me, there were a list, but now there is only one, to do the entity area thing that I forgot to do with #321

### DIFF
--- a/src/main/java/carpet/script/CarpetExpression.java
+++ b/src/main/java/carpet/script/CarpetExpression.java
@@ -2062,6 +2062,10 @@ public class CarpetExpression
 
         this.expr.addLazyFunction("entity_area", -1, (c, t, lv) -> {
             CarpetContext cc = (CarpetContext) c;
+
+            if(lv.size()!=3||lv.size()!=5||lv.size()!=7)
+                throw new InternalExpressionException("Incorrect number of arguments, entity_area requires 3, 5 or 7 arguments");
+
             BlockArgument centerLocator = BlockArgument.findIn(cc, lv, 1);
             BlockPos startpos = centerLocator.block.getPos();
             BlockPos corner = BlockArgument.findIn(cc, lv, centerLocator.offset).block.getPos();
--- a/src/main/java/carpet/script/CarpetExpression.java
+++ b/src/main/java/carpet/script/CarpetExpression.java
@@ -2060,7 +2060,7 @@ public class CarpetExpression
             return (_c, _t ) -> retval;
         });
 
-        this.expr.addLazyFunction("entity_area", 7, (c, t, lv) -> {
+        this.expr.addLazyFunction("entity_area", -1, (c, t, lv) -> {
             CarpetContext cc = (CarpetContext) c;
             BlockArgument pos1Locator = BlockArgument.findIn(cc, lv, 1);
             BlockPos pos1 = pos1Locator.block.getPos();
--- a/src/main/java/carpet/script/CarpetExpression.java
+++ b/src/main/java/carpet/script/CarpetExpression.java
@@ -2062,19 +2062,14 @@ public class CarpetExpression
 
         this.expr.addLazyFunction("entity_area", -1, (c, t, lv) -> {
             CarpetContext cc = (CarpetContext) c;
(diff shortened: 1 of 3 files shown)


src/main/java/carpet/script/CarpetExpression.java
```diff
@@ -2061,23 +2061,28 @@ else if ("!spectating".equalsIgnoreCase(playerName))
         });
 
         this.expr.addLazyFunction("entity_area", 7, (c, t, lv) -> {
```
gnembon, line 2063: Then the number arguments can suddenly be variable. Seems you haven't try that since that would fail if you supply with less than 7 args, right?

Ghoulboy78: oopsi silly me

Ghoulboy78:
should be done now

src/main/java/carpet/script/CarpetExpression.java
```diff
@@ -2060,24 +2060,24 @@ else if ("!spectating".equalsIgnoreCase(playerName))
...
-                    NumericValue.asNumber(lv.get(5).evalValue(c)).getDouble(),
-                    NumericValue.asNumber(lv.get(6).evalValue(c)).getDouble()
+        this.expr.addLazyFunction("entity_area", -1, (c, t, lv) -> {
+            CarpetContext cc = (CarpetContext) c;
+            BlockArgument centerLocator = BlockArgument.findIn(cc, lv, 1);
+            BlockPos startpos = centerLocator.block.getPos();
+            BlockPos corner = BlockArgument.findIn(cc, lv, centerLocator.offset).block.getPos();
+
+            Box area = new Box(startpos).expand(
+                    corner.getX(),
```
gnembon, line 2070: problem is that if the second argument is an actual block argument, you will end up stretching the selected area by absolute coordinates of the other block. Like you have done in the other PRs (scan)  - if the second block argument is a world-localized block -  you should take a difference, othewise one ca assume its a relative coord

src/main/java/carpet/script/CarpetExpression.java
```diff
@@ -2060,24 +2060,24 @@ else if ("!spectating".equalsIgnoreCase(playerName))
...
+                    corner.getY(),
+                    corner.getZ()
             );
             String who = lv.get(0).evalValue(c).getString();
```
gnembon, line 2074: Also make sure since the number of arguments is not guaranteed, and the error will not be thrown, make sure that there it at least 2 arguments.

src/main/java/carpet/script/CarpetExpression.java
Ghoulboy78: image
I say it makes more sense like that

Ghoulboy78: Plus, how do I tell if someone is entering a block position or if they're entering a triple of the distances?

src/main/java/carpet/script/CarpetExpression.java
Ghoulboy78: true

src/main/java/carpet/script/CarpetExpression.java
gnembon: refer to https://github.com/gnembon/fabric-carpet/blob/5ed55ecf9e07bdcb98b803a4c845771f68c3c08f/src/main/java/carpet/script/CarpetExpression.java#L2176

Ghoulboy78: ok thanks

Ghoulboy78:
What's up with the conflicts that must be resolved?

gnembon:
yeah, you must resolve them first

gnembon:
If you have troubles resolving these merge conflicts - I can apply this change directly to master, if you are fine with it.

Ghoulboy78:
yeah thx

Ghoulboy78:
sorry for not doing myself, it's just that I'm rly busy these days.

gnembon:
ok, closing. Will apply manually

PR #383, opened 2020-07-17
Add rule to make phantoms respect the Hostile MobCap when spawning

eatmyvenom:
So basically we were talking about this and xcom said that this was the best idea instead of just toggling the gamerule. I agree so yea i made this :)


gnembon:
Well, I stated a few times that it doesn't need to exist if there is a gamerule. You can add that to carpet-extra, but it will be just confusing in the main carpet to have both.

PR #390, opened 2020-07-23
added "ops" setting to each rule

### LINKED ISSUE #355: ops Permission Level Not in Autocomplete in 1.4
Don't know if you know this or not, but the ops permission level for command settings added in 1.4 is not in autocomplete.

Ghoulboy78:
Fixes #355

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -270,20 +270,21 @@ public class CarpetSettings
     @Rule( desc = "summoning a lightning bolt has all the side effects of natural lightning", category = CREATIVE )
     public static boolean summonNaturalLightning = false;
 
-    @Rule(desc = "Enables /spawn command for spawn tracking", category = COMMAND)
+    @Rule(desc = "Enables /spawn command for spawn tracking", category = COMMAND, options = {"true", "false", "ops"})
     public static String commandSpawn = "true";
 
-    @Rule(desc = "Enables /tick command to control game clocks", category = COMMAND)
+    @Rule(desc = "Enables /tick command to control game clocks", category = COMMAND, options = {"true", "false", "ops"})
     public static String commandTick = "true";
 
     @Rule(
             desc = "Enables /profile command to monitor game performance",
             extra = "subset of /tick command capabilities",
-            category = COMMAND
+            category = COMMAND,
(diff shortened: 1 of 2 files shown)


gnembon:
That's not a correct place to put it, add it here:
https://github.com/gnembon/fabric-carpet/blob/3f593943c9d92a2b5ef1bdf9d7b4f87bebd377fe/src/main/java/carpet/settings/ParsedRule.java#L50

Ghoulboy78:
wdym, put it there?

gnembon:
well, this code applies custom operations to rules that are commands and are registered as a string. You can put options to 'true', 'false', 'ops' there.

Ghoulboy78:
how though?

Ghoulboy78:
don't you mean here?
https://github.com/gnembon/fabric-carpet/blob/3f593943c9d92a2b5ef1bdf9d7b4f87bebd377fe/src/main/java/carpet/settings/ParsedRule.java#L64-L67

gnembon:
yes - there exactly

Ghoulboy78:
ok done, just wanted to confirm before committing, cos now I gotta undo all the stuff I did initially

src/main/java/carpet/settings/ParsedRule.java
```diff
@@ -63,7 +63,7 @@
         }
         else if (this.type == boolean.class || (this.type == String.class && categories.contains(RuleCategory.COMMAND)))
```
gnembon, line 64: if its a boolean type, then do not add 'ops'. Add ops only if its string class and command

src/main/java/carpet/CarpetSettings.java
```diff
@@ -652,4 +650,4 @@ public static void changeSpawnSize(int size)
             validate = Validator.PROBABILITY.class
     )
     public static double creativeFlyDrag = 0.09;
-}
+}
```
gnembon, line 653: while at it, you might revert all the changes to CarpetSettings - not to introduce unnecessary diffs.

Ghoulboy78: done

src/main/java/carpet/settings/ParsedRule.java
Ghoulboy78: done

PR #400, opened 2020-08-01
Corrected while() documentation

SirGavith:
Added `expr` to the args of while() in both places in the documentation
I noticed it was missing the other day, so here's a PR adding the args

### DIFF
--- a/docs/scarpet/Full.md
+++ b/docs/scarpet/Full.md
@@ -1109,7 +1109,7 @@ for(range(1000000,1100000),check_prime(_))  => 7216
 
 From which we can learn that there is 7216 primes between 1M and 1.1M
 
-### `while(cond, limit)`
+### `while(cond, limit,expr)`
 
 Evaluates expression `expr` repeatedly until condition `cond` becomes false, but not more than `limit` times. 
 Returns the result of the last `expr` evaluation, or `null` if nothing was successful. Both `expr` and `cond` will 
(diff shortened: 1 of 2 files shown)


docs/scarpet/language/LoopsAndHigherOrderFunctions.md
```diff
@@ -49,7 +49,7 @@ for(range(1000000,1100000),check_prime(_))  => 7216
 
 From which we can learn that there is 7216 primes between 1M and 1.1M
 
-### `while(cond, limit)`
+### `while(cond, limit,expr)`
```
James103, line 52: To maintain consistency, it's a good idea to put spaces after commas in function arguments.
```suggestion
### `while(cond, limit, expr)`
```

docs/scarpet/Full.md
```diff
@@ -1109,7 +1109,7 @@ for(range(1000000,1100000),check_prime(_))  => 7216
 
 From which we can learn that there is 7216 primes between 1M and 1.1M
 
-### `while(cond, limit)`
+### `while(cond, limit,expr)`
```
James103, line 1112: Full.md is auto-generated from the individual documentation files.

SirGavith: Should I make another commit, or just leave this be?

gnembon: editing Full has no effect - doesn't matter too much

PR #429, opened 2020-08-23
Minor corrections to Entities.md

TheCatSaber:
Moved
<pre>
Run(‘/kill ’ + e~'command_name’);
</pre>
To modify(e, ‘kill’) since that makes more sense than it being at query(e, 'persistence').


Ghoulboy78:
I think it was meant to go into the `query(e,'command_name')` as an example of how that works

gnembon:
yep, that was for the command_name - right above it.

TheCatSaber:
Is that where?

Ghoulboy78:
yes

Issue #420, opened 2020-08-16
issues with ai_tracker and fabulous graphics


bitslinger70:
video settings graphics option on Fast and Fancy are fine but when I set it to Fabulous the ai_tracker visuals can only be seen briefly when in just the right position.

altrisi:
The problem isn't specifically with `ai_tracker`, but with the general Shapes API. It used to work in previous versions (1.16.1, with fabulous graphics on).

- When running fabulous, you aren't able to see Carpet shapes (with the mod in the client-side)
- In lower graphical settings, Carpet shapes are rendered normally (with the mod in the client-side)

Without the server seeing Carpet in the client:
- Shape particles are generated once per shape, and therefore the `duration` argument is ignored (doesn't affect built-in apps, since they send new shapes regularly, but may affect custom apps), but the particles are rendered correctly in fabulous

gnembon:
lol 420

PR #431, opened 2020-08-24
Ai_Tracker app bug

### LINKED ISSUE #430: ai_tracker villager_breeding timer is red when villagers can breed and green then they can't.
For food and bed, they are green when the requirement is met and red when it is not. The timer is the opposite.

Ghoulboy78:
Fixes #430 

Basically it showed breeding_age of 0 as red, when in fact it ought to be green, as it is a necessary requirement for breeding of villagers.


James103:
The commit message should be "Fix #430" instead of "Fix #420", otherwise that issue will be closed instead of the originally-mentioned issue.

Issue #437, opened 2020-08-28
Suggestion: Ability to query and modify an entity's air level


James103:
As of Carpet mod 1.4.0 for Minecraft 1.16.1 (and by extension Carpet mod 1.4.9 for Minecraft 1.16.2 due to relevant code not being touched), there is no way to query or modify an entity's air level via the Scarpet entity API.

I would like if there were two new functions `query(e, 'air')` and `modify(e, `air`, number)` that directly affected an entity's air level. If an entity's air level is set to zero every game tick, then the entity will take drowning damage even if the entity can normally breathe in air, while if an entity's air level is set to non-zero values every game tick, then the entity will not take drowning damage.

gnembon:
good idea

PR #452, opened 2020-09-04
Updating logger to support different types

### LINKED ISSUE #445: [Scarpet Suggestion] Multiple log levels for Scarpet logger() function
Currently, the `logger()` function in Scarpet always uses the `error` log type, and there is no way to change it.
My suggestion would be an argument to specify the "severity" of the message to be logged, since currently the only way to change this is by using `print()` instead of `logger()`, which would use the regular `info` log type, but would also print the message in chat in most cases, and still `debug`, `warn` and `fatal` are unreachable.

(names based in the vanilla launcher's output log viewer)

Ghoulboy78:
Fixes #445 

Im not sure whether it should output error by default, or I should hardcode at 2 by default, or maybe just set it to info by default. I think info makes more sense than warn...

### DIFF
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -245,13 +245,26 @@ Example usages:
   // not a problem in apps
 </pre>
 
-### `logger(expr)`
+### `logger(msg), logger(type, msg)`
 
-Prints the message to system logs, and not to chat.
+Prints the message to system logs, and not to chat. By default prints an info, unless you specify otherwise in the `type` parameter.
 
-### read_file(resource, type)
-### delete_file(resource, type)
-### write_file(resource, type, data, ...)
+Available output types:
+
+`'debug'`: white
+
+`'warn'`: yellow
+
+`'fatal'`: red
+
+`'info'`: grey
+
+`'error'`: orange
+
+
+### `read_file(resource, type)`
+### `delete_file(resource, type)`
+### `write_file(resource, type, data, ...)`
 
 With the specified `resource` in the scripts folder, of a specific `type`, writes/appends `data` to it, reads its
  content, or deletes the resource.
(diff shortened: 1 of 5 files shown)


James103:
The commit "Fixes #437" should either be moved to its own PR or the title of the PR should be changed to account for the additional comment.

gnembon:
Its fine as long as these tickets are closed. 

Please update docs as well for both changes though.

Also - default log level of 'info' is fine by me.

Ghoulboy78:
ok, yeah forgot the thing aboot the docs

Ghoulboy78:
btw, what are all the highlighted comments in the logger() setion about?

Ghoulboy78:
Im not sure if those are the right colours tho.

docs/scarpet/api/Auxiliary.md
```diff
@@ -245,13 +245,26 @@ Example usages:
...
 
-Prints the message to system logs, and not to chat.
+Prints the message to system logs, and not to chat. By default prints an info, unless you specify otherwise in the `type` parameter.
 
-### read_file(resource, type)
-### delete_file(resource, type)
-### write_file(resource, type, data, ...)
+Available output types:
+
+`'debug'`: white
```
gnembon, line 254: there is no need mention colors - every system could have that configured differently.

You also forgot to document the 'air' entity API

gnembon, line 254: there is no need mention colors - every system could have that configured differently.

You also forgot to document the 'air' entity API

Ghoulboy78: ofc i did

gnembon:
Forgot bout docs for `modify(e, 'air', ticks)`;P

Ghoulboy78:
ofc I did

PR #462, opened 2020-09-13
Add hollow argument to drawing cylinders, cuboids, and spheres.

kade-d:
I added the ability to draw some non-solid shapes. Although you can already hollow a shape by replacing the inside with air, this is much easier for the user. I would add this to the other shapes if approved.


gnembon:
nice idea. 
But sphere is hollow, isn't it? I think these should be hollow by default and the last argument should be optional.

kade-d:
I'll make the shapes hollow by default and the arg optional in a week or so

PR #469, opened 2020-09-15
Bots misc

samipourquoi:
Added two miscellaneous rules regarding fake players:
- `fakePlayersTeam`: creates a `fake_players` team which will get added to every fake players joining the server, and will get removed once they leave. You can't remove that team using `/team remove fake_players`
- `fakePlayersStats`: when false, it will prevent fake players from getting stats (useful for a playtime scoreboard for instance, where you don't want bots to hide every other players..)

### DIFF
--- a/src/main/java/carpet/CarpetServer.java
+++ b/src/main/java/carpet/CarpetServer.java
@@ -16,9 +16,12 @@ import carpet.utils.MobAI;
 import carpet.utils.SpawnReporter;
 import com.mojang.brigadier.CommandDispatcher;
 import net.fabricmc.loader.api.FabricLoader;
+import net.minecraft.scoreboard.Scoreboard;
+import net.minecraft.scoreboard.ServerScoreboard;
 import net.minecraft.server.command.ServerCommandSource;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.network.ServerPlayerEntity;
+import net.minecraft.util.Formatting;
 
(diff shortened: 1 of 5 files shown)


gnembon:
not see a reason to have those honestly.

samipourquoi:
I see multiple use cases for the first rule:
1) Clarity for players, to easily distinguish between bots and real players. Could even do it for shadow players I guess.
2) Could be useful to see which online player is a bot, via an external tool. For instance, a command to see who is online via a bot, which separates real players and bots, again for clarity.
3) Allows datapacks to detect if a player is a bot or not, via commands. 

The second one is just a small thing I figured would be useful. On the server I play on, the playtime scoreboard is filled with bots and I find that annoying. I am sure other people have the same issue.

I'll also note they both are part of 1.12 carpet.
I agree, these rules aren't the that useful and have very specific use cases; but I guess they would be handy for some people. May as well just make a carpet extension mod.

gnembon:
But that's the whole point of scarpet - so there is no need of modding things that should be scripted, since its mostly useful in scripts/datapacks.

You can get player type by running a command
`/script run player('Steve')~'player_type'`
which you can fetch player information including if they are fake or shadowed.

You an use that information in a script to add them to different teams to cover case 1)
cases 2) and 3) are covered directly by the fact that you can get that with a script command, and script command is part of carpet.

Feel free to make an extension for that, but an app would do the same anyways, while being much more flexible

PR #473, opened 2020-09-20
Update item_category

Ghoulboy78:
so it doesnt crash with an item which has no category.
Also sorta essential for gnembon/scarpet#133 to work.

### DIFF
--- a/src/main/java/carpet/script/api/Inventories.java
+++ b/src/main/java/carpet/script/api/Inventories.java
@@ -53,14 +53,9 @@ public class Inventories {
         expression.addLazyFunction("item_category", 1, (c, t, lv) ->
         {
             ItemStackArgument item = NBTSerializableValue.parseItem(lv.get(0).evalValue(c).getString());
-            Value res;
-
-            if(item.getItem().getGroup().getName()==null)
-                return LazyValue.NULL;
-
-            res = new StringValue(item.getItem().getGroup().getName());
-
             
+            ItemGroup ig = item.getItem().getGroup();
+            Value res = (ig==null)?Value.NULL:new StringValue(ig.getName());
             return (_c, _t) -> res;
         });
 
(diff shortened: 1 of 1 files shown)


src/main/java/carpet/script/api/Inventories.java
```diff
@@ -53,7 +53,13 @@ public static void apply(Expression expression)
         expression.addLazyFunction("item_category", 1, (c, t, lv) ->
         {
             ItemStackArgument item = NBTSerializableValue.parseItem(lv.get(0).evalValue(c).getString());
-            Value res = new StringValue(item.getItem().getGroup().getName());
+            Value res;
+            try {
+                res = new StringValue(item.getItem().getGroup().getName());
+            } catch (NullPointerException n) {
```
gnembon, line 59: catching exceptions on a regular is a baaad idea. Check if the group is null and return Value.NULL then.

src/main/java/carpet/script/api/Inventories.java
```diff
@@ -53,7 +53,14 @@ public static void apply(Expression expression)
         expression.addLazyFunction("item_category", 1, (c, t, lv) ->
         {
             ItemStackArgument item = NBTSerializableValue.parseItem(lv.get(0).evalValue(c).getString());
-            Value res = new StringValue(item.getItem().getGroup().getName());
+            Value res;
+
+            if(item.getItem().getGroup().getName()==null)
```
gnembon, line 58: well, its not the name that can be null, but the group, right?
should be 
```java
ItemGroup ig = item.getItem().getGroup();
Value res = (ig==null)?Value.NULL:new StringValue(ig.getName());
return (_c, _t) -> res;
```
right?

Ghoulboy78: But if group is null, wont getName() return null, or NullPointerException? Anyways that is faster.

gnembon (approved):

PR #482, opened 2020-09-26
Actions tests

altrisi:
PRing something to test Github Actions


altrisi:
Ok it seems to be ignored if already built

altrisi:
Wrong repo

PR #496, opened 2020-10-12
Add GitHub Actions

altrisi:
This PR adds 4 GitHub Actions to the repo.
TL;DR at the bottom.

#### What are GitHub Actions
GitHub Actions are a GitHub feature that allow you to configure automated workflows to run on their servers when specified events occur.
Those are completely free for open source/public repositories.
They store the logs of everything running in them, and support uploading artifacts from them.
#### What actions does this PR add
This PR includes three Actions. You can preview the entire Actions page of the repo at [altrisi/fabric-carpet/actions](https://github.com/altrisi/fabric-carpet/actions) (in some actions, only last run contains latest latest changes).
The included actions are the following:
##### Build on push (Development Builds)
As the name says, every time you push to the repo (including merging PRs/etc), it will build Carpet and store artifacts for anyone (registered to GitHub) to download as a development build.
This action and build on PR will only run when changes in the `src` directory are being made.

[Preview](https://github.com/altrisi/fabric-carpet/actions?query=workflow%3A%22Development+Builds%22)
##### Build on PR
Builds the content of a PR and tells you if it does actually build. It also uploads build artifacts so you (or anyone) can quickly test the changes.
Useful in order not to get any surprise when merging code where the PR author may have forgotten to add an import or any other typo that prevents things from building.
If it can't build it, it will show `All checks have failed` above the merge button in red to notify that, and will also show crosses on the commits where it failed (green checks where it didn't or nothing if didn't even run).

##### Merge docs
See comment below: https://github.com/gnembon/fabric-carpet/pull/496#issuecomment-712368008

Preview:
See [altrisi/pulls](https://github.com/altrisi/fabric-carpet/pulls), altrisi/fabric-carpet#24 or altrisi/fabric-carpet#25 ([img](https://i.imgur.com/GS8iu5m.png))
##### Merge docs
Automatically runs the mergedoc bash file when a commit modifying the `docs/` path is pushed (including when merged from a PR) and commits the changes with a meaningful enough name (`Merge docs for '[Commit message]'`).

Preview: [altrisi/commits/altrisi-mergedocsaction-test](https://github.com/altrisi/fabric-carpet/commits/altrisi-mergedocsaction-test) or altrisi/fabric-carpet#23

#### What else can you do with Actions?
In actions you can basically do anything in any [event](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows). You can even create your own "actions" to run in the GitHub runners at those specified events.
For example, one thing that could be done would be to automatically, when you create a new release in Github, set the version string in gradle.properties, build the JAR from current source, filter out the `-dev` and `-sources` JARs, upload the final one to the new release and then automatically [upload it to Curseforge ](https://github.com/itsmeow/curseforge-upload) with the given changelog, so people would stop complaining about things like `1.16.3 plz` (see carpet-extra). That would however probably change the GitHub release "scheme" followed by fabric-carpet where new versions are added into an existing release _(edit: technically possible, but probably also overkill, would remove auto version selecting and would be unstable)_.
I could probably help you with some action using pre-existing things if you want (and I have time).

#### Any extra benefits of adding these Actions?
yes. u can have cool badge

Development Builds
```
![Development Builds](https://github.com/gnembon/fabric-carpet/workflows/Development%20Builds/badge.svg)
```

#### TL;DR
Actions are a free feature to run workflows after specified events, running on GitHub servers, where you can upload results after running and basically do anything you want there. They also provide cool badges like the one above.

This adds 4 actions, one that automatically builds on push ("Development Builds), one that builds PRs to make sure they build, one that merges Scarpet docs when changes are pushed, and one that generates the Settings page automatically. They should work as-is.

--------

Sorry for 32 commits, (I'm almost sure that) some Actions only run when present in `master` (so I couldn't squash them into anything), and even though some commits were already grouped, this ended up being a lot. Just add them in a single commit (squash and merge I think it's called) to prevent an unnecessarily long list of them in the history.


gnembon:
I like those. The good things about it is that it can remove some extra checks that need to be done anyways. The good thing to do would be to print out the list of carpet rules into the wiki page with that (there is a method for that), but not sure if that's possible with github deployment scheme.

altrisi:
Adding the rules to the wiki would be relatively easy, since the wiki is basically a repo called `${{github.repository}}.wiki`, which you can even checkout or clone locally (in this case with `https://github.com/gnembon/fabric-carpet.wiki.git`). The only thing that would be needed would be to generate the md file somewhere in the action and then just push it into the wiki repo. I'd probably also filter it to only run when the `CarpetSettings.java` is modified, since it stores all rules, to prevent running actions without changes.

gnembon:
Yeah, feel free to add that - this page is automatically generated from this function:
https://github.com/gnembon/fabric-carpet/blob/e93c7d442cccb67d1b9121325b8beefe29e06659/src/main/java/carpet/settings/SettingsManager.java#L314
The best would be to add a main entrypoint in SettingsManager that would output the ruleset for a category or with all entries, and I assume that content can be piped into an .md file or file generated from that

altrisi:
It's done!
It was a bit difficult, since the classes that control the rules depend a lot in Minecraft, this needed to be ran in a Minecraft environment. Therefore, the action prepares a few classes first and then runs a server that will stop by itself after the printing of the rules is finished.
Let's see what it does:
#### Rules to wiki action
When `CarpetSettings` is modified in `master` (to prevent messing the wiki while testing in other branches), there are some files that are being generated: The `RulePrinter` class, that will run as a pre-launch entrypoint to end faster and avoid loading things like datafixers or just Minecraft itself:
```java
  package carpet.utils;

  import carpet.CarpetServer;
  import carpet.settings.SettingsManager;
  import net.fabricmc.loader.api.entrypoint.PreLaunchEntrypoint;
  import java.lang.System;

  public class RulePrinter implements PreLaunchEntrypoint {
    @Override
    public void onPreLaunch() {
        CarpetServer.onGameStarted();
        CarpetServer.settingsManager.printAllRulesToLog(null);
        System.exit(0);
    }
  }
```
After that, we need to modify the `fabric.mod.json` to tell Loader to actually run that entrypoint, so we replace Carpet's `fabric.mod.json` with a minimal one that shouldn't depend on any specific Minecraft or Fabric version, so it shouldn't need to be maintained (since MC and loader version comes from Gradle) (the version number in here doesn't need to be changed, it's just that the field is mandatory):
```json
{
    "schemaVersion": 1,
    "id": "carpet",
    "version": "1.4.11",
    "entrypoints": {
      "preLaunch": [ "carpet.utils.RulePrinter" ]
    }
}
```
Once that is done, everything is ready, so we call Gradle to `runServer` with our solution, and we ask bash to log it into a temporary file. Gradle does it's magic, and once that is done, we get a raw Gradle log containing lots of Loom-and-Gradle-related information that we don't care about, plus our list of Carpet rules. Therefore we will have to process it first.

However, first we checkout wiki into its subdirectory so we can directly replace the file in there. Then we run a small bash script to remove all that information (it will keep everything from `# Carpet` (inclusive) to `BUILD SUCCESSFUL` (exclusive) ) and we save that as our wiki page name in its repo location. Then we just go in there and make a commit. There we use the pusher's name and Github Actions bot email. That's because that way, in the main wiki page it will show the name of the pusher but if you go into revisions it will show that it was done by the bot (`github-actions[bot]` didn't look very well outside, and with pusher's mail it's impossible to distinguish manual changes from bot changes).

Since all of this happens in the Actions machine and we don't commit it back to Carpet, everything except the wiki is kept as it was when the Action is finished.

Preview:
See [altrisi/fabric-carpet/wiki/Current-Available-Settings](https://github.com/altrisi/fabric-carpet/wiki/Current-Available-Settings/e4420146c9d55def23aff8eccd98516aae8a4d48), [wiki history](https://github.com/altrisi/fabric-carpet/wiki/Current-Available-Settings/_history), [Action Run](https://github.com/altrisi/fabric-carpet/actions/runs/316007781) (ever-so-slightly outdated) and [all actions for rules-to-wiki (including old failed)](https://github.com/altrisi/fabric-carpet/actions?query=workflow%3A%22Update+wiki%22)

All of this without directly modifying Carpet's source code, because, I don't know, I thought this Github-only related code didn't need to live with useful Carpet code, and it's not extremely hard to maintain in this way in my opinion.

altrisi:
(this is ready)

gnembon:
merged, but these classes you add on the fly could very well be added directly to the source, right?

altrisi:
The RulePrinter class could be added, yes. It wouldn't cause problems since it would not be called anywhere (would be dead code in releases). But the `preLaunch` entrypoint in the `fabric.mod.json` shouldn't, since then RulePrinter would be called, instantly stopping the game after printing those rules, before even launching.

PR #497, opened 2020-10-12
Change load method to Loader's entrypoints

altrisi:
Carpet (and extensions) have been using "noop mixins" since the beggining of fabric-carpet.
Recently (mostly after the Scarpet API addition), those noop mixins to initialize extensions have been causing problems to extension developers, with up to four different commits (not including just testing without commiting) just to get those to work in some repos.

This PR changes Carpet loading method to use Fabric Loader's (not API) entrypoints.

Fabric Loader has 4 default entrypoints, 2 of them only being different in the context where they apply (Client/DedicatedServer).
Those are:
- `preLaunch`: This one occurs before Minecraft has finished loading. According to the docs, it is unsafe to use, and may break things, so we will ignore it
- `main`: This one occurs when the game is mostly ready to start, only missing some resources
- `client` and `server`: Those two occur right after `main` (next line of code), each in their context (Client/DedicatedServer)

Since Carpet extensions need to be registered before Carpet loads, this PR uses the `client` and `server` endpoints to leave the `main` entrypoint to extensions.

#### Compatibility
Since `client` and `server` entrypoints are called (ever so slightly) after Carpet's current noop mixin (next line according to a bare-bones `println` on each), old Carpet extensions using working noop mixins should continue to work after this PR without any changes.

When adapting extensions to use the main entrypoint, however, those won't be compatible with older Carpet versions running the mixin startup due to the same reason. That's why (except for the example mod), I wouldn't change current extensions (such as carpet-extra) until some uncompatible change is made to the game that forces players to update (a.k.a. Minecraft updates out of allowed range), since players tend to not update the mod until new MC version or something breaks.

After some empirical observations, it may seem that using these to load Carpet and extensions would be more compatible with other mods and extensions.

Rules, Scarpet and other things seem to work as always after some small tests (tested carpet stand-alone, with extra, with modified extra, and with some other extensions).

-------

DISCLAIMER: I'm not an extension developer (at least yet).

If this PR gets accepted and merged, I will prepare and PR changes to the carpet extension template and the main extensions.


gnembon:
not sure honestly if it fixes the issue. I think the issue was that some extension would load first, and then load carpet, which would somehow register the extension before other extensions could load. I might be wrong on that. 
The solution would be to check when adding of extensions whether extensions has been added to the settings, and if yes - force initialize that extension. This way anything could load in any order they care.

The reason I prefer the mixin method, not mod initializer, is that it adds extra fabric bureaucracy that would need to be maintained, comparing to just having your own mixin for it. Thoughts?

altrisi:
Not sure what you mean by load carpet, since those didn't run the `onGameStarted` method at that point, they just registered themselves and one created its SettingsManager, which shouldn't initialize Carpet AFAIK.

About the Fabric part, I understand your thoughts, but Fabric seems a durable loader, since it doesn't really depend in Minecraft being a specific version (someone adapted it to b1.7.3 in 3 lines, only because it didn't support the version string), and I'd say it wouldn't add much maintenance to carpet. Loader's basic APIs (referring to entrypoints) don't really change (Minecraft classes probably change more than those), and those are basically expected to be kept, and, if things change, they are at least marked and kept as deprecated for some time (quite a lot actually, there are some deprecated methods there since over a year ago).

I'd say that also makes them load "as expected", since instead of "trying to" mix into the Minecraft classes to make things load (since it could fail if, for example, Optifine suddenly decides to edit `Crashreport`, which would add maintenance), this explicitly asks Loader to load them at the specified moment, and shouldn't cause compatibility issues with other mods/unexpected behaviour, like when Carpet had to move from mains. I also think it's easier to see the "order" of things with those entrypoints rather than trying to find a place to mix into, that may at some point move or change in yarn, and can be more difficult to locate (time).

I also usually prefer to keep things minimal, but since Loader is present anyway, I think it would be ok.

This wasn't really thought as an (at least full) fix to those issues, but I thought it may help to have everything basically load at the same place.

replaceitem:
> I think the issue was that some extension would load first, and then load carpet, which would somehow register the extension before other extensions could load.

If you mean my issue with that, just to clarify on my issue i had again:
i solved it, and it was entirely my fault
i named both `mixin.json`'s in the extensions the same name (instead of `modid.mixins.json`)
so my case wasnt carpets fault at all, just mine, sorry for bothering with that in the past...

so for me, noop mixins work perfectly fine (if that is what caused this PR)

altrisi:
This last problem definitely made me finally send this PR, but the thing is before all this latest problems with the noop Mixins and extensions, I had already thought of this, a little bit before the move from mains but mostly after that, and because it also adds extra Mixins that, since Loader already provided those functionalities, I thought were adding extra files to the mod that may not have been necessary, and creating an "unneeded" complexity, all in my opinion.

FracturedCode:
@altrisi I definitely agree with the idea in principle. I recently made an extension and had a lot of difficulty making my noops work correctly. It wasn't obvious how exactly to make things load and why it worked for the example extension but not mine. Took about an hour to figure it out.
[Experience Beacons github](urlsl.me/xpBeaconsSource)

gnembon:
right. Lets then merge that and fix once in for all so that order of static loading wouldn't matter at all.

PR #503, opened 2020-10-16
Update Intellij Custom Language support

SirGavith:
Update Intellij Custom Language support 4.txt to include arguments and arrow operators for all events.


gnembon:
technically that means that you would need to use the same argument names, with the same spacing. But I can see how that can speed up typing not needing to check the args

gnembon:
can you at least fix spacing so its consistent?

SirGavith:
Thanks for letting me know, should be fixed now.

gnembon (approved):

PR #508, opened 2020-10-25
[Actions] Fix and compatibility

altrisi:
- Fixes `merge-docs` failing if there is nothing new to merge (was implemented in rules to wiki, but not for this)
   - See 7d98f023efdb2d1c628474f137a9f361a7af2094 and its [Action Run](https://github.com/gnembon/fabric-carpet/runs/1303970195?check_suite_focus=true)
- Adds "compatibility" with Gradle 6 in rules to wiki (see also gnembon/carpet-extra#143)
  - Doesn't update this Gradle to 6, just makes the action compatible
  - Also useful for any other version or change using deprecated Gradle features

### DIFF
--- a/.github/workflows/merge-docs.yml
+++ b/.github/workflows/merge-docs.yml
@@ -12,6 +12,7 @@ jobs:
       - name: Merge docs
         run: ./mergedoc.sh
       - name: Commit merged docs
+        continue-on-error: true
         run: |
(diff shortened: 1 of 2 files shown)


altrisi:
BTW is there a reason for the double spaces at the end of basically every line printed by `printAllRulesToLog`?

PR #511, opened 2020-10-26
Add chain rotation to flippinCactus

### LINKED ISSUE #509: FlippinCactus doesn't work with chains
Expected behavior: when right clicked with cactus, chains should toggle between X, Y, and Z axes.

Observed behavior: chain doesn't change on right click with cactus

This would be great to have, especially with the addition of chainStone!

fezboy:
Resolves #509

This could also easily be amended to work with all pillar blocks, which includes Hay, logs, and other blocks that face in the X, Y or Z axes.

Unfortunately the builtin rotate function on pillar blocks only switches them between X and Z axes, so this solution is a bit less elegant than calling `state.rotate`

### DIFF
--- a/src/main/java/carpet/helpers/BlockRotator.java
+++ b/src/main/java/carpet/helpers/BlockRotator.java
@@ -8,6 +8,7 @@ import net.minecraft.block.BedBlock;
 import net.minecraft.block.Block;
 import net.minecraft.block.BlockState;
 import net.minecraft.block.Blocks;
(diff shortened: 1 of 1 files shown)


Ghoulboy78:
dont worry, im implementing in a scarpet app. This will be morecustomizable, with options to keep old behavious or have new ones.

fezboy:
oh, ok then, I'll close this.

Will the scarpet app still be in the base install or relegated to carpet-extra?

PR #513, opened 2020-10-27
Replacing flippinCactus and combineXPOrbs rules with scarpet apps

Ghoulboy78:
This is because the apps are more customizable, and this matches some features requested by a few people while still being able to keep old behaviour that many are used to and more comfortable with.

Im still debating on whether to add chain blocks and other such things to the flippin cactus one, so i just didnt add them.

- [ ] Also I still gotta fix the averaging bit on the xp orbs merging, cos it's currently not taking the proper average of the position of the orbs when merging ,which is an issue I need to fix. I think it should spawn in between the orbs, like it did in old version, and with average of velocity, like in old version (Im not sure about that last one)


Ghoulboy78:
ok, its done now. Any others I should migrate to scarpet?

Ghoulboy78:
I was thinking about the /info command, but that would go to the reimplementation of commands in scarpet.

altrisi:
Why would you even do that? If the apps are going to be included anyways, what's the difference with having them as actual mod code?

I may be blind, but I don't see the benefit other than changing how things have been working for a long time. I know that, in some cases, apps have more flexibility, but I can't see, for example, the benefit in porting `combineXPOrbs` here, and with `flippinCactus` I'd say that having them as mod code actually makes them more flexibility, because there are some things you can't check directly via Scarpet that you can with Java code (like `instanceof`), you can't actually customize what is rotatable (in your script) and most `flippinCactus` code is used anyway for other rules.

There's also another thing: With Java code, the IDE will warn you when, after an update, something is just wrong, while, since there are no IDEs for Scarpet (yet), those apps could just fail silently after an update.

Ghoulboy78:
That's a question I was fearing and yet hoping to answer, as I was expecting it..

To your point of the flippincactus code being used more elsehwere, yes you would be correct, but this version allows for more customizability in terms of the method of rotation. I may also add a way to let you choose which blocks can be rotated, but we'll see.
To your point of the mod version being more flexible, I must disagree as you can very much customze what is rotatable in the app. I know this because I literally copied the java code and then squished it down to be smaller cos in Java there are different methods for different types of rotation, whereas in scarpet it's basically the same function over and over, so I just shortened.

And lastly, as to the outdating of the script, first of all I am always on hand to fix it, and gnembon is SUPER obsessive over backward compatibility, which is a good thing, and it also means that it will more than definitely work. For example, with the usage of `[]` for lists instead of `l()` (same for maps). Also like the usage of the old `store_app_data` and `load_app_data` which mean that it is unlikely to get outdated.

But thanks anyways for the question cos I had similar doubts myself and was experiencing severe emotional difficulties in telling myself that this was ok, and now I have finally struggled through my fears and decided that this is the best way. I hope you too find peace in life.

gnembon:
querying orbs every tick is inefficient - its the datapack way of doing things, which is brute force to me. I am working on more efficient way of attaching events to entities and blocks, but modding would still be preferred. /c/s have been yeaten, not because of the /cam script but because of the f3f4 introduction. 
While flippinCactus can very well be implemented as an app, I suggest leaving it be due to the past and dependency with rotator block. 

Why don't you add those to the app repo - as an alternative / learning material?

Ghoulboy78:
ok. Way to let me down. 

(jk)

Ghoulboy78:
@gnembon gnembon/scarpet#148

I hope you're happy now

Issue #221, opened 2020-04-15
better command system for scarpet


gnembon:
And types handling

something like that maybe:
in foo.sc
```
__command() -> m(
  l('', 'foo_handler'),
  l('sub', 'foo_sub_handler'),
  l('sub quux', 'quux_handler')
);
quux_handler(pos(target), block(smthing)) -> set(target, smthing);
```
would enable command like `/foo sub quux ~ ~ ~ stone`

Any suggestions? Improvements? different implementations?

gnembon:
my system would not be able to handle things like `/fill <from> <to> block` as well as `/fill <from> <to> block <replace> block` since that would mean variable number of arguments

but I don't think its very limiting

as long as you can deep dive into `/app (sub moresub subscribe) (arg arg1) ` that woudl be powerful enough

where $1 is call descriptor and $2 are args

replaceitem:
Maybe having different numbers of arguments could be solved with using NBT as the arguments?
So its a tree-like structure, and inside nbt can be set what types the arguments have and what to execute

Ghoulboy78:
```
__command()->m(
	l('0','draw','tree'),
	l('1','sphere','func',l('pos','pos'),l('int','radius'),l('block','block')),
	l('1','cone','func',l('pos','pos'),l('int','radius'),l('int','height'),l('block','block')),
	l('1','prism','tree'),
	l('2','circular',l('pos','pos'),l('int','radius'),l('int','height'),l('block','block')),
	l('2','square',l('pos','pos'),l('int','radius'),l('int','height'),l('block','block')),
)
```
How would this be?

Ghoulboy78:
The structure would be
```
l(depth_level,'parent','this_name','func/tree/literal?',args))
args: l('type','name')
```
where args is obviously there only if it is a func, and also, you can have multiple args in order, with types being similar to those found in java (or not):
```
int
str
pos(tab-completion?)
block(Like BlockStateArgument?)
others?
```
What do you think of that, @gnembon ?

altrisi:
Maybe for variable arguments there could add a third element to that list like 'variable' and pass an array to the function with all arguments instead?
Something like
```py
__command() -> m(
  l('variablesub', 'variable_handler','variable'),
);
variable_handler(arguments) -> (print(arguments) );
```
That would allow commands like printing a paginated list with the same command, like `/foo list` for first page and `/foo list [n]` for other pages without requiring different subcommands for that. It would also allow to input strings without quotes.

[...]

PR #517, opened 2020-10-29
Update commands

Ghoulboy78:


Ghoulboy78:
So basically I got rid of draw and distance and replaced with apps. That is to be ignored, as it is a sideshow. Tee main menu is coming up.

I wrote my idea down in #221 if u scroll to the bottom, but here it is in full.
Basically, if you want to define a command, you need:
1) A `__command()` function defined in your app, no changes there
2) A public function (that is, one with no preceding underscores in its name). No changes there either
3) A hidden function called `___functionName()`. It has three (3) underscores before it, not to confuse with otehr functions that people may have done before. It MUST return a map, or an error will be thrown. It may define some arguments in its map, and assign them to keywords, such as `'string'`, or `'number'`. This will make the parser inrerpret them as strings, and treat them as such. If an argument is not in the map, it will be added as a string argument.

As of now, there is uncertainty on whether the 3rd function should be optional or otherwise, as it would mess with ALL older scripts using commands. The following are arguments in favour and against the third requirement being optional or not:

Pro Optional:
Backwards compatibility

Pro Obligatory:
It means you can choose what functions become commands, and customize them more.

While I personally would advocate the Pro Obligatory side, I understand the Pro Optional side, and believe that it should therfore be decided by consensus, maybe though a poll by asking all those with @scarpet-helper role in SciCraft discord.

Ghoulboy78:
ocrap wrong branch SOZ!

PR #518, opened 2020-10-29
Update commands

Ghoulboy78:
Wanna add support for:
- [ ] block type arguments (entering it with removed minecraft: prefix ofc)
- [ ] position arguments
- [ ] boolean arguments (technically NumericValue, in practice very useful)
- [ ] entity arguments (limit of 1 entity)
- [ ] entity type arguments (entered without minecraft: prefix)
- [ ] nbt arguments
- [ ] optional arguments
- [ ] updated all apps (this one Im fine on taking help for.)
- [ ] literals

Gnembon's tasklist (also to be done):

- [ ] - handling basic argument types in commands via some standard types (pos_... etc)
- [ ] - adding option of customized types
- [ ] - adding ability of customizing command tree for apps
- [ ] - (there is already a PR to move handling of built-in commands form /script load to /carpet commandX)
- [ ] - rewrite some commands that make sense (distance, draw, perimeterinfo etc)


Ghoulboy78:
So basically I got rid of draw and distance and replaced with apps. That is to be ignored, as it is a sideshow. Tee main menu is coming up.

I wrote my idea down in #221 if u scroll to the bottom, but here it is in full.
Basically, if you want to define a command, you need:
1) A `__command()` function defined in your app, no changes there
2) A public function (that is, one with no preceding underscores in its name). No changes there either
3) A hidden function called `___functionName()`. It has three (3) underscores before it, not to confuse with otehr functions that people may have done before. It MUST return a map, or an error will be thrown. It may define some arguments in its map, and assign them to keywords, such as `'string'`, or `'number'`. This will make the parser inrerpret them as strings, and treat them as such. If an argument is not in the map, it will be added as a string argument.

As of now, there is uncertainty on whether the 3rd function should be optional or otherwise, as it would mess with ALL older scripts using commands. The following are arguments in favour and against the third requirement being optional or not:

Pro Optional:
Backwards compatibility

Pro Obligatory:
It means you can choose what functions become commands, and customize them more.

While I personally would advocate the Pro Obligatory side, I understand the Pro Optional side, and believe that it should therfore be decided by consensus, maybe though a poll by asking all those with @scarpet-helper role in SciCraft discord.

Ghoulboy78:
@altrisi would you kindly tell me the reason for that reaction? I just wanna know

altrisi:
Yes (this all is, of course, in my opinion):

As of now, changing current rules for scarpet apps has many downsides, mostly that there is no way* to autoload them if they are built-in. There is also the part of maintaining it automatically via the IDE, and the part that basically anyone who were using those as rules will suddenly find they no longer working, and won't understand what is going on (those should be probably deprecated first, just like the info command for entities).

Also I never really liked the app command system as you proposed in #221, not sure why, I just prefer the one proposed by gnembon some comments above (maybe it's just there is some part I don't fully understand about yours, idk).

BTW: Please note that I'm not in any way against Scarpet apps, I actually think that some things like the draw command may actually be better as a Scarpet app if some things were available (and of course done correctly). I've also made some apps for myself and some I'd like to publish once I finish them.

_*Yes, it can be done by autoloading a non-built-in Scarpet app that loads them, which I already have made one and will PR to the scarpet repo Soon™._

Ghoulboy78:
I understand, but the thing is with gnembon's system is that while it's undoubtedly excellent, it has a few problems:
1) It is even less backwards compatible. As you mentioned, we should make the old way deprecated for those who don't read patch notes carefullt. With that one , you can't, whereas with my one, you can. Anyways if it's not backwards compatible, they can always ask in discord, but I'm still keeping that in. Also as of now, it's backwards compatible
2) This one makes more sense in terms of java programming, and is 1000x easier to do
3) The command system as proposed in #221 was not fleshed out yet, I improved in DMs with Firigion, and there is still some stuff I have planned that I wanna do later, but haven't written here. 
4) I am aware of the bit about not being able to pre-load built-in apps, and maybe I'll look into that once this bit is done.
5) Also, that is why I put it here, so receive criticism and feedback. Thanks for that

altrisi:
You will end up forcing me to code something to make a PR before I wanted to... (in a good way for things like this)

Ghoulboy78:
unless I make the script autoload thing work, in which case there will be no need for the
> autoloading a non-built-in Scarpet app that loads them

altrisi:
To fc I mean. And yes, about the autoload, but in an extremely backwards-compatible way.

Ghoulboy78:
what needs to be backwards compatible about the autoloading of built-in scripts?

altrisi:
I had thought about adding a `scarpet` parameter to rules so apps can basically be loaded as carpet rules and would load or unload apps when they change. That way changing those wouldn't theoretically affect anyone (except if some bug is found in them ofc).

That would still leave the autoloading of built-in apps untouched, but I already decided that I'm too lazy to make that into fc anytime soon (thought about csv for that, like defaultLoggers).

Ghoulboy78:
nah man no need.

Ghoulboy78:
IT WORKS!

Ghoulboy78:
IT WOOORRKSS!!!!!

Ghoulboy78:
all thats missing is adding this compatibility to in-built apps. I can do that

Ghoulboy78:
ok, @gnembon, don't accept this pls untill the following are accomplished:
I have added support for:
- [ ] block type arguments (entering it with removed minecraft: prefix ofc)
- [ ] position arguments
- [ ] boolean arguments (technically NumericValue, in practice very useful)
- [ ] entity arguments (limit of 1 entity)
- [ ] entity type arguments (entered without minecraft: prefix)
- [ ] optional arguments
- [ ] updated all apps (this one Im fine on taking help for.)

Anything else?

altrisi:
Consider marking it as WIP

Ghoulboy78:
how? by making it a draft?

altrisi:
Yes that

Ghoulboy78:
accident

Ghoulboy78:
btw @gnembon I can't do pos cos it requires a list of some form, and there is no way to do that or it returns that lists are not allowed in invoke.

Ghoulboy78:
also, for entities so far I give UUID, which u then need to get via `entity_id()` of the input. Unfortunately, there seems to be no other way until the invoke bit is overhauled, which im not gonna do here. Maybe after the checklist is done.

Firigion:
The main downside of moving some carpet rules to scripts for stuff like draw is that you lose the autocomplete that minecraft commands have for stuff like positions (which you just adressed as not easy to tackle). Being able to do `/draw ` tab tab tab is realy useful, especialy being that you also can use `~` for position definition.

Ghoulboy78:
ye thats why I asked gnemobn to work on that

Ghoulboy78:
hey, maybe for optional arguments I should do:
```
___test()->{
    'arg1'->['type','option1','option2'],
    'arg2'->'type'
}
```
I can just check through the magic of `instanceof`

Ghoulboy78:
ok ye nbt dont work either cos it uses { and } to define, and thats not allowed in invoke

Ghoulboy78:
Ill continue tomorrow. Also having difficulty wiht eintities.

Ghoulboy78:
but imma figure out tomorrow.

altrisi:
BTW (in my opinion), a better solution for the backwards compatibility command dilemma would be to add for example in the `__config` or in `__command` something like `command_version`, running the old way when no version is specified, and not limiting when the version `1` is there. That would also allow for a future `command_version` `2` without breaking either.

That was how I thought a command system overhaul would work.

Ghoulboy78:
oh ye I was thinking of that. IT would certainly be good.

Ghoulboy78:
and nice suggestion.

Ghoulboy78:
Yes It would, that's a great idea

altrisi:
About pos arguments, have you tried using a BlockPosArgumentType? (`net.minecraft.command.argument`, like in `net.minecraft.server.command.FillCommand`) or a Vec3ArgumentType (this one I think has decimals, used in `SummonCommand`). Then converting them into a list and passing into the param.

Not sure if you are doing it this way or the problem is another, but I thought it may be helpful.

Ghoulboy78:
ye the problem was that it wouldnt let me input the three arguments as a list

Ghoulboy78:
so I had an idea of assigning to three arguments.

Ghoulboy78:
I have done nbt arguments, still uncommitted tho. Gitta shift handling of : operator safely, which is a big deal

Ghoulboy78:
ok to hell with my idea cos gnembon has a better one here: https://github.com/gnembon/fabric-carpet/issues/221#issuecomment-719924886

PR #519, opened 2020-10-29
Allow rules to manage Scarpet apps (Scarpet Rules)

### LINKED ISSUE #520: [Scarpet Suggestion] Get values of Carpet Rules
Maybe in `system_info`, basically a way to get the values of Carpet Rules.

May be useful for things like enabling or disabling a functionality (a carpet, for instance) depending on the value of a rule (the `carpets` rule, for instance), reacting to changes of vanilla behaviour that some rules may present, or decide whether or not to send updates based on `fillUpdates`.

Shouldn't throw if rule isn't found, in case it is checking for rules that may be added by extensions that aren't present.

### LINKED ISSUE #525: Rule changes always call Carpet's SettingsManager even from extensions
https://github.com/gnembon/fabric-carpet/blob/45506d22f7ae2451779eda7a44a55182f2c1bc9b/src/main/java/carpet/settings/ParsedRule.java#L159

This is currently blocking namespacing in my Scarpet `__on_carpet_rule_change()` event for #519.

Will probably fix it soon.

altrisi:
This PR adds the ability to define Scarpet apps in the `@Rule` annotation, and those will be enabled when the rule is enabled, and obviously disabled when disabled. This opens possibilities like replacing current Carpet rules or commands (like in #518) without the "hassle" for the user to change from rule to app, etc. It also mantains backwards compatibility with the rules users had before the change, allows to have both Java coded and Scarpet coded content in the same rule, and of course this also gets printed into the list of Current Available Settings and can be easily found in the `/carpet` command.

Extensions should also be able to register their own "Scarpet rules".

Working vs not working things:

- [x] Basic functionality: Rules load and unload apps when changed on the fly
- [x] Commands from Scarpet Rules are registered (when rule is edited on the fly)
- [x] Make sure apps and commands work from setDefaults (~~currently they don't register, probably called too early~~)
- [x] Check if rule boolean defaulting to true would load the script (~~currently it relies on the `set()` method, so if it doesn't get called with defaults the apps won't be loaded~~ now startup is deferred. It works)
- [x] Check that rule boolean defaulting to true wouldn't crash everything because of trying to load an app while unloading (~~`set()` is called to reset defaults if I recall correctly, and that is called on unload~~ it doesn't, and the CarpetScriptServer is killed anyways)
- [x] Test Scarpet events: They work correctly, according to `event_test`
- [x] Remove the app loaded message from Scarpet Rules (with something like what already exists in the unload function)
- [x] Get Scarpet rule's apps out of the unload command (the unload command reads `scriptServer.modules`, which contains them since loaded rule apps aren't separated in ScriptServer. Moving them out of there makes them not editable from `/script`, removing some of their versatility)
- [x] Maybe accept non-boolean rules to allow saving and editing some behaviour changes easily (like `int > 0` -> app loaded, then in app check that int, or even `string != "false"`, world of customization)
- [x] Add something like #520 to make previous point more useful and check command perms (Resolves #520)
- [x] Check compatibility with Command category (since it changes things): Should be compatible once strings are supported (would require a way to read rule value to fully work)

Feel free to suggest more points, give feedback and/or help with some. ~~Until now, I've only tested this with the overlay app and event_test as rules.~~ I've basically tested every built-in app at this point.

Extra:
- Fixes #525: Now changing a rule will call the appropriate code from the SettingsManager that registered the rule instead of always calling Carpet's. That means rule observers will also be separated from each manager, so in order to register an observer for all extensions the extension should register their observers at each extension's SettingsManager (probably never going to happen anyway).

### DIFF
--- a/src/main/java/carpet/script/CarpetEventServer.java
+++ b/src/main/java/carpet/script/CarpetEventServer.java
@@ -654,7 +654,7 @@ public class CarpetEventServer
                 );
             }
         };
-        public static final Event CARPET_RULE_CHANGE = new Event("carpet_rule_change", 2, true)
+        public static final Event CARPET_RULE_CHANGES = new Event("carpet_rule_changes", 2, true)
         {
         	@Override
         	public void onCarpetRuleChange(ParsedRule<?> rule, ServerCommandSource source)
--- a/src/main/java/carpet/script/CarpetEventServer.java
+++ b/src/main/java/carpet/script/CarpetEventServer.java
@@ -11,6 +11,7 @@ import carpet.script.value.NBTSerializableValue;
 import carpet.script.value.NumericValue;
 import carpet.script.value.StringValue;
 import carpet.script.value.Value;
+import carpet.settings.ParsedRule;
 import carpet.utils.Messenger;
 import net.minecraft.block.BlockState;
 import net.minecraft.entity.damage.DamageSource;
@@ -694,6 +695,25 @@ public class CarpetEventServer
                 );
             }
         };
+        public static final Event CARPET_RULE_CHANGE = new Event("carpet_rule_change", 2, true)
+        {
+        	@Override
+        	public void onCarpetRuleChange(ParsedRule<?> rule, ServerCommandSource source)
+        	{
+        		String identifier = rule.settingsManager.getIdentifier();
+        		final String namespace;
+        		if (!identifier.equals("carpet")) 
+        		{
+        			namespace = identifier+":";
+        		} else { namespace = "";}
+        		handler.call(
+        				() -> Arrays.asList(
+        						((c,t) -> new StringValue(namespace+rule.name)),
+        						((c,t) -> new StringValue(rule.getAsString()))
+        				), () -> source
+        		);
+        	}
+        };
 
         // on projectile thrown (arrow from bows, crossbows, tridents, snoballs, e-pearls
 
@@ -733,6 +753,7 @@ public class CarpetEventServer
 
         public void onWorldEvent(ServerWorld world, BlockPos pos) { }
         public void onWorldEventFlag(ServerWorld world, BlockPos pos, int flag) { }
+        public void onCarpetRuleChange(ParsedRule<?> rule, ServerCommandSource source) { }
     }
 
 
--- a/src/main/java/carpet/script/CarpetEventServer.java
+++ b/src/main/java/carpet/script/CarpetEventServer.java
@@ -656,22 +656,22 @@ public class CarpetEventServer
         };
         public static final Event CARPET_RULE_CHANGES = new Event("carpet_rule_changes", 2, true)
         {
-        	@Override
-        	public void onCarpetRuleChange(ParsedRule<?> rule, ServerCommandSource source)
-        	{
-        		String identifier = rule.settingsManager.getIdentifier();
-        		final String namespace;
-        		if (!identifier.equals("carpet")) 
-        		{
-        			namespace = identifier+":";
-        		} else { namespace = "";}
-        		handler.call(
-        				() -> Arrays.asList(
-        						new StringValue(namespace+rule.name),
-        						new StringValue(rule.getAsString())
-        				), () -> source
-        		);
-        	}
+            @Override
+            public void onCarpetRuleChanges(ParsedRule<?> rule, ServerCommandSource source)
+            {
+                String identifier = rule.settingsManager.getIdentifier();
+                final String namespace;
+                if (!identifier.equals("carpet")) 
+                {
+                    namespace = identifier+":";
+                } else { namespace = "";}
+                handler.call(
+                        () -> Arrays.asList(
+                                new StringValue(namespace+rule.name),
+                                new StringValue(rule.getAsString())
+                        ), () -> source
+                );
+            }
         };
 
         public static String getLoadEvent(EntityType<? extends Entity> et)
@@ -741,7 +741,7 @@ public class CarpetEventServer
 
         public void onWorldEvent(ServerWorld world, BlockPos pos) { }
         public void onWorldEventFlag(ServerWorld world, BlockPos pos, int flag) { }
-        public void onCarpetRuleChange(ParsedRule<?> rule, ServerCommandSource source) { }
+        public void onCarpetRuleChanges(ParsedRule<?> rule, ServerCommandSource source) { }
     }
 
 
--- a/src/main/java/carpet/script/CarpetEventServer.java
+++ b/src/main/java/carpet/script/CarpetEventServer.java
@@ -667,8 +667,8 @@ public class CarpetEventServer
         		} else { namespace = "";}
         		handler.call(
         				() -> Arrays.asList(
-        						((c,t) -> new StringValue(namespace+rule.name)),
-        						((c,t) -> new StringValue(rule.getAsString()))
+        						new StringValue(namespace+rule.name),
+        						new StringValue(rule.getAsString())
         				), () -> source
         		);
         	}
--- a/src/main/java/carpet/script/CarpetScriptServer.java
+++ b/src/main/java/carpet/script/CarpetScriptServer.java
@@ -194,7 +194,8 @@ public class CarpetScriptServer
         boolean reload = false;
         if (modules.containsKey(name))
         {
-            removeScriptHost(source, name, false, isRuleApp);
+            if (isRuleApp) return false;
+        	removeScriptHost(source, name, false, isRuleApp);
             reload = true;
         }
         Module module = isRuleApp ? getRuleModule(name) : getModule(name, false);
--- a/src/main/java/carpet/script/CarpetScriptServer.java
+++ b/src/main/java/carpet/script/CarpetScriptServer.java
@@ -88,7 +88,12 @@ public class CarpetScriptServer
     public void initializeForWorld()
     {
     	worldInitialized = true;
-    	CarpetServer.settingsManager.initializeScarpetRules(server.getCommandSource());
+    	CarpetServer.settingsManager.initializeScarpetRules();
+    	CarpetServer.extensions.forEach(e -> {
+    		if(e.customSettingsManager() != null) {
+    			e.customSettingsManager().initializeScarpetRules();
+    		}
+    	});
         if (CarpetSettings.scriptsAutoload)
         {
             Messenger.m(server.getCommandSource(), "Auto-loading world scarpet apps");
--- a/src/main/java/carpet/script/CarpetScriptServer.java
+++ b/src/main/java/carpet/script/CarpetScriptServer.java
@@ -94,7 +94,7 @@ public class CarpetScriptServer
             Messenger.m(server.getCommandSource(), "Auto-loading world scarpet apps");
             for (String moduleName: listAvailableModules(false))
             {
-                addScriptHost(server.getCommandSource(), moduleName, true, true);
+                addScriptHost(server.getCommandSource(), moduleName, true, true, false);
             }
         }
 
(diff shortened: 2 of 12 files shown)


altrisi:
This is how it would work to register a Scarpet rule:

- Register the rule script somewhere (for example, with other bundled scripts at CarpetScriptServer):

```java
CarpetScriptServer.registerRuleScript(BundledModule.carpetNative("overlay", false));
```

- In CarpetSettings, add a new Rule with the `scarpetApp` parameter in the `@Rule` annotation being the name of the rule app:

```java
    @Rule(
    		desc = "Enables seeing structure bounding boxes",
    		extra = {
    				"Use /overlay to get started"
    		},
    		scarpetApp = "overlay",
    		category = {FEATURE, "scarpet"} // Probably needs its own RuleCategory (either changing 
                                                // SCARPET purposes or something along those lines)
    )
    public static boolean commandOverlay = false;

```

- Now you should see the `commandOverlay` rule under `/carpet`, and should be able to load and unload the app by changing that rule.

gnembon:
nice idea - to streamline apps and custom commands

altrisi:
2 tasks (and extra testing) to go! Should be ready soon.

Anything else I should do or change?

To get them out from the unload commands I'll probably be making a second `modules` map in ScriptServer (something like `unloadableModules`, maybe only an array with names) and check from there, also `removeScriptHost` will be limited to not allow unloading rule apps except when specified by a new param.

About the way to get the value of the rule in the app itself, I was thinking in maybe not adding that in this PR, or only inject the rule name as a global in a dirty way. I may end up doing #520, however.

gnembon:
removing built-in apps completely from /script makes sense. I sometimes use /script run to debug things, but with finished products you should not need to do that

altrisi:
Oh, when I thought about that was for example editing globals in the apps, like the `update_interval` in `chunk_display` or other configuration options that may make sense to be changed at some cases, but aren't important enough to make a command or use the rule value for them.

BTW coordinated comments, lol.

gnembon:
I think removing ability to load / unload / reload makes 100% sense - with the other usecases like /script in <app> run - I would leave it. The thing is  - you have to have proper permissions to use these, and they can be disabled, so it is still possible to 100% sandbox scripts (disable script command, or not give players perms to run it), yet they could be using built-in apps via their commands and interactions.

altrisi:
Should I add an event like `on_rule_change`? Also what params should I pass to it?

altrisi:
Oops, I broke Scarpet.

altrisi:
This did work in previous build and in local dev, weird.

altrisi:
Ready to test and review.

altrisi:
Ok I found something: `getBoolValue` always returns false with ints because it seems they aren't assignable from Number.class?

altrisi:
Fixed that last problem (in some way) and everything seems to work.

I've also tested in some different ways and didn't find bugs.

I'd say it's ready to merge, you can test and review the changes. Feel free to give any feedback.

src/main/java/carpet/settings/ParsedRule.java
```diff
@@ -10,26 +10,43 @@
...
     public boolean isClient;
+    public boolean hasScarpet;
     public final Class<T> type;
     public final List<Validator<T>> validators;
     public final T defaultValue;
     public final String defaultAsString;
+    public final SettingsManager settingsManager;
+    private static final Set<Class<?>> NUMBER_CLASSES;
+    static {
+        Set<Class<?>> s = new HashSet<>();
```
gnembon, line 39: What about Sets.newHashSet from guava?

src/main/java/carpet/settings/ParsedRule.java
```diff
@@ -50,6 +69,11 @@
                 this.validators.add((Validator<T>) callConstructor(Validator._COMMAND_LEVEL_VALIDATOR.class));
             }
         }
+        if (!scarpetApp.isEmpty())
+        {
+            this.hasScarpet = true;
```
gnembon, line 74: isn't hasScarpet litttle redundant? should be sufficient to set `this.scarpetApp = null` and use that instead

src/main/java/carpet/settings/ParsedRule.java
```diff
@@ -105,7 +129,7 @@ else if (this.type.isEnum())
 
     public ParsedRule<T> set(ServerCommandSource source, String value)
     {
-        if (CarpetServer.settingsManager != null && CarpetServer.settingsManager.locked)
+        if (settingsManager != null && settingsManager.locked)
```
gnembon, line 132: adding reference to setting manager is a good change - this way each setting group can be managed independently

src/main/java/carpet/settings/ParsedRule.java
```diff
@@ -156,7 +180,7 @@ else if (type.isEnum())
             if (!value.equals(get()) || source == null)
             {
                 this.field.set(null, value);
-                if (source != null) CarpetServer.settingsManager.notifyRuleChanged(source, this, stringValue);
+                if (source != null) settingsManager.notifyRuleChanged(source, this, stringValue);
```
gnembon, line 183: not so sure about this one - this means that extensions will miss rule updates form non-`carpet` managers. Not sure if this is good, since the settings manager can be fetched per rule via 'this' send as a param

src/main/java/carpet/script/CarpetEventServer.java
```diff
@@ -653,6 +654,25 @@ public void onWorldEventFlag(ServerWorld world, BlockPos pos, int flag)
...
+                String identifier = rule.settingsManager.getIdentifier();
+                final String namespace;
+                if (!identifier.equals("carpet")) 
+                {
+                    namespace = identifier+":";
+                } else { namespace = "";}
+                handler.call(
+                        () -> Arrays.asList(
+                                new StringValue(namespace+rule.name),
+                                new StringValue(rule.getAsString())
```
gnembon, line 671: can technically fetch the proper 'value' (NumericValue for numbers etc), but I am not sure if you would want to merge more carpet (modded) code with scarpet (that 'pretends' is just a missing vanilla piece and assumes no mods exist)

src/main/java/carpet/script/CarpetScriptServer.java
```diff
@@ -150,17 +185,20 @@ public ScriptHost getHostByName(String name)
         return modules.get(name);
     }
 
-    public boolean addScriptHost(ServerCommandSource source, String name, boolean perPlayer, boolean autoload)
+    public boolean addScriptHost(ServerCommandSource source, String name, boolean perPlayer, boolean autoload, boolean isRuleApp)
     {
         //TODO add per player modules to support player actions better on a server
+        if (!worldInitialized) return false;
```
gnembon, line 191: Can you explain why you want to prevent adding apps before world is initialized?

gnembon:
A few comments:
Only a few comments in the ParsedRule class.
1) notifyRuleChanged on 'global' Carpet settings app - this way there is still one point to add yourself for rule snooping
2) isScarpet is reduntant if scarpetApp is set to null
3) world initialized - why making sure that no app can load before world starts? This should not be a problem for normal apps

Rest of the comments in code is just notes.

gnembon:
The reason the rules were loaded before world was loading was for some rules that needed to take an effect before chunks. This was important for example for kelpGenerationGrowthLimit, which is no more since its already fixed in vanilla. It makes sense that rules can load before, but also makes sense that apps start a bit later. Maybe.

src/main/java/carpet/script/CarpetScriptServer.java
altrisi: Basically rules for a world are loaded (world is attached from CarpetServer) before the ScriptServer is reset, and the attachServer calls `set()` for most rules, if not all. Since rule apps are toggled from `set()`, that basically makes rule apps to "randomly" load or unload in that period, and not start if coming from setDefaults or from being their default value. Also for some reason it made rule apps that were loaded in that period from setDefault to be unusable until world reload without them enabled as default. That fixed it. Also that autoloaded apps are loaded in the same moment.
https://github.com/gnembon/fabric-carpet/blob/c827b041262dde181961c185dea4430f72bb2991/src/main/java/carpet/CarpetServer.java#L72-L78
https://github.com/gnembon/fabric-carpet/blob/a9e7bdce15ba825def2016b3f92869562e56c791/src/main/java/carpet/CarpetServer.java#L84-L88
(not sure if the name of that moment would really be `worldInitialized`, it's basically at `onServerLoadedWorlds` in CarpetServer)

src/main/java/carpet/settings/ParsedRule.java
altrisi: Well, I'd say when most extensions register their rule observers they usually only want to get changes from their rules. Anyway, extensions can still register observers for other managers if later than extensions load they add their observers by checking for extensions at `CarpetServer.extensions`.

Because also else that would make each manager having their own observer list rather pointless, since those would never be called.

src/main/java/carpet/settings/ParsedRule.java
altrisi: Well, not actually `null`, because it seems that `null` can't be a default for an annotation, but will check for `empty()`.

This was just the way I started because I found `isClient`, but it's true that it's quite redundant, since it's only used in two places.

src/main/java/carpet/settings/ParsedRule.java
altrisi: True, forgot Minecraft included guava

src/main/java/carpet/settings/ParsedRule.java
altrisi: I've checked in every extension [here](https://github.com/gnembon/fabric-carpet/wiki/List-of-Carpet-extensions) and all (the only) rule observers I found only care about their rule changes.

src/main/java/carpet/script/CarpetEventServer.java
altrisi: I decided to let it always a StringValue since you can convert those/check what they are easily with Scarpet functions. Also there would be some "fake types", like `COMMAND` rules usually being booleans, but can also be the `ops` String and the numbers `0-4` for perms, so it may be counter-intuitive to find a COMMAND `false` as a String but the `carpets` `false` as a boolean. Also the code is way simpler.

src/main/java/carpet/script/CarpetScriptServer.java
altrisi: Wait. Actually that check may be completely useless, let me recheck.

altrisi: Yes, it is completely useless. My reasoning was to prevent it loading/unloading apps on SettingsManager reset, but totally forgot that the ScriptServer is killed _after_ that reset, not before, so the check and variable is pointless.

However, here there is still the discussion about whether rule apps should load at `onServerLoadedWorlds` (with autoloaded apps) or `onServerLoaded` (with regular rules)

src/main/java/carpet/settings/ParsedRule.java
altrisi: If you really want, maybe having 2 observer types? (global/per manager)
Or making observers static among all SettingsManager?

altrisi:
I left replies in the comments.

src/main/java/carpet/script/CarpetEventServer.java
gnembon: Yeah, that makes sense - the client should know what value types to expect, and rules values can be more flexible in this regard than native scarpet values (bool+'ops'+1-4). Lets leave it strings

src/main/java/carpet/settings/ParsedRule.java
gnembon: I get that having it 'carpet' instance specific may not be intuitive with custom managers, but I get the appeal of one global one. Lets do what you suggested - having a static one shared between managers, and for each manager. Each manager would notify all static observers as well as their own. This way you can subscribe to what you want

PR #544, opened 2020-11-14
Javadocs, cleanup, extension's client rules, built-in BundledModules generators, etc

### LINKED ISSUE #235: verify switching between SP worlds resets  hopper counters properly.

altrisi:
This PR does the following (this ended up doing more things than it intended to do in the first place):

- Adds javadocs to many methods (mostly those that are most used in extensions). Those javadocs contain their params, returns, some internal links, etc
   Feel free to suggest more or changes if I messed up what does something do
  
- Adds built-in ways to produce `BundledModule`s outside the Carpet native scope:
   - `fromPath`, creates one from a specified path and name
   - `fromPathWithCustomName`, creates one from a full path, but with a custom name
   - The methods in `BundleModule` now depend on each other to generate modules (aka `carpetNative` calls `fromPath`, and `fromPath` calls `fromPathWithCustomName`)
- Adds Eclipse's `.launch` files to gitignore (those are used to launch dev envs)
- Allows extensions to have client-side rules in their `SettingsManager`s (untested, should work) (carpet-less only due to #546)
- Removes around 50 unused imports (most from the last cleanup, didn't remove all since some are used in commented code)
   Not tested against 1.17 branch, should work
   Some numbers:
   According to Eclipse, before these changes there were `327` warnings in the project. After this PR, there are "only" `267` (most from raw types or casts, also I should disable serializable warns). That's `-60` warns plus the extra features.
- Removes a piece of dead and unused code around app loading, that I ended up annoyed of always finding it when working on Scarpet Rules
- Changes maximum tick warp duration to `Integer.MAX_VALUE`, since there is no reason to limit that (is there?)
- Moves `RulePrinter` (from the `rules-to-wiki` action) into the source so it can be managed from the IDE (as suggested in https://github.com/gnembon/fabric-carpet/issues/496#issuecomment-714730868)
- Fixes #235 (no, it didn't reset them AFAIK). Now they get reset even when entering the same world, though
- Allow merge-docs to run even if `Full.md` was modified (aka prefer merging the files than trusting the commit, since it will change next time anyway and will be ignored if the result is the same)

### DIFF
--- a/src/main/java/carpet/commands/TickCommand.java
+++ b/src/main/java/carpet/commands/TickCommand.java
@@ -35,7 +35,7 @@ public class TickCommand
                                 executes((c) -> setTps(c.getSource(), getFloat(c, "rate"))))).
                 then(literal("warp").
                         executes( (c)-> setWarp(c.getSource(), 0, null)).
-                        then(argument("ticks", integer(0,4000000)).
+                        then(argument("ticks", integer(0,Integer.MAX_VALUE)).
                                 suggests( (c, b) -> suggestMatching(new String[]{"3600","72000"},b)).
                                 executes((c) -> setWarp(c.getSource(), getInteger(c,"ticks"), null)).
                                 then(argument("tail command", greedyString()).
--- a/src/main/java/carpet/commands/TickCommand.java
+++ b/src/main/java/carpet/commands/TickCommand.java
@@ -35,7 +35,7 @@ public class TickCommand
                                 executes((c) -> setTps(c.getSource(), getFloat(c, "rate"))))).
                 then(literal("warp").
                         executes( (c)-> setWarp(c.getSource(), 0, null)).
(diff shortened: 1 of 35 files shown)


src/main/java/carpet/commands/TickCommand.java
```diff
@@ -35,7 +35,7 @@ public static void register(CommandDispatcher<ServerCommandSource> dispatcher)
                                 executes((c) -> setTps(c.getSource(), getFloat(c, "rate"))))).
                 then(literal("warp").
                         executes( (c)-> setWarp(c.getSource(), 0, null)).
-                        then(argument("ticks", integer(0,4000000)).
+                        then(argument("ticks", integer(0,Integer.MAX_VALUE)).
```
gnembon, line 38: you technically only need to specify the min
```suggestion
                        then(argument("ticks", integer(0)).
```

gnembon (approved):

PR #553, opened 2020-11-16
Replacing draw and distance with apps

Ghoulboy78:
Replaced commandDraw and commandDistance with their own apps, with new command support and evrything. Also, added some new things such as much needed funcs to math.scl, and that's about it. Also fixes #462 ig


altrisi:
Sadly you didn't see the PR I did to your repo so you actually added the apps. I've opened a new one at Ghoulboy78#18.

Also some suggestions: The hollow thing, I'd make it so instead of true/false, which can be difficult to understand if brigadier doesn't tell you what that is (and only says true/false), be a string either hollow or whatever the best for the opposite is (filled? idk).

Also can't draw be global? Since it doesn't really save any per-player info, so it only exists once.

src/main/resources/assets/carpet/scripts/distance.sc
```diff
@@ -0,0 +1,55 @@
...
+        if(has(global_positions,player()),
+            calculate(end_pos),
+            print(player(),format('r There was no initial point for '+player()));
+            set_start(end_pos)
+        )
+    )
+);
+
+__on_player_places_block(player, item_tuple, hand, block) ->(
+    if(system_info('world_carpet_rules'):'carpets'&&block=='brown_carpet',
```
altrisi, line 31: `world_carpet_rules` always returns strings, so you need a string equals `'true'` check instead.

Ghoulboy78: ah

Ghoulboy78: Ill test.

Ghoulboy78: but tomorrow

altrisi:
Gnembon, does Scarpet have that thing from Java that in an `if` statement, if the first condition fails and it is followed by `&&` it doesn't process the second one?
It's because I can't remember how costly was `world_scarpet_rules`, to suggest adding it after checking the block maybe to prevent running it on every block placed.

Ghoulboy78:
o ye scarpet has that. so Ill swap. but tomorrow

gnembon:
I suggest you don't delete the java stuff for now, just add the apps, so we can run them a few days in parallel to compare them.

Regarding && - the other operands are not executed if they encounter the first fail.

gnembon:
@altrisi I suggest apps to stay player if needed. It this case it doesn't matter, but in the future if there is some animations added to custom fills, or maybe some client assist using client side shapes then it should be a player app then. All in all - doesn't matter that much.

src/main/resources/assets/carpet/scripts/draw.sc
```diff
@@ -0,0 +1,196 @@
...
+
+__config() -> {
+    'commands'->{
+        'sphere <center> <radius> <block>'->['draw_sphere',null, false],
+        'sphere <center> <radius> <block> replace <replacement>'->['draw_sphere',false],
+        'ball <center> <radius> <block>'->['draw_sphere', null, true],
+        'ball <center> <radius> <block> replace <replacement>'->['draw_sphere', true],
+        'diamond <center> <radius> <block>'->['draw_diamond',null],
+        'diamond <center> <radius> <block> replace <replacement>'->'draw_diamond',
+        'cone <center> <radius> <height> <pointing> <orientation> <block> <hollow>'->['draw_pyramid',null, false],
```
gnembon, line 11: you can even replace initial selection of shapes into a term argument <shape> - that would reduce it to two lines

src/main/resources/assets/carpet/scripts/draw.sc
```diff
@@ -0,0 +1,196 @@
...
+        'cylinder <center> <radius> <height> <orientation> <block> <hollow> replace <replacement>'->['draw_prism', false],
+        'cuboid <center> <radius> <height> <orientation> <block> <hollow>'->['draw_prism', null, true],
+        'cuboid <center> <radius> <height> <orientation> <block> <hollow> replace <replacement>'->['draw_prism', true]
+    },
+    'arguments'->{
+        'center'->{'type'->'pos', 'loaded'->'true'},
+        'radius'->{'type'->'int', 'suggest'->[]},//to avoid default suggestions
+        'replacement'->{'type'->'block'},
+        'height'->{'type'->'int', 'suggest'->[]},
+        'orientation'->{'type'->'string', 'suggest'->['x','y','z']},
```
gnembon, line 25: type term is what you are looking for

src/main/resources/assets/carpet/scripts/draw.sc
```diff
@@ -0,0 +1,196 @@
...
+        'pyramid <center> <radius> <height> <pointing> <orientation> <block> <hollow>'->['draw_pyramid',null, true],
+        'pyramid <center> <radius> <height> <pointing> <orientation> <block> <hollow> replace <replacement>'->['draw_pyramid', true],
+        'cylinder <center> <radius> <height> <orientation> <block> <hollow>'->['draw_prism', null, false],
+        'cylinder <center> <radius> <height> <orientation> <block> <hollow> replace <replacement>'->['draw_prism', false],
+        'cuboid <center> <radius> <height> <orientation> <block> <hollow>'->['draw_prism', null, true],
+        'cuboid <center> <radius> <height> <orientation> <block> <hollow> replace <replacement>'->['draw_prism', true]
+    },
+    'arguments'->{
+        'center'->{'type'->'pos', 'loaded'->'true'},
+        'radius'->{'type'->'int', 'suggest'->[]},//to avoid default suggestions
```
gnembon, line 22: min 0?

src/main/resources/assets/carpet/scripts/draw.sc
```diff
@@ -0,0 +1,196 @@
...
+        'replacement'->{'type'->'block'},
+        'height'->{'type'->'int', 'suggest'->[]},
+        'orientation'->{'type'->'string', 'suggest'->['x','y','z']},
+        'pointing'->{'type'->'string','suggest'->['up','down']},
+        'hollow'->{'type'->'bool'}
```
gnembon, line 27: since its a last arument - just leave it as a literal, right?

src/main/resources/assets/carpet/scripts/draw.sc
```diff
@@ -0,0 +1,196 @@
...
+);
+
+length_sq(vec)->return(_vec_length(vec)^2);//cos of lengthSq func in DrawCommand.java
+
+fill_flat(pos, offset, dr, rectangle, orientation, block, hollow, replacement)->(
+    successes = 0;
+    r = floor(dr);
+    drsq = dr^2;
+    if(orientation=='x',
+        c_for(a=-r,a<=r,a+=1,
```
gnembon, line 57: these double forloops cannot be replaced by more efficient scans?

src/main/resources/assets/carpet/scripts/draw.sc
```diff
@@ -0,0 +1,196 @@
...
+        nextYn = 0;
+        c_for(y=0,y<=ceilRadiusY,y+=1,
+            yn = nextYn;
+            nextYn = (y + 1) / radiusY;
+            nextZn = 0;
+            c_for(z=0,z<=ceilRadiusZ,z+=1,
+                zn = nextZn;
+                nextZn = (z+1)/radiusZ;
+
+                if(length_sq([xn, yn, zn]) > 1,break());
```
gnembon, line 117: try not to use breaks / continues especially in the innermost loop

src/main/resources/assets/carpet/scripts/draw.sc
Ghoulboy78: what is that?

src/main/resources/assets/carpet/scripts/draw.sc
Ghoulboy78: tru

src/main/resources/assets/carpet/scripts/draw.sc
Ghoulboy78: last arg is replacement...

src/main/resources/assets/carpet/scripts/draw.sc
Ghoulboy78: First of all, u did that (copied directly from ur code), second of all, I little other way to do it.

src/main/resources/assets/carpet/scripts/draw.sc
Ghoulboy78: u do realise that scan() function does the same thing right?

Ghoulboy78: It is literally a triple for loop, isn't it?

Ghoulboy78:
I added the commands back. I wonder if it will compile...

Ghoulboy78:
Ok it did, cool. Basically I re-added draw and distance commands, and renamed apps to draw_rc_1.sc and distance_rc_1.sc
Tho tbh I don't see the necessity as in the case of draw, it has literally the same code, and in the case of distance, it has better, less convoluted (in my opinion) code.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
```diff
@@ -0,0 +1,55 @@
...
+            calculate(end_pos),
+            print(player(),format('r There was no initial point for '+player()));
+            set_start(end_pos)
+        )
+    )
+);
+
+__on_player_places_block(player, item_tuple, hand, block) ->(
+    if(block=='brown_carpet' && system_info('world_carpet_rules'):'carpets'=='true',
+        if(player~'sneaking',
```
altrisi, line 32: If the player isn't sneaking, it should calculate with previous start pos (or set start pos if not present), right?

src/main/resources/assets/carpet/scripts/draw.sc
altrisi: Anyway, hollow could be a string to make it simpler to understand when brigadier wants you to not see things (and doesn't tell you the arg name, just true/false) (`hollow`|`(whatever is the opposite to that and isn't similar to other args)`)

src/main/resources/assets/carpet/scripts/draw.sc
altrisi: AFAIK, it's because in Scarpet `return` calls (and probably `break` calls too) are expensive.

altrisi:
Poor #462, waiting for 2 months without getting merged and with this its target will just be killed. Progress.
(why didn't it ever get merged btw?)

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
```diff
@@ -0,0 +1,55 @@
...
+        'from <start_pos>'->'set_start',
+        'from <start_pos> to <end_pos>'->'_calculate',
+        'to <end_pos>'->_(end_pos)->(_calculate(null,end_pos))
+    }
+};
+
+global_positions={};
+
+set_start(pos)->(
+    global_positions:player() = pos;
```
altrisi, line 14: Since app is player scope, is there really a need to save it in a map with the player's name?

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
Ghoulboy78: ye, thats what it does mate.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
Ghoulboy78: true, I forgot that bit.

src/main/resources/assets/carpet/scripts/draw.sc
Ghoulboy78: I get that, but that doesn't change the fact that I see little way around it.

Ghoulboy78: I'm open to suggestions

src/main/resources/assets/carpet/scripts/draw.sc
Ghoulboy78: ah ok I see.

src/main/resources/assets/carpet/scripts/draw.sc
Ghoulboy78: thx

src/main/resources/assets/carpet/scripts/draw.sc
Ghoulboy78: wait how? they have differing arg numbers.

altrisi:
Suggestion: Replacing default draw and distance normal rules with the apps so they are used and tested, and keep the old commands at `commandLegacyDraw` (or `legacyCommandDraw`, as `/legacydraw`) disabled by default for around 3-5 months before finally replacing them.

Not sure what does gnembon think about that.

src/main/resources/assets/carpet/scripts/draw.sc
altrisi: Maybe wrapping everything into some big `if` statements and do nothing if the condition isn't met, for the break (if I'm not mistaken and breaks exits the entire loop) maybe set a variable that is reset outside the loop to true and check in the same `if`. Don't know if that's actually better or worse (didn't test it), but just an idea.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
altrisi: Oops, read it wrong. Now after actually testing: Maybe not send an error if no initial pos was available on first carpet.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
```diff
@@ -51,5 +64,15 @@ calculate(end_pos)->(
         'w  - Spherical: ', 'w '+spherical+'\n',
         'w  - Cylindrical: ', 'w '+cylindrical+'\n',
         'w  - Manhattan: ', 'w '+manhattan
```
altrisi, line 66: Parity: In old command, distance numbers are in bold. Not sure if making them bold or what, but at least something to make it different.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
```diff
@@ -51,5 +64,15 @@ calculate(end_pos)->(
...
-    ))
+    ));
+
+    schedule(0,'_draw_line',end_pos);
+);
+
+_draw_line(end_pos)->(
+    if(global_show_line,
+        draw_shape('line',100,'from',global_position,'to',end_pos, 'player',player()),
```
altrisi, line 74: This is almost invisible in non-carpet clients (or "not carpet client enabled" servers) since without Carpet in the client, the particles for the line are only generated once per call (probably a bug I should have reported).

That's why other built-in apps usually have a lot lower durations for shapes.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
```diff
@@ -0,0 +1,78 @@
...
+        calculate(end_pos),
+        if(global_position,
+            calculate(end_pos),
+            print(player(),format('r There was no initial point for '+player()));
+            set_start(end_pos)
+        )
+    )
+);
+
+__on_player_places_block(player, item_tuple, hand, block) ->(
```
altrisi, line 43: Another parity thing: The old distance command would also calculate distance without the carpet actually being placed, only when right-clicking, no matter if it got placed or not (like right clicking a wall).

Basically changing it to `player_right_clicks_block` and finding the position by face.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
```diff
@@ -0,0 +1,78 @@
...
+    null
+);
+
+_calculate(start_pos, end_pos)->(
+    if(start_pos!=null,
+        set_start(start_pos);
+        calculate(end_pos),
+        if(global_position,
+            calculate(end_pos),
+            print(player(),format('r There was no initial point for '+player()));
```
altrisi, line 37: Maybe not print this if it's the player's first carpet (again parity, and I think looks nicer) (should still be shown ofc when calculating normally).

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
Ghoulboy78: Ye, overlay uses delay 15. Problem is then it can get laggy, so I made it refresh each second.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
Ghoulboy78: tru dat

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
Ghoulboy78: ye I hadnt read old code properli

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
Ghoulboy78: It send error message and registers that as first pos.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
Ghoulboy78: veri fair point

src/main/resources/assets/carpet/scripts/draw_rc_1.sc
```diff
@@ -0,0 +1,196 @@
...
+//"Boilerplate" code
+
+set_block(x, y, z, block, replacement)-> (
+    success=0;
+    if(block != block(x, y, z),
+        set([x, y, z],block);
+        success+=1;
+
+        //todo add check for carpet setting CarpetSettings.fillUpdates
+        update(block(x, y, z))
```
altrisi, line 45: Scarpet's `set()` function already sends updates if `fillUpdates` is `true`, so removing the update() call is the same as checking fillUpdates (maybe even better, prevent double update).

Ghoulboy78: ah k nice, but Im afraid I just checked in the code and no bueno.

altrisi: @Ghoulboy78

Ghoulboy78: what?

altrisi: Sorry your comment didn't appear but the commit did, thought you didn't see it

altrisi: BTW `carpet_rules` is _way_ too expensive to be called for every single block being set (according to `profile_expr` in my computer, you can only run it 2100 times in a tick before taking >50ms, and that is without counting the rest of the code)

src/main/resources/assets/carpet/scripts/draw.sc
altrisi: Wow you can only really use `break()` 750 times a tick in my computer, same for `continue()`.

Empty `if` can, however, be ran 107k times.

Ghoulboy78: oof wow Imma change that

altrisi:
Found a way to reduce overhead of draw_sphere to about 66% (via changing breaks & continue to big ifs). May PR you soon.

src/main/resources/assets/carpet/scripts/draw.sc
altrisi: Did it and made a PR to your repo at Ghoulboy78#19

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
```diff
@@ -0,0 +1,79 @@
...
+        'w  - Cylindrical: ', 'wb '+cylindrical+'\n',
+        'w  - Manhattan: ', 'wb '+manhattan
+    ));
+
+    schedule(0,'_draw_line',end_pos);
+);
+
+_draw_line(end_pos)->(
+    if(global_show_line,
+        draw_shape('line',20,'from',global_position,'to',end_pos, 'player',player()),
```
altrisi, line 75: This would probably need `start_pos` too, since start position may change in a future and then the line would just get weird, right?

altrisi:
Suggestion (not 100% sure about it):

Maybe adding a text (`label`) with the distance(s) above the line? What do you think? May get messy with many lines around, and may be difficult to identify what line is which. Maybe assigning a random color to the line and text, instead of always white? Then some lines _may_ be invisible over some specific (the very rare completely solid colored) blocks, but I don't think it would be _that_ bad.

I think the label would at least look really good for carpets, not so sure about commands (same with line though)

Ghoulboy78:
ummm, @altrisi , this is what a sphere with ur changes looks like: 
image

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
Ghoulboy78: tru

Ghoulboy78:
ok ye I redid draw_sphere. I replaced with a new function that a) works, b)has the same ms count, and way less than the previous one which I copied from gnembon's code.

src/main/resources/assets/carpet/scripts/draw.sc
Ghoulboy78: Done

Ghoulboy78: ah wait thats fill_flat func, didnt do dat.

src/main/resources/assets/carpet/scripts/draw.sc
Ghoulboy78: ye I tried, and id dont work.

src/main/resources/assets/carpet/scripts/draw.sc
Ghoulboy78: ok did it now.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
Ghoulboy78: I ould have to make it so it stops when it sees that the position is changed.

Ghoulboy78:
This is messed up... It compiled for me.

altrisi:
Wow, at the begining it worked for me. Probably in the end I changed something without testing that made it go really wrong. Sorry about that.

About the build failing: Don't worry, the error will be gone next push. It seems that connection to external servers failed while preparing loom.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
```diff
@@ -0,0 +1,82 @@
...
+        'w  - Manhattan: ', 'wb '+manhattan
+    ));
+
+    global_end_position = end_pos;
+
+    schedule(0,'_draw_line',start_pos,end_pos);
+);
+
+_draw_line(start_pos,end_pos)->(
+    if(global_show_line && global_position == start_pos && global_end_position == end_pos,//so if it changes, it will stop drawing this line.
```
altrisi, line 77: Oh, thought having multiple lines was a feature and not a bug, lol. I was quite used to it. nvm.

However, this new change will still create up to infinite lines if the distance is calculated against the same two points over and over (which may seem unlikely, but if someone wants to forcefully lag/crash a server, they just have to spam `/distance to [pos]`).

Would suggest maybe not scheduling a new line on creation but just reuse the same schedule and use the globals instead of locals to draw it. You would have to start the schedule at some point, maybe starting and stopping when running `toggle_line` (starting is scheduling, stopping is not rescheduling, there's still no way to unschedule in Scarpet sadly).
It could still be abused if really willing to, but it's more difficult (to prevent this, maybe make the schedule toggle another bool off and only schedule a new one if that variable got false).

An implementation of the safe one would be like (please don't trust any of the following code more than a bad reference, first of all `global_` being `g_`):
```py
_toggle_line -> (
  if(g_draw_line,
    g_draw_line = false;
  ,!g_draw_line
    g_draw_line = true;
    if(g_stopped,schedule(0,'_draw_line');
    g_stopped = false;
  )
);

draw_line -> (
  if(g_draw_line,
    draw_shape('line',20,'from',global_start_pos,'to',global_end_pos, 'player',player());
    schedule(20,'_draw_line');
  ,
    g_stopped = true
  )
)
```
This one would make sure that the schedule is actually stopped before creating a new one, because else toggle_line could be called multiple times before the schedule actually stops, creating a new one and keeping the existing one since draw would still be on.

src/main/resources/assets/carpet/scripts/draw.sc
Ghoulboy78: @gnembon any suggestions?

gnembon: yeah, different arg numbers - need to stay this way.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
```diff
@@ -0,0 +1,87 @@
...
+    ));
+
+    global_end_position = end_pos;
+
+    schedule(0,'_draw_line');
+);
+
+_draw_line()->(
+    if(global_show_line,//It will always draw from the global positions, so it will turn on if there is a line
+        draw_shape('line',20,'from',start_pos,'to',end_pos, 'player',player());
```
gnembon, line 84: is start_pos and end_pos not defined here?
Would be cool to show a label with a distance.

gnembon: I should add anchors to each point spec, like from, to - in this case from is static (world position) and to should be player anchored. Otherwise it would be little wobbly

src/main/resources/assets/carpet/scripts/draw_rc_1.sc
```diff
@@ -0,0 +1,137 @@
...
+            )
+        ),
+        print(player(),format('r Error while running command: orientation can only be "x", "y" or "z", '+orientation+' is invalid.'));
+        global_affected = 0;
+    );
+);
+
+//drawing commands
+
+draw_sphere(centre, radius, block, replacement, hollow)->(
```
gnembon, line 87: you know that we will need to rewrite the hollow sphere from this from O(n^3) to a O(n^2) algorithm that is currently in java for draw...

gnembon: but we can do that later.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
Ghoulboy78: umm, could u explain that pls?

src/main/resources/assets/carpet/scripts/draw_rc_1.sc
Ghoulboy78: u do realise that ur algorithm for the DrawCommand.java has O(n^3) right...

Ghoulboy78: im resarcing rasterization algorithms rn.

gnembon: true, that should use rather trig functions over angles, roughly the way I generate sphere meshes in shapes at https://github.com/gnembon/fabric-carpet/blob/7645312856c1c5970bd5b47970c07a381fc75c75/src/main/java/carpet/script/utils/ShapesRenderer.java#L851
which is definitely rendering at 2D.

Don't worry about that here - do whatever is valid and we will continue to work on it as it goes.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
gnembon: I am not sure where start_pos and end_pos take from, since its not used before in here.

Ghoulboy78: oh shit I forgot. lemme change that real quick

src/main/resources/assets/carpet/scripts/draw_rc_1.sc
Ghoulboy78: image

Ghoulboy78: I believe that this justifies the bruteforce code. Best I could come up with was caching already printed shapes, so u can spam same shape over and over and it's very fast.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
Ghoulboy78: ah oops.

Ghoulboy78: done.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
```diff
@@ -0,0 +1,88 @@
...
+set_start(pos)->(
+    global_position= pos;
+    print(player(),format('gi Initial point set to: ', 'g '+pos,'?/tp '+(str(pos)-','-'['-']')))
+);
+
+_toggle_line()->(
+    global_show_line= !global_show_line;
+    if(global_show_line,
+        print('Toggled lines on');
+        if(global_position && global_end_position,_draw_line()),//making sure that neither of them are null
```
altrisi, line 32: This would cause the message to be sent without the line being actually scheduled for next time. Also the variable would have changed too, so next time you run the command it would first say "turned off" and you would need to run it again.

I'd suggest to check those either when drawing (more flexible) or before even toggling the variable.

Also you only need to check for `end_position` since you never set it without start.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
```diff
@@ -0,0 +1,88 @@
...
+);
+
+_calculate(start_pos, end_pos)->(
+    if(start_pos!=null,
+        set_start(start_pos);
+        calculate(end_pos),
+        if(global_position,
+            calculate(end_pos),
+            print(player(),format('r There was no initial point for '+player()));
+            set_start(end_pos)
```
altrisi, line 45: Wait, why do you set start position when given an end one? Should either notify the player you are doing that or not do it at all. _EDIT: The old one does send the notification_

~Also, is there a need say "...initial point **for [player]**"? And in past?~ _EDIT: Nvm found it's parity_

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
Ghoulboy78: nah, it works good. Ill take that last suggestion tho, tru.

src/main/resources/assets/carpet/scripts/distance_rc_1.sc
Ghoulboy78: I do it cos in the case that that if statement is false, it meant that function received no start pos AND there is no pre-saved start pos, so it would fail if I didnt do that. And I do notify player that im setting the start pos, in the `start_pos()` function at line 23

Issue #546, opened 2020-11-14
Extension's rules aren't syncronized with clients


altrisi:
Currently, `ServerNetworkHandler` only sends rules from Carpet's `SettingsManager`.
https://github.com/gnembon/fabric-carpet/blob/15239ecb0f92879196ff383c22a5f5e0003db73c/src/main/java/carpet/network/ServerNetworkHandler.java#L80

This basically means that rules from other `SettingsMangers` will never* be received by clients from remote servers.

There are various ways to solve this:
- Send all rules to clients by appending them with the current `withRule` method and then loop through all `SettingsManager` instances in the client (probably performance intensive in the client if extensions have many rules/many custom managers, but would keep compatibility)
- Change the protocol so it also sends the `SettingsManager` `identifier` with each rule (or another way to identify it) and find `SettingsManager` by that identifier in the client. May, however, probably break when connecting old clients/servers (would it?).

Maybe a good way would be to add the latest to the 1.17 branch (if it breaks), since most versions are basically incompatible with the previous. _May_ provide a PR at some point

_* Technically, those are sent in the `updateRuleWithConnectedClients` method, since it is called from any`SettingsManager`, but those aren't properly added in the client since it only checks if that rule exist in Carpet's._

Fallen-Breath:
I'll suggest to keep compatibility. There are people who use client to connect to servers in different minecraft versions / carpet versions by mods or plugins or whatever things. If it's not backwards compatible it will restrict the client carpet version that people can use.

Performance issue should be fine since even rule synchronization is not something that will happen quite often.

Fallen-Breath:
Ah wait, why would extension have its own SettingsManager. Extension only need to let carpet's setttingManager to parse its own setting class, which will add its all ruls into carpet's setting manager. For example [this](https://github.com/gnembon/carpet-extra/blob/c402a39ecf208311ba243f7a5f6778950c5d2055/src/main/java/carpetextra/CarpetExtraServer.java#L34-L35) is how carpet extra does:

```java
// let's /carpet handle our few simple settings
CarpetServer.settingsManager.parseSettingsClass(CarpetExtraSettings.class);
```

It's also the same in [fabric-carpet-extension-example-mod](https://github.com/gnembon/fabric-carpet-extension-example-mod/blob/32f25e50202826d44334136e958fb8864c5ac042/src/main/java/carpet_extension/ExampleExtension.java#L26-L27)

[...]

PR #555, opened 2020-11-17
[Scarpet] Add display_title function

altrisi:
_Title_ says it all.

Works with both `StringValue` and `FormattedTextValue`, and supports custom timings.

See docs for more info.

Should I mention in docs that times are per-player and are (AFAIK) reset when player rejoins?

Also that `subtitle` requires to have a running `title` to be shown?

### DIFF
--- a/src/main/java/carpet/script/api/Auxiliary.java
+++ b/src/main/java/carpet/script/api/Auxiliary.java
@@ -459,25 +459,21 @@ public class Auxiliary {
             if (lv.size() < 2) throw new InternalExpressionException("'display_title' needs at least a target, type and message, and optionally times");
             Value pVal = lv.get(0).evalValue(c);
             TitleS2CPacket.Action action;
-            List<Value> targets;
+            List<ServerPlayerEntity> targets;
             Text title = null;
             int in, out, stay;
+            MinecraftServer server = ((CarpetContext)c).s.getMinecraftServer();
             if (pVal instanceof ListValue)
             {
-                targets = ((ListValue) pVal).getItems();
-                for (Value player:targets)
-                {
-                    if (!(player instanceof EntityValue && ((EntityValue) pVal).getEntity() instanceof PlayerEntity))
-                        throw new InternalExpressionException("'display_title' only takes players or player lists as first argument");
-                }
-            }
-            else if (pVal instanceof EntityValue && ((EntityValue) pVal).getEntity() instanceof PlayerEntity)
-            {
-                targets = Arrays.asList(pVal);
+            	targets = ((ListValue) pVal).getItems().stream()
+            			  .map(v -> EntityValue.getPlayerByValue(server, v))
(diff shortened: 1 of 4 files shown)


Ghoulboy78:
I don't see the point, tbh. Why not just add to that pr and add to the print() function? the framework is very much present.

altrisi:
#### About actionbar being there without currently being customizable

It's there because the vanilla title command has it, and that probably means that at some time (in the far future) it will support those. Also it can be done with the same packet.

#### About sending 2 packets for timings (and not resetting to default)

Minecraft requires 2 packets for that (sadly, tested) even though the packet has all the required params, and trying to reset to default would just be guessing, since client could have received a change from a `/title times`, and it will also reset anyway when player rejoins (AFAIK, since it's not saved anywhere).

altrisi:
Because print() would have a lot of arguments if you add times

altrisi:
BTW gnembon could you reopen #546?
It wasn't fixed by that PR, that PR only allowed the fully client-side rules (with `inspectClientSideCommand`, like `fogOff`), and it mentioned that it only worked in carpet-less servers because of this.

gnembon:
Looks good. The only thing I would request is change the name - `send_title` is unclear what is 'sending' - its all packets in the game anyways, for instance `draw_shape` is also sending something to the client to let them draw it. More clear names would be 'display_title', `show_title`, `title_text`, `display_text`. 
`Title` can stay in the name only because the vanillla command that controls it is 'title'

altrisi:
Renamed it and all its references to `display_title()`.

src/main/java/carpet/script/api/Auxiliary.java
```diff
@@ -452,6 +455,81 @@ else if (!interactable && targetBlock == null)
...
+            TitleS2CPacket.Action action;
+            List<Value> targets;
+            Text title = null;
+            int in, out, stay;
+            if (pVal instanceof ListValue)
+            {
+                targets = ((ListValue) pVal).getItems();
+                for (Value player:targets)
+                {
+                    if (!(player instanceof EntityValue && ((EntityValue) pVal).getEntity() instanceof PlayerEntity))
```
gnembon, line 470: should be able to accept player names as valid entries as well, You can use 
```
if (pVal instanceof ListValue) targets = pVal.getItems() else targets = SingletonList.of(pVal):
List<ServerPlayerEntity> players = targets.stream().map(v -> EntityValue.getPlayerByValue(.....)).collect() - > that should take care of validation
... proceeed
```

gnembon:
`display_title` sounds better
also please that display_title to editor support files for IDEA and np++

gnembon:
looks good so far. Small details, but should be fine :)

altrisi:
Changes done. Should be ready for review again.

gnembon (approved):

PR #559, opened 2020-11-21
Added team functions

replaceitem:
Added the following functions to deal with teams:

team_list() -> list teams => returns list of teams
team_list(team) -> list players in team => returns list of players, or null if team not found

team_add(team) -> create team => returns team name if sucessfull, otherwise null, if team already exists
team_add(team,player) -> join player to team	=> returns true if player was added, false if player is already in team

team_remove(team) -> removes team => returns true if team was removed, null if team didnt exist

team_leave(player) -> leaves player out of his team	=> returns true if player was in a team, false if not

team_empty(team) -> remove all players from team => returns number of players that were in the team, null if team not found

team_modify(team,property) -> get team property => returns value of property, null if team not found
team_modify(team,property,value) -> modify team property with value	=> returns true, null if team not found

I hope i implemented it right, if not, i can to go over it again and change that
Also i wasn't sure if i added the functions correctly to the notepad++ editor files


altrisi:
Suggestions:

Since `team_modify()` also gets properties, maybe renaming it to `team_property()` or something similar.

For docs, maybe put the examples right below the function docs so they can be seen faster, else when for instance looking at `team_list()` you can't see their examples. Also some maybe don't need examples (such as `team_list()` with no params, and other simple ones).

Maybe also "compressing" some of the properties docs, since they use quite a lot of space (like placing all options in a line instead of one for each).

Also I'd recommend you not to use master branch for changes, it can sometimes mess some things up (like merge docs, which should have only ran when merging) and it's also easier to have a forked master to track. (now if your changes get squashed you get your master with your commits ahead and the squashed one behind).

replaceitem:
Yeah makes sense to rename it, moving examples is a good idea too, and some are not needed for sure. Compressing docs is also good.
About the branches, im sorry if i did something wrong, im not that experience with git sadly, do you mean i should do changed on my fork on a seperate branch?

altrisi:
Don't worry, the git branches are ok. It's just for next time (s), it's usually more convinient for you after merging.

gnembon:
`team_empty(t)` seems redundant as for `team_empty(t) -> for(team_list(t), team_leave(t, _))`

+1 on team_modfiy -> team_property

Also everything else in scapet is snake case - I can see that these values need to be camel since vanilla accepts them this way - looks little wierd

replaceitem:
Yeah, snake case is a good idea, ill change that.
Also, ill probably close the PR and start a new one, since i just made a whole mess with my branches and commits (like docs already merged).

PR #569, opened 2020-11-22
Added scarpet team functions

replaceitem:
**Continuation of #559**

*(new branch)*

Added the following functions to deal with teams:

team_list() -> list teams => returns list of teams
team_list(team) -> list players in team => returns list of players, or null if team not found

team_add(team) -> create team => returns team name if sucessfull, otherwise null, if team already exists
team_add(team,player) -> join player to team => returns true if player was added, false if player is already in team

team_remove(team) -> removes team => returns true if team was removed, null if team didnt exist

team_leave(player) -> leaves player out of his team => returns true if player was in a team, false if not

team_empty(team) -> remove all players from team => returns number of players that were in the team, null if team not found

team_modify(team,property) -> get team property => returns value of property, null if team not found
team_modify(team,property,value) -> modify team property with value => returns true, null if team not found

### DIFF
--- a/docs/scarpet/resources/editors/npp/scarpet_v1.7_notepad++.xml
+++ b/docs/scarpet/resources/editors/npp/scarpet_v1.7_notepad++.xml
@@ -60,6 +60,8 @@
 
                 scoreboard scoreboard_add scoreboard_remove scoreboard_display
 
+                team_add team_list team_remove team_leave team_empty team_property
+
                 player spawn entity_id entity_list entity_area entity_selector query modify entity_event
 
                 stack_limit item_category recipe_data crafting_remaining_item
--- a/docs/scarpet/resources/editors/npp/scarpet_v1.7_notepad++.xml
+++ b/docs/scarpet/resources/editors/npp/scarpet_v1.7_notepad++.xml
@@ -60,7 +60,7 @@
 
                 scoreboard scoreboard_add scoreboard_remove scoreboard_display
 
-                team_add team_list team_remove team_leave team_empty team_property
+                team_add team_list team_remove team_leave team_property
 
                 player spawn entity_id entity_list entity_area entity_selector query modify entity_event
 
--- a/src/main/java/carpet/script/api/Scoreboards.java
+++ b/src/main/java/carpet/script/api/Scoreboards.java
@@ -284,7 +284,7 @@ public class Scoreboards {
             if(!(propertyVal instanceof StringValue)) throw new InternalExpressionException("'team_property' requires a string as the second argument");
 
             switch (propertyVal.getString()) {
-                case "collisionRule":
+                case "collision_rule":
                     if(!modifying) {
                         Value ret = new StringValue(team.getCollisionRule().name);
                         return (_c, _t) -> ret;
@@ -292,9 +292,18 @@ public class Scoreboards {
 
                     if(!(settingVal instanceof StringValue)) throw  new InternalExpressionException("'team_property' requires a string as the third argument for the property " + propertyVal.getString());
 
-                    AbstractTeam.CollisionRule collisionRule = AbstractTeam.CollisionRule.getRule(settingVal.getString());
-                    if(collisionRule == null) throw new InternalExpressionException("Unknown value for property " + propertyVal.getString() + ": " + settingVal.getString());
-                    team.setCollisionRule(collisionRule);
+                    switch(settingVal.getString()) {
+                        case "always":
+                            team.setCollisionRule(AbstractTeam.CollisionRule.ALWAYS); break;
+                        case "never":
+                            team.setCollisionRule(AbstractTeam.CollisionRule.NEVER); break;
+                        case "push_other_teams":
+                            team.setCollisionRule(AbstractTeam.CollisionRule.PUSH_OTHER_TEAMS); break;
+                        case "push_own_team":
+                            team.setCollisionRule(AbstractTeam.CollisionRule.PUSH_OWN_TEAM); break;
+                        default:
+                            throw new InternalExpressionException("Unknown value for property " + propertyVal.getString() + ": " + settingVal.getString());
+                    }
                     break;
                 case "color":
                     if(!modifying) {
@@ -309,7 +318,7 @@ public class Scoreboards {
                     team.setColor(color);
 
                     break;
-                case "deathMessageVisibility":
+                case "death_message_visibility":
                     if(!modifying) {
                         Value ret = new StringValue(team.getDeathMessageVisibilityRule().name);
                         return (_c, _t) -> ret;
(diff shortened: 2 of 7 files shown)


docs/scarpet/resources/editors/npp/scarpet_v1.7_notepad++.xml
```diff
@@ -60,6 +60,8 @@
 
                 scoreboard scoreboard_add scoreboard_remove scoreboard_display
 
+                team_add team_list team_remove team_leave team_property
```
gnembon, line 63: this is technically not need - its 1.7 whatever that was - its just for posterity.

src/main/java/carpet/script/api/Scoreboards.java
```diff
@@ -147,5 +154,295 @@ public static void apply(Expression expression)
...
             return (_c, _t) -> new NumericValue(slot);
         });
+
+        expression.addLazyFunction("team_list", -1, (c, t, lv) ->
+        {
+            if(lv.size() > 1) throw new InternalExpressionException("'team_list' requires zero or one parameters");
+            CarpetContext cc = (CarpetContext)c;
+            ServerScoreboard scoreboard = cc.s.getMinecraftServer().getScoreboard();
+            if(lv.size() == 0) {
+                Value ret = ListValue.wrap(scoreboard.getTeamNames().stream().map(StringValue::new).collect(Collectors.toList()));
```
gnembon, line 164: I suggest you use StringValue::of - it handles better null values if encounters one. Going forward that would be the default way of getting these.

src/main/java/carpet/script/api/Scoreboards.java
```diff
@@ -147,5 +154,295 @@ public static void apply(Expression expression)
...
+                return (_c, _t) -> ret;
+            } else if(lv.size() == 1) {
+                Value teamVal = lv.get(0).evalValue(c);
+                if(teamVal instanceof StringValue) {
```
gnembon, line 168: I suggest not being that strict here. use .getString() on any value. If team '123' exists, that's fine I would say.

src/main/java/carpet/script/api/Scoreboards.java
```diff
@@ -147,5 +154,295 @@ public static void apply(Expression expression)
...
+
+
+        expression.addLazyFunction("team_add", -1, (c, t, lv) ->
+        {
+            if(!(lv.size() < 3 && lv.size() > 0)) throw new InternalExpressionException("'team_add' requires one or two parameters");
+
+            CarpetContext cc = (CarpetContext)c;
+            ServerScoreboard scoreboard = cc.s.getMinecraftServer().getScoreboard();
+            Value teamVal = lv.get(0).evalValue(c);
+            if(!(teamVal instanceof StringValue)) throw new InternalExpressionException("'team_add' requires a string as the first argument");
```
gnembon, line 189: same - be lenient with types, just use .getString()

src/main/java/carpet/script/api/Scoreboards.java
```diff
@@ -147,5 +154,295 @@ public static void apply(Expression expression)
...
+                    scoreboard.addTeam(teamName);
+                    Value ret = new StringValue(teamName);
+                    return (_c, _t) -> ret;
+                } else {
+                    return LazyValue.NULL;
+                }
+            } else if(lv.size() == 2) {
+                Value playerVal = lv.get(1).evalValue(c);
+                String player;
+                if(playerVal instanceof StringValue) {
```
gnembon, line 204: EntityValue::getPlayerByValue should handle all the cases for you - unless you want to cover usecases where its a non-logged player. In that case I suggest adding a new function that looks similar EntityValue::getPlayerNameByValue that does the same thing, but honors nonexisting players and return String, not the ServerPlayerEntity instead. - that function would be handy in several other places.

src/main/java/carpet/script/api/Scoreboards.java
```diff
@@ -147,5 +154,295 @@ public static void apply(Expression expression)
...
+                return LazyValue.NULL;
+            }
+        });
+
+        expression.addLazyFunction("team_remove", 1, (c, t, lv) ->
+        {
+            CarpetContext cc = (CarpetContext)c;
+            ServerScoreboard scoreboard = cc.s.getMinecraftServer().getScoreboard();
+            Value teamVal = lv.get(0).evalValue(c);
+            if(!(teamVal instanceof StringValue)) throw new InternalExpressionException("'team_remove' requires a string as the first argument");
```
gnembon, line 231: same, just ignore it.

src/main/java/carpet/script/api/Scoreboards.java
```diff
@@ -147,5 +154,295 @@ public static void apply(Expression expression)
...
+        });
+
+
+        expression.addLazyFunction("team_leave", 1, (c, t, lv) ->
+        {
+            CarpetContext cc = (CarpetContext)c;
+            ServerScoreboard scoreboard = cc.s.getMinecraftServer().getScoreboard();
+            Value playerVal = lv.get(0).evalValue(c);
+            String player;
+            if(playerVal instanceof StringValue) {
```
gnembon, line 251: you could use that ::getPlayerNameByValue here as well.

src/main/java/carpet/script/api/Scoreboards.java
```diff
@@ -147,5 +154,295 @@ public static void apply(Expression expression)
...
+            Value propertyVal = lv.get(1).evalValue(c);
+
+            Value settingVal = null;
+            boolean modifying = false;
+            if(lv.size() == 3) {
+                modifying = true;
+                settingVal = lv.get(2).evalValue(c);
+            }
+
+            if(!(teamVal instanceof StringValue)) throw new InternalExpressionException("'team_property' requires a string as the first argument");
```
gnembon, line 279: ping

src/main/java/carpet/script/api/Scoreboards.java
```diff
@@ -147,5 +154,295 @@ public static void apply(Expression expression)
...
+
+            Team team = scoreboard.getTeam(teamVal.getString());
+            if(team == null) return LazyValue.NULL;
+
+            if(!(propertyVal instanceof StringValue)) throw new InternalExpressionException("'team_property' requires a string as the second argument");
```
gnembon, line 284: pong

gnembon:
Hmm, checked the /team command - everything there is Camel as well - therefore I think we should stick to camel (everything else in commands is snake as well, so if Mojang is onboard with that inconsistentcies, we have no reason to do otherwise).

gnembon:
Also please remove all that strictness about StringValues - .getString() of anything should be just fine.

If there are more string-like types in the future that would stop working

gnembon:
tl;dr; just a few nitpick details, but PR looks good and ready
Thanks for helping :)

replaceitem:
Alright thanks dor looking over it!
So should i put back the camel case for the properties and it's values?

replaceitem:
Also @gnembon i wasn't sure if i should just return null if the team wasn't found or throw an exeption?
The scoreboard functions throw an error, but not sure if that should be the case for the team functions too...

gnembon:
You are correct - objective names are not predefined and can be set arbitrarily by the players, so passing a non-existing objective shoudn't result in an exception, but in null or something like that. Wrong criterion on the other side - since its one of fixed options, if incorrect should fail the program. I corrected that.

Regarding camelCasing - its part of vanilla commands, and even if not super consistent, should be preserved. That should simplify setting values based on enums.

gnembon:
Hope that was clear from my previous post - since vanilla uses camel case in these cases - lets keep that throughout here. Will merge that right after.

replaceitem:
I already replaced it again with camelCase in 49b994b21ce1fbc1c13b94075d7c09fff8e59d78
However i just looked over it and if forgot to change it in the docs, thats now done too, so it should be ready now!
Thanks for the help!

gnembon (approved):

PR #585, opened 2020-11-28
Added scarpet bossbar function

replaceitem:
Added a `bossbar()` function for functionality like with `/bossbar` command in vanilla
For syntax, see the [docs](https://github.com/replaceitem/carpet-discarpet/releases)


src/main/java/carpet/script/api/Scoreboards.java
```diff
@@ -396,6 +402,132 @@ public static void apply(Expression expression)
...
         });
+
+        expression.addLazyFunction("bossbar", -1, (c, t, lv) ->
+        {
+            BossBarManager bossBarManager = ((CarpetContext)c).s.getMinecraftServer().getBossBarManager();
+            if(lv.size() > 3) throw new InternalExpressionException("'bossbar' accepts max three arguments");
+            if(lv.size() == 0) {
+                Value ret = ListValue.wrap(bossBarManager.getAll().stream().map(CommandBossBar::getId).map(Identifier::toString).map(StringValue::of).collect(Collectors.toList()));
+                return (_c, _t) -> ret;
+            } else if(lv.size() == 1) {
```
gnembon, line 413: since you have a return in the previous id statement, this can be replaced with `if`, not `else if`
This will make so that getting the identifier for a bossbar from this else if and next else if can be extracted and shared.

src/main/java/carpet/script/api/Scoreboards.java
```diff
@@ -396,6 +402,132 @@ public static void apply(Expression expression)
...
+                }
+                String property = lv.get(1).evalValue(c).getString();
+                CommandBossBar bossBar = bossBarManager.get(identifier);
+                Value propertyValue;
+                if(lv.size() == 3) {
+                    propertyValue = lv.get(2).evalValue(c);
+                } else {
+                    propertyValue = null;
+                }
+                if(bossBar == null) return LazyValue.NULL;
```
gnembon, line 440: this can be called right after fetching the bossbar, skipping evaluating lv.get(1) and lv.get(2)

gnembon:
a few nitpicks, otherwise good :)

replaceitem:
Definetly agree, it was a real mess before, sorry for making you read that, should be cleaner now

gnembon (approved):

PR #590, opened 2020-11-29
Allow loading apps from a global path

### LINKED ISSUE #577: Add the ability to store global user scripts
I love scarpet and use it whenever I want to test something in creative. The problem is every script I write ends up in the specific world save folder. 

There are already built-in scripts provided by carpet that you can load from any world. Would be very nice to have a place to store a user-created library of scripts, e.g. in `~/.minecraft/scripts/` folder.

altrisi:
This is an implementation that would resolve #577.

It works as defined in https://github.com/gnembon/fabric-carpet/issues/577#issuecomment-735025354 (global scripts would basically be stored at `[root]/config/carpet/scripts`, root being either `.minecraft` or the server's root).

I've also taken the time to change some things (can revert if you want), basically changing `File` -> `Path` everywhere related to this (that is, in `listAvailableModules`, `getModule` and `FileModule`). Therefore, it instead uses Streams to do the work of finding the appropriate script or to list those. Did this because many of those methods already start with a `Path` that they convert to a `File`, require a `Path` that they convert from a `File` or even do both steps. And because you introduced me into the magic of Streams.

This implementation would automatically create the `[root]/config/carpet/scripts` and `[world]/scripts` folders automatically when being called (that is, either when searching for an app under `/script load` or when loading a world with `scriptsAutoload`). Since using either of those usually means you would like to use Scarpet in that world, I think it would be ok to create them (makes them discoverable and user was probably already searching for them), but if you want I can change it in order to not create them (by wrapping it into `if(Files.exists(path))`, maybe for the global one at least on servers).

### DIFF
--- a/src/main/java/carpet/script/CarpetScriptServer.java
+++ b/src/main/java/carpet/script/CarpetScriptServer.java
@@ -16,19 +16,24 @@ import com.mojang.brigadier.arguments.StringArgumentType;
 import com.mojang.brigadier.builder.LiteralArgumentBuilder;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 import com.mojang.brigadier.tree.CommandNode;
+
+import net.fabricmc.loader.api.FabricLoader;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.command.ServerCommandSource;
 import net.minecraft.server.network.ServerPlayerEntity;
 import net.minecraft.util.WorldSavePath;
 import net.minecraft.util.math.BlockPos;
 
-import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


altrisi:
Well, maybe I do it like in config, just do it in client.

gnembon:
autocreating the folder is fine - otherwise it might be pain to create it. We can do the same with the scripts folder - its now only created automatically if any built-in apps want to store some files.

altrisi:
I mean in the server, don't automatically create the `config/carpet/scripts` folder, since those scripts can just be in the world (or maybe don't even check the `config/carpet/scripts` folder at all in servers, since `FabricLoader#getConfigDir` would create an empty `config` dir in carpet-only servers).

gnembon:
Also if you could edit the docs to say where to put these in SP, that would be perfect

gnembon:
There is also a wiki page about installing scripts, but that bit can be updated later

src/main/java/carpet/script/CarpetScriptServer.java
```diff
@@ -173,17 +188,23 @@ public Module getRuleModule(String name)
...
-            return moduleNames;
-        for (File script : listOfFiles)
-        {
-            if (script.getName().endsWith(".sc"))
+        try {
+            Path worldScripts = server.getSavePath(WorldSavePath.ROOT).resolve("scripts");
+            Files.createDirectories(worldScripts);
+            Files.walk(worldScripts)
+                .filter(f -> f.toString().endsWith(".sc"))
+                .forEach(f -> moduleNames.add(f.getFileName().toString().replaceFirst("\\.sc","").toLowerCase(Locale.ROOT)));
```
gnembon, line 196: while you are at it you can fix the potential issue if .sc is somewhere earlier in the name somehow (not sure if its possible), replaing to "\\.sc$"

altrisi:
Made it so you can only use the scripts in the `config` folder in singleplayer, changed the regex and mentioned the route in the docs. Currently editing the wiki.

gnembon (approved):

PR #591, opened 2020-11-29
Add a way to load a default config file

### LINKED ISSUE #211: Add a global config file for use by all singleplayer worlds.
It's a pain to turn off certain things I would rather not have available to me in a survival setting on a per-world basis. I understand the utility of a lot of these things but personally I'd rather have to force myself to do the open to lan enable cheats trick when I want to use some of these more powerful, cheaty tools.

### LINKED ISSUE #356: Default carpet.conf when creating new world
Having a carpet.conf file in `.minecraft/config` that gets copied to the world folder when a new world is created would prevent you from having to change the same settings every time you create a new world (or having to copy carpet.conf from another world).

altrisi:
Resolves #211 and resolves #356.

This PR makes Carpet read from a `default_carpet.conf` (or any identifier, for extensions) file if there is no `carpet.conf` in the current's world directory. Since those defaults will also be read when adding new rule defaults (`setDefault` or `removeDefault`), that means that the world's config file, when created, will be a copy of the default plus the new values. The default config files are stored at `[root]/config/carpet/default_carpet.conf` (with root usually being `.minecraft`, but basically install dir).

This functionality is explicitly disabled when Carpet is in a dedicated server, to prevent creating a useless `config/carpet` folder in servers, since those have a single world.

The default `conf` files are explicitly created whenever trying to load them, (mostly to show that the feature exists). Those have no effect by default since they are empty when created.

This PR also makes some other minor changes:
- To work, it requires `readSettingsFromConf()` to be able to read from different files, so now it requires the path to be passed and other functions are updated accordingly.
- I also changed some logic around files to use `Path` instead. That means that `getFile` will return a `Path` object instead of a `File`, and logic around writing and reading to/from the config has been slightly simplified by using functions like `newBufferedReader` or `...Writer` directly. The writer will now also use system's line separator instead of the default (allows for better reading in default editors in some systems).
- The closables around reading and writing are now specified in a "try-with-resources" block. That means they will be automatically closed after the block ends, no matter if it finishes or throws (previously they would not be closed after throwing).

### DIFF
--- a/src/main/java/carpet/settings/SettingsManager.java
+++ b/src/main/java/carpet/settings/SettingsManager.java
@@ -13,6 +13,9 @@ import com.mojang.brigadier.builder.LiteralArgumentBuilder;
 import com.mojang.brigadier.context.CommandContext;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 import com.mojang.brigadier.exceptions.SimpleCommandExceptionType;
+
+import net.fabricmc.api.EnvType;
+import net.fabricmc.loader.api.FabricLoader;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.ServerTask;
@@ -24,13 +27,13 @@ import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.util.TriConsumer;
 
 import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.FileReader;
-import java.io.FileWriter;
+import java.io.BufferedWriter;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.lang.reflect.Field;
+import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


gnembon (approved):

gnembon:
looks great, thanks :)

chexo3:
hooray!

supersaiyansubtlety:
I'm having trouble getting this to work. 
I copied an existing `carpet.conf` file from a save to `.minecraft/config/carpet/` and renamed the file to `default_carpet.conf`. 
I created a new world, and none of the carpet settings take effect (it's as though there's no config file).

altrisi:
Cannot reproduce, works for me.

Can you make sure that;
- The `config/carpet/` folder is located inside the actual "installation folder", in case you changed the default
- You are using one of the latest Carpet versions (>=1.4.20)
- You are not loading a world that already has a config file. In that case, Carpet will always prefer the world's file

Tested with a `default_carpet.conf` containing:
```
defaultLoggers tps
```

In theory, when loading a Carpet version matching those requirements a blank `default_carpet.conf` should have been generated automatically in the correct folder.

In case it still doesn't work, could you make an issue with more details about your system, to check it's not system dependant? If so make sure to include logs.

supersaiyansubtlety:
Oh, just had to update, my bad. 
I thought the feature had been around since older versions.

0RadicalLarry0:
> This functionality is explicitly disabled when Carpet is in a dedicated server, to prevent creating a useless config/carpet folder in servers, since those have a single world.

I would like the option to use it on my dedicated server as I create worlds often when playing with friends.

PR #610, opened 2020-12-05
Adding Gamerule modification to scarpet

### LINKED ISSUE #547: Scarpet: Missing support for wether?
Both querying and manipulating wether through scarpet would be useful. Support for interacting with the gamerule(s) could be useful too, besides jsut using `run('gamerule ...')`.

Ghoulboy78:
properly fixes #547 
TODOS:
- [x] Vote on whether or not it should be read-only
- [x] Add check for non-int or bool gamerules (from potential modded context, good call @altrisi )
- [x] Add mixin to get IntRule.value from GameRules.java (if vote says yes)


Ghoulboy78:
It's noteworthy to mention something gnembon mentioned to me, which is should gamerules be read-only through system_info or shuold they be writable?
On the one hand, carpet rules are read only, but on the other hand, they also control scripting engine, and can be more devastating

Please vote 👍 if you think gamerules should be read and writable (what it is now) and 👎  if you think gamerules should be read-only

BisUmTo:
Please, be consistant!

Ghoulboy78:
> Please, be consistent!

wdym?

Ghoulboy78:
I will make a decision after a week. I feel like that's enough time to allow evryone to vote. If after a week gnembon hasn't voted, his will be the final vote I will count.

altrisi:
What would happen if a rule isn't neither int nor boolean? (not possible in a vanilla context, but possible in a modded context)

Ghoulboy78:
Tru, I should add a check for that. In thtat case, it should fail and say oops.

Ghoulboy78:
cos there is no vanilla architecture for string gamerules, so its up to modmaker to implement, and I cant predict how its gonna be.

Ghoulboy78:
nice typo in the title there

altrisi:
Isn't `setValue()` client-only? (`@Environment(EnvType.CLIENT)`)

Ghoulboy78:
Look, that's the only way I could find to set the value.

Ghoulboy78:
ofc I dont lament that gnembon complains about my code, that's how I learned java in the first place...

altrisi:
But, the method being only present in the client means it doesn't exist in the server, so it won't work at all (it will even crash the server if unhandled).

Ghoulboy78:
I see, but there is literally no other method there cos mojank didnt add one. Ig I could add yet ANOTHER mixin magic bit, but im lazy so ill do it if vote passes. Rn were just waiting for gnembons vote, so idk.

altrisi:
If you want I can do that, I have more or less an idea of how to do it without mixing, and also supporting (most) non-vanilla types, both ways.

Ghoulboy78:
Sure, pr to this branch if u have an idea.

gnembon:
I get BisUmTo's idea about being constent wrt the gamerules and carpet rules, but its more about giving minecraft a proper API, not API to carpet - I actually didn't even think to add an API to even get carpet rules at the first place, but its fine that system_info supports that. 

I think that's good as its proposed - gamerule(...) to support both listing (no args - return a dict of everything), one arg (reading) and two args (changing).

Another argument to leave carpet rules as is (read-only) apart from the fact its not part of vanilla per se, is the fact that carpet rules may have various permissions set on each stuff that can be customized by the admin, so setting it via API might be little too backdoor'y, plus carpet can have other rulesets defined by other extensions using different command managers.  I think its ok to be this way, as Ghoulboy sugests, even if little inconsistent wrt carpet rules.

gnembon:
on the other hand, yeah - you can always call run('gamerule ...') to change it and honestly - have no problem with that. 
Leaning towards just returning the gamerules via system_info.

gnembon:
I think another reason to keep it read-only is to discourage changing them in scripts - its nothing more annoying than datapacks that change your gamerules and don't change them back.

Ghoulboy78:
k good i rly didnt feel like mixing again. @altrisi if u want do pr the better way to get the values, imma remove the set value thing.

gnembon:
yeah, maybe it will sound stupid, but I really liked my point on discouraging playing with gamerule values, by allowing to easy set them in the API - they could have been set by the administrator and it would be not a good idea encourage to play with them in the apps. Apps can still do that using run('gamerule ...') but that should give enough sensation they are doing something unintended to prevent that.

docs/scarpet/api/Auxiliary.md
```diff
@@ -486,7 +486,9 @@ Available options in the scarpet app space:
   * `world_path` - full path to the world saves folder
   * `world_folder` - name of the direct folder in the saves that holds world files
   * `world_carpet_rules` - returns all Carpet rules in a map form (`rule`->`value`). Note that the values are always returned as strings, so you can't do boolean comparisons directly. Includes rules from extensions with their namespace (`namespace:rule`->`value`). You can later listen to rule changes with the `on_carpet_rule_changes(rule, newValue)` event.
- 
+  * `world_gamerules` - returns all gamerules in a map form (`rule`->`value`). Unlike carpet rules, values are returned as numbers, so you can do boolean comparisons directly. Haven't tested with rules from other mods, we'll cross that bridge when it comes to it. Gamerules are read-only due to a ~~authoritarian decision~~ democratic vote.
```
gnembon, line 489:
```suggestion
  * `world_gamerules` - returns all gamerules in a map form (`rule`->`value`). Unlike carpet rules, values are returned as numbers, so you can do boolean comparisons directly. Haven't tested with rules from other mods, we'll cross that bridge when it comes to it. Gamerules are read-only to discourage app programmers to mess up with the settings intentionally applied by server admins. It is still possible to change them though using `run('gamerule ...`.
```

Ghoulboy78:
sounds fair

Ghoulboy78:
and gnembon, nothing sounds stupid when u say it. And if it does, then we are at fault, like a neural network outputting wrong result. We need to backpropagate and see internally what we did wrong. dw mate, keep it up.

gnembon:
Drop the mixins in favour of the built-in visitor.  Check MinecraftServer::dumpGamerules as an example.

gnembon:
Also - the example above returns as strings - see no problem with that tbh - why carpet rules need to be returned as strings, and these as numbers. Just copy that code.

Ghoulboy78:
I dont see that method

gnembon:
This way you can fit it all in a few lines of code - no need for a utility class for that

Ghoulboy78:
but i dont see the method

gnembon:
Capture

gnembon:
just 
```
Map<Value, Value> rules = new ...;
final GameRules gameRules = this.getGameRules();
GameRules.accept(new GameRules.Visitor() {
            public <T extends GameRules.Rule<T>> void visit(GameRules.Key<T> key, GameRules.Type<T> type) {
               rules.put(StringValue.of(key.getName()), StringValue.of(gameRules.get(key).toString())));
            }
         });
return MapValue.wrap(rules);
```

Ghoulboy78:
but why need to have `final GameRules gameRules = this.getGameRules();` be final?

gnembon:
well, just copied from mojang, no reason. Probably its just a decompiler

docs/scarpet/api/Auxiliary.md
```diff
@@ -486,7 +486,9 @@ Available options in the scarpet app space:
   * `world_path` - full path to the world saves folder
   * `world_folder` - name of the direct folder in the saves that holds world files
   * `world_carpet_rules` - returns all Carpet rules in a map form (`rule`->`value`). Note that the values are always returned as strings, so you can't do boolean comparisons directly. Includes rules from extensions with their namespace (`namespace:rule`->`value`). You can later listen to rule changes with the `on_carpet_rule_changes(rule, newValue)` event.
- 
+  * `world_gamerules` - returns all gamerules in a map form (`rule`->`value`). Unlike carpet rules, values are returned as numbers, so you can do boolean comparisons directly. Haven't tested with rules from other mods, we'll cross that bridge when it comes to it. Gamerules are read-only to discourage app programmers to mess up with the settings intentionally applied by server admins. It is still possible to change them though using `run('gamerule ...`.
```
gnembon, line 489:
```suggestion
  * `world_gamerules` - returns all gamerules in a map form (`rule`->`value`). Like carpet rules, values are returned as strings, so you can use appropriate value conversions using `bool()` or `number()` to convert them to other values. Gamerules are read-only to discourage app programmers to mess up with the settings intentionally applied by server admins. Isn't that just super annoying when a datapack messes up with your gamerule settings? It is still possible to change them though using `run('gamerule ...`.
```

gnembon (approved):

PR #617, opened 2020-12-07
Fix optimizedTNT entity behaviour

altrisi:
Fixes gnembon#616

This is basically because in 1.16 snapshots the explosion code changed to not be eye height only for TNT.

I don't know if this fully fixes that or if it produces any other bug. I've just tested that it fixed visually (and in the end pos in the tnt logger) the specific case in the issue. I'm not an expert in explosion code. Please test.

### DIFF
--- a/src/main/java/carpet/helpers/OptimizedExplosion.java
+++ b/src/main/java/carpet/helpers/OptimizedExplosion.java
@@ -132,7 +132,8 @@ public class OptimizedExplosion
 
                 if (d12 <= 1.0D) {
                     double d5 = entity.getX() - eAccess.getX();
-                    double d7 = entity.getY() + (double) entity.getStandingEyeHeight() - eAccess.getY();
+                    // Change in 1.16 snapshots to fix a bug with TNT jumping
+                    double d7 = (entity instanceof TntEntity ? entity.getY() : entity.getEyeY()) - eAccess.getY();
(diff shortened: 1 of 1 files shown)


gnembon:
Very likely that's the problem. Optimized TNT code has this problem that is a rewrite, so small details like that will pass unnoticed.

Please @Fallen-Breath  grab the proposed fix from https://github.com/gnembon/fabric-carpet/pull/617/checks and confirm it works as in vanilla.

Fallen-Breath:
just finished a quick testing in my 360ftl heavy cannon in 1.16.4 with normal config and maximum firepower config and it worked well. tho i tested with other mods but i think this issue is solved

altrisi:
Actually, I think the change was to fix TNT rendering in soulsand, and then this had to be added to prevent a bug that caused stacked TNT to jump.

BTW just for relating: Relates to jellysquid3/lithium#66. Else I wouldn't have known the reason for this.

Fallen-Breath:
> Actually, I think the change was to fix TNT rendering in soulsand, and then this had to be added to prevent a bug that caused stacked TNT to jump.
> 
> BTW just for relating: Relates to [jellysquid3/lithium-fabric#66](https://github.com/jellysquid3/lithium-fabric/issues/66). Else I wouldn't have known the reason for this.

ye that's why mojang did this change iirc

PR #618, opened 2020-12-08
Fixing/upgrading app commands

### LINKED ISSUE #594: Few issues with Scarpet draw and distance apps
List of issues (detailed below):
- [x] [Distance] Lines and labels are too near to the ground
- [ ] [Distance] Right clicks block functionality was changed again after merging
- [ ] [Draw] Update handling may be overcomplicated
- [x] ~[Distance] If lines are enabled, it will draw infinite lines in the same position~ Not Applicable anymore

<details>
<summary> [Distance] Lines and labels are too near the ground</summary>

When using the line functionality in the distance app (mostly with the brown carpet), the label and the line gets inside of the carpet, and it's directly on the floor. This could be solved by drawing the line slightly higher

</details>

### [Distance] Right clicks block was changed back to places block after merging

The `right_clicks_block` event was used instead of the `places_block` due to parity with the Java command. The Java command allows to set positions with the carpet even when the carpet doesn't actually get placed (like right clicking in walls), and that's the reason why it was changed (see https://github.com/gnembon/fabric-carpet/pull/553#discussion_r526396965). It was "reverted" in 4c6d449553fc6fa6d4373a899011fd4d22bf9681

### [Draw] Update handling may be overcomplicated

This was readded right before merging, in 623f9f39e16d77dee0ea1f3994508e2eef4788d3. As already commented in the pull request (and it was removed at that time), updates are already sent from Scarpet's `set` function when `fillUpdates` is true. However, it was added back last minute without a second to comment, overcomplicating both `affected()` and `set_block`, due to adding each pos into a list instead of just adding +1.

The change needed to revert this would basically be about removing the `without_updates` wrap in `set_block`, only adding +1 to global_affected (maybe even remove the `if block != block` since that is already handled by Carpet and do `global_affected += set(...)` since booleans are numbers, or similar), changing `global_affected` to be and be reset to a number again and printing just the variable.

<details>
<summary>[Distance] If lines are enabled, it will draw infinite lines in the same position</summary>

Tired of writing, but this is an important one. See the thing and a sketch for a workaround in https://github.com/gnembon/fabric-carpet/pull/553#discussion_r527476529 (slightly outdated). "Part of the fix" was kind-of added, but the most important part wasn't (not scheduling a new line)

</details>

#### Other small things that may or may not be considerable
- ~[Distance] The label has almost no information (probably still WIP)~
- ~[Distance] The label recalculates the distance every time it's rendered~
- ~[Distance] It's not necessary to check for start when drawing since it will always be present when end is~

### LINKED ISSUE #637: [Suggestion] Abstract core functionality of draw app into a library
With the rise of world editing software since the recent changes in other software, there are suggestions of developing and integrating new methods of doing the same as the `draw` (beta) app inside other, bigger apps.

Therefore, my suggestion here would be to make the draw command app just a wrapper for a more generic library that would contain the shape processing code, that could then also be included in other apps, instead of having duplicated code.

My suggestion would be to separate every shape building method from the app into the library, and then following the next format to be called:

```py
draw_[shape]([...params], set_handler);

//# Example set_handler implementation
set_handler(pos)-> (
    existing = block(pos);
    if (global_block != existing && (!global_replacement || _block_matches(existing, global_replacement) ),
        global_positions += bool(set(existing, global_block))
    )
);

//# Example call
import('shapes','draw_sphere');

global_replacement = 'dirt';
global_block = 'iron_block';
global_positions = 0;

draw_sphere([100,60,100], 25, true, 'set_handler'); //# draw_sphere(pos, radius, hollow, set_handler)
print('Affected '+global_positions+' blocks');
```

Relates to Ghoulboy78/World-edit-scarpet#4

Ghoulboy78:
Fixes #594 
Fixes #637 
Changes: Adds proper caching to /draw(_beta), with toggleable options for how to cache
Distance now returns a message saying which display mode you were on. Also setting the mode to something, then back to chat didn't work
Also updated commands for ai_tracker and cammera.sc
And added docs for built in apps to scarpet docs, cos they may be useful. Tbh, I was not so sure about adding overlay, ai_Traker, draw and distance cos they have little functionality in developing scarpet apps, but im not sure where else to put them. maybe the wiki?


gnembon:
Don't think docs about built-in apps should be in language reference. Its just carpet mod using scarpet as an implmentation medium - has little to do with scarpet docs at all.
I would keep those - as part of wiki and linked up from the main README.

Ghoulboy78:
ye thought soo too

Ghoulboy78:
https://github.com/gnembon/fabric-carpet/wiki/Built-in-Scarpet-apps

altrisi:
I'm not sure if the docs for the distance and draw apps should be with built-in apps, but with regular Carpet commands, once they get released. Currently in beta not many people will look at them, since they aren't even suggested.

Also, how much processing do you save with the cache? Didn't you say that the overhead comes from placing the blocks into the world, not from the loop? Then it shouldn't make much change, would it? It may even be worse, depending on the speed of saving/loading from cache. Anyway, it at least makes the code a bit messier, and btw why don't you just use `save_app_data`?

Also not sure about keeping debug statements in code that should be final and ready-ish for release.

Gotta test this to get proper feedback first (sorry if something I said is completely wrong!). Will do it Soon™, whenever I get a bit more time (time is relative).

Ghoulboy78:
I don't use save_app_data cos its deprecated.
I agree about the wiki thing, but thats after the apps come out of beta, so its fine for now tbh.

About the cahce thing, I suprised even myself, but it decreases ms used 8x with spheres, which are the most intensive to generate. The first time, it took 56 ms, the second time it took like 7. I know, it's not exactly cacnonical testing, but I rly havent had time, I just slapped it together and thourght it might be more useful.


And anyways the debug is only for people in developer environments and for people with gbhs sgnf sadsgras fhskdpri!!!

Ghoulboy78:
Ill remove it ofc.

altrisi:
Don't confuse it!
`save_app_data()` by itself (default loc) isn't deprecated!
Only `save_app_data(nbt, otherfilename)` (and load ofc).

Ghoulboy78:
ok ye, but apparently its `store_app_data`  not `save_app_data`. Plus what's the point? the ccode for store_app_data and write_file('nbt') is literally copypasta, so it changes nothing

Ghoulboy78:
and also Im only saving/loading from cache at load, unload, and when the player requests it.

src/main/resources/assets/carpet/scripts/ai_tracker.sc
```diff
@@ -382,6 +322,55 @@ global_functions = {
...
+};
+
+
+__config() ->{
+    'commands'->{
+        ''->'_command',
+        'clear'->'clear',
+        'toggle boxes'->_()->global_display_boxes = !global_display_boxes,
+        'toggle <display>'->['__toggle',null],
+        'toggle villager_hostile_detection <hostile>'->_(h)->__toggle('villager_hostile_detection',h),
```
gnembon, line 346: I don't see the point of rewriting the commands to keep the flat call structure for the sake of it. In this case clearly villager_ was just a prefix to fix issues with lack of commands customization.
villager should be its own path with options like hostile_detection, golem_spawning or buddy_detection

src/main/resources/assets/carpet/scripts/camera.sc
```diff
@@ -57,6 +57,47 @@ __command() ->
...
+        'interpolation gauss'->['__interpolation','gauss',true],
+        'interpolation gauss <float>'->_(float)->(__interpolation('gauss_'+str(float),true)),
+        'repeat <seconds> <last_delay>'->'repeat',
+        'stretch <factor>'->'stretch'
+    },
+    'arguments'->{
+        'seconds'->{'type'->'int','suggest'->[]},
+        'last_delay'->{'type'->'int','suggest'->[]},
+        'name'->{'type'->'string','suggest'->[]},
+        'interpolation'->{'type'->'term','options'->['linear','cr']}
```
gnembon, line 79: suggest using fuller names as now they are suggested, cr -> catmull_rom

src/main/resources/assets/carpet/scripts/draw_beta.sc
```diff
@@ -15,7 +15,14 @@ __config() -> {
         'cuboid <center> <radius> <height> <orientation> <block> <hollow>'->['draw_prism', null, true],
         'cuboid <center> <radius> <height> <orientation> <block> <hollow> replace <replacement>'->['draw_prism', true],
         'cylinder <center> <radius> <height> <orientation> <block> <hollow>'->['draw_prism', null, false],
-        'cylinder <center> <radius> <height> <orientation> <block> <hollow> replace <replacement>'->['draw_prism', false]
+        'cylinder <center> <radius> <height> <orientation> <block> <hollow> replace <replacement>'->['draw_prism', false],
+        'cache <mode>'->'set_cache_mode',
```
gnembon, line 19: not sure why you need a cache. setting the block should be majority of the resources it should require.

src/main/resources/assets/carpet/scripts/draw_beta.sc
```diff
@@ -42,26 +52,21 @@ _block_matches(existing, block_predicate) ->
...
-            global_affected += [x, y, z];
-        ));
+    if(block != existing && (!replacement || _block_matches(existing, replacement) ),
+        global_positions:1 += bool(set(existing,block));
     );
+    global_positions:0+=[x, y ,z]-cpos//so it's relative to centre pos, and we can draw from cache more smartly
 );
 
-global_affected = [];
+global_positions = [[],0];
```
gnembon, line 63: why do you worry about checking list length?

gnembon:
if that cache thingy really improves calculations, this means the algorithms are bad.
Also altrisi is right - if the only thing you want to have synced is one compound tag - use app_data since these are non-blocking on the IO and prevent constant read-writes that might not be necessary.

src/main/resources/assets/carpet/scripts/draw_beta.sc
Ghoulboy78: cos the number indicates no. of successes, list indicates all blocks in that shape.

src/main/resources/assets/carpet/scripts/draw_beta.sc
Ghoulboy78: Nah man, it rly helps. (at least it does on my craptop)

Ghoulboy78:
but im not constantly reading and writing. I mean, sure I will change it, but there is no constant read-writing. Only (as I said), at start, end, and whenever the user wants.

Ghoulboy78:
and anyways, thats just for the sphere. Ill see how things are for other shapes, and see whether cache rly helps.

src/main/resources/assets/carpet/scripts/ai_tracker.sc
Ghoulboy78: ye i thought abt that tbh...

src/main/resources/assets/carpet/scripts/camera.sc
Ghoulboy78: done

Ghoulboy78:
k yeah I was testing the cache speed for other shapes and noticed that they're messed up now. I will turn this into a draft pr, and continue working on it when I have time (later)

src/main/resources/assets/carpet/scripts/draw_beta.sc
gnembon: that's because the sphere drawing code is wrong.

Ghoulboy78: k then, what code should I use?

gnembon:
ohboi, just throw that caching to the garbage and fix the algoritms. While filling up the solid cube can be defended using scanning of the area, drawing hollow figures just cannot.
Run the following example and you will feel the difference:
```
__config() -> 
{
   'commands' ->
   {
      '<center> <radius> <block> one ' -> 'draw_sphere_baldy',
      '<center> <radius> <block> two' -> 'draw_sphere_better'
   },
   'arguments'->{
        'center'->{'type'->'pos', 'loaded'->'true'},
        'radius'->{'type'->'int', 'suggest'->[], 'min'->0},//to avoid default suggestions
   }
};

length_sq(vec) -> reduce(vec, _a + _*_, 0);

draw_sphere_baldy(centre, radius, block)->(
   start = time();
    scan(centre,[radius,radius,radius],
        l = length_sq([_x,_y,_z]-centre);
        if((l<=radius^2+radius) && l>=radius^2-radius,
            null//set(_, block)
        )
    );
    print('took '+(time()-start));
);

draw_sphere_better(centre, radius, block)->
(
   start = time();
   for ( range(-90, 90, 45/radius),
      cpitch = cos(_);
      spitch = sin(_);
      for ( range(0, 360, 45/radius),
         cyaw = cos(_);
         syaw = sin(_);
         bpos = centre+[cyaw*cpitch, spitch, syaw*cpitch]*radius;
         null//set(bpos, block);
      )
   );
   print('took '+(time()-start));
);
```

gnembon:
its still not how its supposed to be, but that's the type of reasoning that needs to be applied.

Ghoulboy78:
but I like baldy spheres...

Ghoulboy78:
Ill fix ofc.

Ghoulboy78:
k, i rewrote sphere and ball thing.

Ghoulboy78:
Current algorithms and O():
sphere:O(n^2) (previous n^3)
ball:O(n^3)
diamond:O(n^2)
hollow pyramid:O(n^2) (previous n^3)
filled pyramid:O(n^3)
hollow cone:O(n^2) (previous n^3)
filled cone:O(n^3)
hollow cylinder:O(n^2) (previous n^3)
filled cylinder:O(n^3)
hollow cuboid:O(n^2) (previous n^3)
filled cuboid:O(n^3)

Ghoulboy78:
the spheres and balls are done, jsut need to finish up the cones and cylinders.

Ghoulboy78:
Now that Ive commited, the above stats are true. But the cones are still messed up... Ill try to figure wtf is going on.

gnembon:
looks fine now - not sure if its correct, but at least looks fine by the book.

You have messed up merging from upstream - you need to fix it.

Ghoulboy78:
ye the circle algorithm doesnt work for small circles, so cones are screwed

Ghoulboy78:
@gnembon this is ready now.

Ghoulboy78:
btw also partially fixes Ghoulboy78/World-edit-scarpet#40

Ghoulboy78:
tho I cant do that from here it seems.

altrisi:
Note: This does NOT resolve #637, since you cannot `import()` things from (at least unloaded) ~non-library~ _rule_ apps _(since when can you import from non-libraries btw?)_. The idea in that issue was to move those basic functions into another (library) app, with them returning a list to be used in the draw app, and any other app that would need them.

I would suggest to keep that separate from this PR, since this already has many things needing review in it blocking it from being merged.

BTW, you forgot to remove the debug timers (the times can already be gotten by executing the functions with `/script in [app] run ...`)

Ghoulboy78:
K, im closing this now

Issue #548, opened 2020-11-15
Suggestion: Allow Scarpet to detect exactly which trade a player used (besides just the output item)


James103:
Currently (Carpet mod 1.4.18), there is no event that detects exactly which trade a player used when the player trades with a villager. `__on_statistic(...)` can be used to detect when a player trades with a villager, but that only reveals the output item as `event == crafted.<item>`.

I would like if there was an event that was triggered every time a player trades with a villager. The event would be something like `__on_player_trade(player, entity, buy, buyB, sell)`, where
* `player` is the player that traded with the villager / triggered the event,
* `entity` is the villager or wandering trader that the player traded with,
* `buy` is the `item_tuple` for the item that the villager bought for that trade,
* `buyB` is the `item_tuple` for the other item that the villager bought for that trade, and
* `sell` is the `item_tuple` for the item that the villager sold for that trade.

If the former is not possible, then an alternative would be `__on_player_trade(player, entity, trade)`, where
* `player` is the player that traded with the villager / triggered the event,
* `entity` is the villager or wandering trader that the player traded with, and
* `trade` is the index into the villager or wandering trader's list of trades.

Ghoulboy78:
umm, currently we are not in 1.4.12. Were in like 1.4.19. But cool idea ngl

James103:
Closing as fixed (feature implemented).

PR #623, opened 2020-12-11
Added onTrade Scarpet Event

BisUmTo:
Implemented James103's suggestion #548
Added `__on_player_trades(player, entity, buy, buyB, sell)` event handler.
The parameter `entity` can be `null` if the merchant is not an entity _[in vanilla should not occur]_

### DIFF
--- a/src/main/java/carpet/mixins/TradeOutputSlot_scarpetEventMixin.java
+++ b/src/main/java/carpet/mixins/TradeOutputSlot_scarpetEventMixin.java
@@ -20,7 +20,7 @@ import static carpet.script.CarpetEventServer.Event.PLAYER_TRADES;
 public abstract class TradeOutputSlot_scarpetEventMixin {
     @Shadow @Final private Merchant merchant;
 
-    @Inject(method = "onTakeItem", require = 0, at = @At(value = "INVOKE",
+    @Inject(method = "onTakeItem", at = @At(value = "INVOKE",
             target = "Lnet/minecraft/village/Merchant;trade(Lnet/minecraft/village/TradeOffer;)V"),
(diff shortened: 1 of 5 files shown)


gnembon:
Would be better to have more generic event when player interacts with screens, one of them would be the merchant, but I haven't yet thouht how to do that properly.

Its not a big deal though since we could deprecate that then. Its probably gonna take quite some time until sceen API is added to the scarpet.

src/main/java/carpet/mixins/TradeOutputSlot_scarpetEventMixin.java
```diff
@@ -0,0 +1,33 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+import static carpet.script.CarpetEventServer.Event.PLAYER_TRADES;
+
+@Mixin(value = TradeOutputSlot.class)
+public abstract class TradeOutputSlot_scarpetEventMixin {
+    @Shadow @Final private Merchant merchant;
+
+    @Inject(method = "onTakeItem", require = 0, at = @At(value = "INVOKE",
```
gnembon, line 23: why did you make is require=0? I would skip that and assume it successful. This way if mojang changes something we will immediately know, or have a chance to know.

gnembon, line 23:
```suggestion
    @Inject(method = "onTakeItem", at = @At(value = "INVOKE",
```

gnembon (approved):

PR #628, opened 2020-12-14
Add FormattedText addition

### LINKED ISSUE #622: Format of text is stripped when concatenating two formatted text strings from within Scarpet
As of Carpet mod 1.4.20 and MC 1.16.4, if you do something like the following, the result is plain white text:
```
display_title(player(), 'actionbar', format('gi This should be green italics ') + format('r This should be red'));
```

altrisi:
Resolves #622.

This makes the addition `+` operator correctly join two formatted text, or a formatted text first and anything else after.

~This however, doesn't allow anything `+` formatted text (reverse order), since that would make the function quite complicated (probably), and required in base `Value`. May do that at some point.~

Ok, after some time (around 20 minutes), I've decided to also add the functionality to the other side, that is: anything `+` formatted text.

```py
format('y Hello ') + format('d World') = format('y Hello ', 'd World')

format('y Hello ') + 'World' = format('y Hello ', 'w World') //# Assuming "w" as default color

format('y Hello ') + 5 = format('y Hello ', 'w 5') //# Assuming "w" as default color

'Hello '+format('d World') = format('w Hello ', 'd World' //# Assuming "w" as default color
```

~Also calls `shallowCopy()` on the `Text` element when cloning, since it's technically mutable.~

### DIFF
--- a/src/main/java/carpet/script/value/FormattedTextValue.java
+++ b/src/main/java/carpet/script/value/FormattedTextValue.java
@@ -48,10 +48,25 @@ public class FormattedTextValue extends StringValue
         return StringTag.of(Text.Serializer.toJson(text));
     }
 
+    @Override
+    public Value add(Value o) {
+        if (o instanceof FormattedTextValue)
+        {
+            Text mergedText = ((BaseText) text).shallowCopy().append(((FormattedTextValue) o).getText());
+            return new FormattedTextValue(mergedText);
+        }
+        else
+        {
+            Text mergedText = ((BaseText) text).shallowCopy().append(o.getString());
+            return new FormattedTextValue(mergedText);
+        }
+    }
+
     public String serialize()
     {
         return Text.Serializer.toJson(text);
     }
+
     public static FormattedTextValue deserialize(String serialized)
     {
         return new FormattedTextValue(Text.Serializer.fromJson(serialized));
--- a/src/main/java/carpet/script/value/FormattedTextValue.java
+++ b/src/main/java/carpet/script/value/FormattedTextValue.java
@@ -3,6 +3,7 @@ package carpet.script.value;
 import net.minecraft.nbt.StringTag;
 import net.minecraft.nbt.Tag;
 import net.minecraft.text.BaseText;
+import net.minecraft.text.LiteralText;
 import net.minecraft.text.Text;
 
 public class FormattedTextValue extends StringValue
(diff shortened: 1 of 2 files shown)


src/main/java/carpet/script/value/Value.java
```diff
@@ -79,6 +81,14 @@ public String getPrettyString()
 
     public Value add(Value o) {
         String lstr = this.getString();
+        if (o instanceof FormattedTextValue)
```
gnembon, line 84: That's the only part I am little squinting about, but I guess its fine. '+' on non-numerical values is rare and doesn't need to be super efficient, so adding special cases is I guess fine.

src/main/java/carpet/script/value/FormattedTextValue.java
```diff
@@ -27,7 +27,7 @@ public boolean getBoolean() {
     @Override
     public Value clone()
     {
-        return new FormattedTextValue(text);
+        return new FormattedTextValue(text.shallowCopy());
```
gnembon, line 30: I used clone only in the cases of assigning variables, and since text values are immutable here (and you kept that), this is not needed.

src/main/java/carpet/script/value/Value.java
```diff
@@ -79,6 +81,14 @@ public String getPrettyString()
...
+        {
+            if (lstr == null)
+                return ((FormattedTextValue) o).clone();
```
gnembon, line 87: since this piece of code is same/similar than the .add() in FormattedText, why don't you have static combine(Formatted 1, Formatted 2) in FormattedTextValue and use that here and in .add()

gnembon:
Yeah, might be useful.

src/main/java/carpet/script/value/FormattedTextValue.java
```diff
@@ -14,6 +15,33 @@ public FormattedTextValue(Text text)
...
+    public static Value combine(Value left, Value right) {
+        BaseText text;
+        if (left instanceof FormattedTextValue)
+        {
+            text = (BaseText) ((FormattedTextValue) left).getText().shallowCopy();
+        }
+        else
+        {
+            if (left instanceof NullValue)
+                return right;
```
altrisi, line 27: Should I use `getString() == null` instead?

As far as I understood, `getString()` can't ever be null, since `NullValue` would return `"null"` as the String, that's why I changed it.

Also should I clone the variables before returning them in those cases?

gnembon: left.isNull() should be preferred.

gnembon: but it doesn't matter that much. I wanted to stop using instanceof, since isNull may get more complex in the future. Don't worry about it.

gnembon (approved):

PR #631, opened 2020-12-16
Added ability to change criterion with scoreboard_add Fixes #624

### LINKED ISSUE #624: [scarpet][suggestion]add a function to modify the criterion of the scoreboards.

replaceitem:
Fixes #624

I left the function name `scoreboard_add` for now, because i wasn't sure if it was already decided to deprecate it and rename it to `scoreboard_set` https://github.com/gnembon/fabric-carpet/issues/624#issuecomment-742965687_

Also i really hope i did the mixin stuff correctly, im really not experienced in that stuff.


altrisi:
_Tip: Add "Fixes" (or one of [this](https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)) before the issue number in the PR description or commit message for the issue to be automatically closed on merge, and to add some nice relations panels both in the issue and in the PR._

replaceitem:
Will add a check if the criterion is already what it will be changed to, and in that case return null (meaning nothing changed)

docs/scarpet/api/Scoreboard.md
```diff
@@ -10,8 +10,9 @@ With specified `objective` and
  
 ### `scoreboard_add(objective, criterion?)`
 
-Adds a new objective to scoreboard. If `criterion` is not specified, assumes `'dummy'`. Returns `false` if objective 
-already existed, `true` otherwise.
+Adds a new objective to scoreboard. If `criterion` is not specified, assumes `'dummy'`.
+If the objective already exists, changes the criterion of that objective and returns false.
+If the objective was added, returns true.
```
gnembon, line 15:
```suggestion
If the objective was added, returns `true`. If nothing is affected, returns `null`
```

src/main/java/carpet/script/api/Scoreboards.java
```diff
@@ -133,8 +136,13 @@ public static void apply(Expression expression)
             }
 
             ScoreboardObjective objective = scoreboard.getObjective(objectiveName);
-            if (objective != null)
+            if (objective != null) {
+                ((Scoreboard_scarpetMixin)scoreboard).getObjectivesByCriterion().get(objective.getCriterion()).remove(objective);
```
gnembon, line 140: Yeah, I would check if the criterion is same and skip doing anything, returning `null` for a change.

replaceitem: Yes that makes a lot of sense! I was also wondering, when creating an objective without specifying a criterion, it defaults to dummy. But should it also default to dummy when changing a criterion?
I can imagine the scenario of someone just running `scoreboard_add(objective)` with an already existing objective by accident, and not realizing that the criterion got reset to dummy. Maybe that should only change when a criterion is actually specified.

In that case, running the above could return the criterion as a string maybe?
It wouldnt fir the name of the function, as well as scoreboard_set if it got changed though

replaceitem:
Alright so now i made it so it returns the current criterion if the objective already exists, but no criterion is specified, and returns null if nothing changed (when a criterion *is* specified

gnembon:
Looks great - thanks for your help.

gnembon (approved):

PR #634, opened 2020-12-20
❄️ Freezing Scarpet ❄️

### LINKED ISSUE #398: `tick freeze` command still runs `#minecraft:tick` functions
During a `tick freeze`, the server still runs functions tagged `#minecraft:tick` very often (I donʼt know if itʼs still 20Hz like normal); this makes debugging of tick-perfect data packs much more difficult.

To reproduce:
1. Create a datapack with a function tagged `#minecraft:tick` with some observable effect (see attached, which says “Tick!” each tick as long as thereʼs an entity named “Announce ticks”)
2. Create a world with this data pack, and run `/tick freeze`
3. Observe effect still happens.

Affected version: carpet 1.4.0 on Minecraft 1.16.1.  I didnʼt see anything about this in the changelog for post-1.4.0, but I havenʼt tested on 1.4.5 because I donʼt have a prerelease instance with fabric.

[announce ticks.zip](https://github.com/gnembon/fabric-carpet/files/5009981/announce.ticks.zip)

### LINKED ISSUE #581: Scheduled functions in Scarpet tick during /tick freeze deep
May be intended. Happens both in regular and deep freezes (maybe stopping them in deep ones but not in regulars, if intended?).

Relates to #398

From https://github.com/gnembon/fabric-carpet/issues/581#issuecomment-735455786:

My view of this:

PROS:
- Functions or apps that rely on actual ticks would always get accurate information
- It seems logical ("makes sense") that processing stops when the `Game is frozen`
- If shapes expiry time is also paused, would not affect debuggers with shapes, but may even make them better in case they vary based on execution number
- Apps that get data from the environment would not incorrectly update (e.g. app getting info from ticking command blocks)

CONS:
- Functions wouldn't run when the `Game is frozen` (not really a con in most cases, it's basically the premise of the issue)
- If shapes expiry time is not paused, would make debuggers useless when frozen
- Carpet-less clients wouldn't be sent new particles for shapes in debuggers, including regular particles

Tickrate is synced with clients, so shapes expiry time could be easily paused. Therefore the only con I see would be the last one.

Maybe making it be deep-only to allow compatibility if someone really needs the feature, but that would also make it harder to find, and I personally can't find any use case. Please comment one if you have it.

EDIT: Ok I found one: Actually controlling a `freeze`, something like temporary freezing the game for x ticks.

<details>
<summary>Messed up poll</summary>

## What do you think?

(clickable)

[](https://api.gh-polls.com/poll/01ERWVFZ2RDDB4H7WJ37J1CWDN/Always%20freeze%20Scarpet%20schedules/vote)
[](https://api.gh-polls.com/poll/01ERWVFZ2RDDB4H7WJ37J1CWDN/Freeze%20Scarpet%20schedules%20on%20deep%20freeze/vote)
[](https://api.gh-polls.com/poll/01ERWVFZ2RDDB4H7WJ37J1CWDN/Never%20freeze%20Scarpet%20schedules/vote)

</details>

In progress PR:
- [x] Stop tick functions from datapacks
- [x] Stop tick events and freeze schedules in Scarpet
- [x] Sync frozen state to the client
  - [x] Make the required packet info
  - [x] Handle that new packet info
  - [x] Find a way to tick the client's `TickSpeed` in servers (to make `tick step` and maybe `tick superHot` render properly)
  - [x] Add 1 to duration of shapes every client tick if frozen

The PR should actually also make freezes look good/better in servers.

altrisi:
_And the client_

Actually, the most important (visible) thing added from this PR is freezing the client, but Scarpet is also frozen.

This is the PR that resolves #581 and resolves #398.

### What does it do?

_It adds new data to the Carpet Protocol to sync more features, such as frozen state and deep frozen state with Carpet clients._

Now, to the full PR:

Syncs the `is_paused`, `deepFreeze` and `super_hot` variables from `TickSpeed` with clients. With the next changes, this will allow us to freeze entity rendering, like in singleplayer, and to add client-side frozen features. The `super_hot` variable doesn't really change absolutely anything, but I added it for completeness sake to maybe try to get it working in the future (because AFAIK it does change things in SSP, doesn't it? if not, I'm dumb and that should be reverted).

Since freezes can be stepped forward, whenever the `step` command is ran, that time is also synced to clients to unfreeze the rendering during those ticks. It doesn't look as good as in singleplayer (client tends to process a bit more than server), but it's better than not having it.

As you probably know, those booleans alone don't really change how the world is rendered directly, since that is handled by the `process_entities` variable, that is determined every tick in `TickSpeed`'s `tick()` method. Since it didn't get naturally ticked in the client, there is new mixin functionality in `MinecraftClientMixin` to tick `TickSpeed` without an attached server. That allows us to get the (almost always) correct value for each client tick depending on the frozen state. This will only work 100% (-ish) accurate when the server's tickrate is 20, or when it has `smoothClientAnimations` and tickrate <= 20 (`smoothClientAnimations` affects our client tick source, affecting also shape duration, etc).

One thing is that in order to tick `TickSpeed` from the client, I had to remove the `MinecraftServer` parameter from the `tick` method, which wasn't used anyway. If needed, it can be readded and just pass `null` from client ticks, but that would need checking if that is null before using.

Now, to the Scarpet freezing part.

First, this prevents `EventServer` (schedules) and `tickTasks` (`on_tick` events) from being ticked when the server is under a deep freeze. I personally think tick events shouldn't tick in regular ones either, but reasoning for this below. There is also an added check in the datapack profiler mixin that just cancels the event if there is a deep freeze, since the mixin seems to be in the perfect location (not tested, I've never used any datapacks, but should work).

Since we are freezing schedules and ticks, that would kill shapes, since there would be no way to send new ones, and we don't want that. Therefore we make use of the newly added synced frozen state to extend shapes lifetime while the world is deeply frozen. That prevents shapes to disappear when they can't be renewed by anything else. The way this is done may not be perfect, but I couldn't find an easy way with the current implementation.

This PR works and is technically feature complete, but there are some questions/considerations that I'd like to ask/suggest:

- [x] As stated above, I personally think `__on_tick()` (and derivatives, including datapack's tick) should neither tick in regular freezes. I think that since, even though yes, the server still ticks, to the user it is supposed to be "frozen", therefore for a user it may seem counter-intuitive for them to tick with the game "frozen". This doesn't apply to `schedule()` due to the reasons mentioned in the rfc. There wasn't really any discussion about the tick events, so I just added all of them only in deep freezes. What do you think?
- [x] `TickSpeed`'s frozen state may need setters/getters. Since now those are synced to clients, I think it makes sense to have setters for them that would automatically notify the client, to prevent anyone (e.g. in an extension) from modifying those variables without syncing them. Maybe both regular and deep frozen state in the same function to prevent unnecessary packets/calls. Thoughts?
- Please tell me a better name for `updateTickPlayerActiveTimeoutToConnectedPlayers` (and derivatives). It's extremely long, but is the only way I've found for it to explain what does it mean.
- [x] ~In case `super_hot` doesn't have any kind of interaction with the client (like visible frozenness, etc), please tell me because it should not be synced, and probably `tickingState` could be just `frozenState` or `pauseState`~ SuperHot has been separated from frozen state
- `renewShapes` may need a better name, and also probably a better implementation

I think this is ready for review. Open to change requests, as always. Will do some more testing since I accidentally rebased the branch down while working, but I think I readded all the necessary changes. Sorry for not being more concise in the description.

### DIFF
--- a/src/main/java/carpet/script/CarpetEventServer.java
+++ b/src/main/java/carpet/script/CarpetEventServer.java
@@ -961,7 +961,7 @@ public class CarpetEventServer
 
     public void tick()
     {
-        if (!TickSpeed.process_entities && TickSpeed.deeplyFrozen())
+        if (!TickSpeed.process_entities)
             return;
         Iterator<ScheduledCall> eventIterator = scheduledCalls.iterator();
         List<ScheduledCall> currentCalls = new ArrayList<>();
--- a/src/main/java/carpet/script/CarpetEventServer.java
+++ b/src/main/java/carpet/script/CarpetEventServer.java
@@ -961,7 +961,7 @@ public class CarpetEventServer
 
     public void tick()
     {
-        if (!TickSpeed.process_entities && TickSpeed.deepFreeze)
+        if (!TickSpeed.process_entities && TickSpeed.deeplyFrozen())
             return;
         Iterator<ScheduledCall> eventIterator = scheduledCalls.iterator();
         List<ScheduledCall> currentCalls = new ArrayList<>();
--- a/src/main/java/carpet/script/CarpetEventServer.java
+++ b/src/main/java/carpet/script/CarpetEventServer.java
@@ -2,6 +2,7 @@ package carpet.script;
 
 import carpet.CarpetServer;
 import carpet.CarpetSettings;
+import carpet.helpers.TickSpeed;
 import carpet.script.exception.InternalExpressionException;
 import carpet.script.value.BlockValue;
 import carpet.script.value.EntityValue;
@@ -960,6 +961,8 @@ public class CarpetEventServer
 
     public void tick()
     {
+        if (!TickSpeed.process_entities && TickSpeed.deepFreeze)
+            return;
         Iterator<ScheduledCall> eventIterator = scheduledCalls.iterator();
         List<ScheduledCall> currentCalls = new ArrayList<>();
         while(eventIterator.hasNext())
--- a/src/main/java/carpet/CarpetServer.java
+++ b/src/main/java/carpet/CarpetServer.java
@@ -109,7 +109,7 @@ public class CarpetServer implements ClientModInitializer,DedicatedServerModInit
 
     public static void tick(MinecraftServer server)
     {
-        TickSpeed.tick(server);
+        TickSpeed.tick();
         HUDController.update_hud(server);
         scriptServer.tick();
 
--- a/src/main/java/carpet/commands/TickCommand.java
+++ b/src/main/java/carpet/commands/TickCommand.java
@@ -2,6 +2,7 @@ package carpet.commands;
 
 import carpet.CarpetSettings;
 import carpet.helpers.TickSpeed;
+import carpet.network.ServerNetworkHandler;
 import carpet.settings.SettingsManager;
 import carpet.utils.CarpetProfiler;
 import carpet.utils.Messenger;
@@ -107,6 +108,7 @@ public class TickCommand
             TickSpeed.deepFreeze = false;
             Messenger.m(source, "gi Game runs normally");
         }
+        ServerNetworkHandler.updateTickingStateToConnectedPlayers();
         return 1;
     }
 
@@ -119,6 +121,7 @@ public class TickCommand
     private static int toggleSuperHot(ServerCommandSource source)
     {
         TickSpeed.is_superHot = !TickSpeed.is_superHot;
+        ServerNetworkHandler.updateTickingStateToConnectedPlayers();
         if (TickSpeed.is_superHot)
         {
             Messenger.m(source,"gi Superhot enabled");
--- a/src/main/java/carpet/commands/TickCommand.java
+++ b/src/main/java/carpet/commands/TickCommand.java
@@ -96,19 +96,15 @@ public class TickCommand
 
     private static int toggleFreeze(ServerCommandSource source, boolean isDeep)
     {
-        TickSpeed.is_paused = !TickSpeed.is_paused;
-        if (TickSpeed.is_paused)
+        TickSpeed.setFrozenState(!TickSpeed.isPaused(), isDeep);
+        if (TickSpeed.isPaused())
         {
-            TickSpeed.deepFreeze = isDeep;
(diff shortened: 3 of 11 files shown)


altrisi:
TODO to self: Update docs

gnembon:
wdym update docs? docs are only for scarpet, right?

Oh - update scarpet docs to explain what to expect if world is frozen with a mod. Get ya.

gnembon:
Looks great. I agree that these tick indicators that you are syncing with the clients should be there. 

Regarding the questions you have raised.
1) agree - regular tick events should also be paused in `regular` freeze - feel free to change it.
2) handling with setters and getters makes sense to handle changes to client nicer - feel free to change it.
3) name's fine
4) its fine to sync super hot - its not a big deal to sync it, and it might be useful in the future.
5) its all fine - it runs through all the shapes, but it would run through them anyways. Its better than ticking them down when not-frozen.

gnembon (approved):

gnembon:
approved for now - but as I said - agree with point1 and point2, so feel free to adjust that.

altrisi:
Should I make the `deepFreeze` getter check for `process_entites`? It's only used with both.

src/main/java/carpet/script/CarpetEventServer.java
```diff
@@ -961,7 +961,7 @@ public CarpetEventServer(MinecraftServer server)
 
     public void tick()
     {
-        if (!TickSpeed.process_entities && TickSpeed.deepFreeze)
+        if (!TickSpeed.process_entities && TickSpeed.deeplyFrozen())
```
gnembon, line 964: well this is for 'schedule' events that rely on ticks - since tick events are paused in any frozen state, these should also be halted in regular frozen state.

altrisi: Wasn't it decided in the RFC at #581 that `schedule()` would use deep freeze instead of regular? (for things like "controlling" a freeze, for example?)

gnembon: well, 'schedule' is meant to run alongside the game - if entities are ticking, or blocks are interacting, schedule is something that follows that routine.

altrisi: Changed

PR #635, opened 2020-12-21
Use latest ServerPlayerEntity after respawn

### LINKED ISSUE #633: Setting a player's inventory in the __on_player_respawns event doesn't work
Doing this doesn't do anything:
```
__on_player_respawns(player)->(
   inventory_set(player, 0, 64, 'dirt');
);
```

If you were to schedule a tick, the items would appear in the player's inventory, but if the player is in survival, the items would disappear when the player clicks on the item or tries to place the item.
```
__on_player_respawns(player)->(
  schedule(1, 'giveItem', player);
);

giveItem(player) -> (
  inventory_set(player, 0, 64, 'dirt'); // Item doesn't really exist
);
```

If anyone else is having this problem, doing this works for me:
```
__on_player_respawns(player)->(
  schedule(1, 'giveItem', player);
);

giveItem(player) -> (
  inventory_set(player(player ~ 'name'), 0, 64, 'dirt');
);
```

altrisi:
Fixes #633.

This happens because the `ServerPlayerEntity` that is passed to the event is destroyed during the function the mixin is in, therefore inventory (and other data) ends up being incorrect. The new SPE is generated at around line 465 of `PlayerManager` (yarn).

Therefore this PR changes the target to `TAIL` and passes the newly generated `ServerPlayerEntity` to the Scarpet event instead of the old one.

With this, the code in #633 works properly.

### DIFF
--- a/src/main/java/carpet/mixins/PlayerManager_scarpetEventsMixin.java
+++ b/src/main/java/carpet/mixins/PlayerManager_scarpetEventsMixin.java
@@ -12,9 +12,9 @@ import static carpet.script.CarpetEventServer.Event.PLAYER_RESPAWNS;
 @Mixin(PlayerManager.class)
 public class PlayerManager_scarpetEventsMixin
 {
-    @Inject(method = "respawnPlayer", at = @At("HEAD"))
+    @Inject(method = "respawnPlayer", at = @At("TAIL"))
     private void onRespawn(ServerPlayerEntity player, boolean olive, CallbackInfoReturnable<ServerPlayerEntity> cir)
     {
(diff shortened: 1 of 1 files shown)


gnembon:
that's not fixing the issue, but changing the behaviour of the event.

I didn't noticed replacement of player entity - weird flex, but ok. I think adding a not about that event to the docs is sufficient, with examples how to avoid issues with passing player object to scheduled functions.

altrisi:
Ok, that makes sense.

I made this because at the beginning I thought since it was the `respawn` event, it made sense to get the respawned player instead of the old one there, leaving the old one to other events like the `dies` event.

PR #641, opened 2020-12-26
Removed two useless lambdas in logging/HUDController that caused an error in vscode (?)

woofymax:
title, lambdas were arguments of the LoggerRegistry#log method and had not params, hence they were not applicable arguments.

### DIFF
--- a/src/main/java/carpet/logging/HUDController.java
+++ b/src/main/java/carpet/logging/HUDController.java
@@ -124,7 +124,7 @@ public class HUDController
         });
         return res.toArray(new BaseText[0]);
     }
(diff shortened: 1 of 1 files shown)


altrisi:
Gradle doesn't seem to think the same...

woofymax:
could you elaborate?

altrisi:
imagen

gnembon:
yep, this is incorrect:
```
method Logger.log(Supplier<BaseText[]>) is not applicable
      (argument mismatch; BaseText[] cannot be converted to Supplier<BaseText[]>)
```

PR #643, opened 2020-12-27
Reading/Writing JSON in Scarpet

altrisi:
This PR adds the ability to read and write JSON files in Scarpet.

To write them, it uses the nice, already present, `toJson` methods that were written into `Value`s. To read them, it adds logic to the new `ScarpetJsonDeserializer`, a custom deserializer for Gson.

Other minor things this changes:
- Removes hardcoded file types from the "files API". Before, everything was either `txt` or `nbt`, without using the existing `supportedTypes` map, that provides the extension for the given type. Since getting that is mostly used in `Module`, it has also been moved there
- Removes two useless usages of the old Java `File`. It was literally doing `.toFile()` on one side and `.toPath()` on the other
- Adds `Iterable<Value>` interface to `AbstractListValue`. Since it is iterable

Missing things (from my point of view) to mark this PR as ready (always accepting comments):
- [x] Change JSON deserialization to not return directly `Value`s in `FileModule`, but get converted into them higher in the stack (this is currently "polluting" every class until there)
- [x] ~Maybe move Gson object to a more generic place~ Not really. Now it's only used there
- [x] Test no behaviours were broken when changing the extension system (from initial testing it seems to work)
- [x] Docs ofc

### DIFF
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -305,12 +305,15 @@ for folder listing.
  
 Supported values for resource `type` are:
  * `nbt` - NBT tag
+ * `json` - JSON file
  * `text` - text resource with automatic newlines added
  * `raw` - text resource without implied newlines
- * `folder` - for `list_files` only - indicting folder listing instead of files
- * `shared_nbt`, `shared_text`, `shared_raw`, `shared_folder` - shared versions of the above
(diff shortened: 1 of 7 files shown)


altrisi:
Everything seems to work. Ready for review.

gnembon:
Yes you were correct. the toJson methods were to start the work on json file support. Thanks for doing the work and cleaning up the code.

PR #652, opened 2020-12-30
Update Gradle and reintroduce automatic version in the JSON

altrisi:
This PR updates Gradle and its files to Gradle 6.7. That is the latest version currently included in the Fabric example mod.

It also changes ("modernizes" according to the PR in the example mod repo) some things from the buildscript. I tried to keep all changes done for Carpet when making those changes.

Last but not least, it reintroduces something that was (probably accidentally) overridden in the fourth commit of the repo (c7c0da9c46dae5200ecebcbd9efa53ad4b2b47d9):
Automatic versioning for the `fabric.mod.json` file.

That makes it so the `fabric.mod.json` file dynamically changes its version when building, so there is no need to update it every time. This change, however, required the build date to be removed from the version field, but I readded it in Gradle's buildscript to be generated automatically in the same format. I can add a field if you want to set it statically.

It was not fully removed, just "overridden", because the code meant to replace the `${version}` field was still present, it just couldn't find anything to replace.

BTW, talking about updating Gradle and things, you should check gnembon/fabric-carpet-extension-example-mod#3.


gnembon:
is that some changes that we missed from the fabricexamplemod? I don't want to deviate too much from that honestly (for no reason really, than just depending on one framework in terms of developing the mod).

altrisi:
Yes, all (minus one) changes in this PR makes this match example mod's state.

The only change was keeping your publishing section and versioning.

Autoversioning in JSON, as mentioned, was already in the example mod, just accidentally removed. I had to adapt it slightly to match current versioning (mod_version instead of regular version field, and autoadd the date to the build so it is not needed to be present in the version).

Summary of changes in respect to the example mod:
- Your publishing section
- Adding MC version to file name (and now date automatically too) 
- Using `mod_version` instead of `version` for the JSON field

gnembon:
perfect :)

PR #654, opened 2021-01-02
Add the `lightEngineMaxBatchSize` rule

Andrews54757:
Later versions of the game which have a separate light engine thread have an arbitrarily small maximum light task batch size of `5` set by Mojang.

A smaller batch size generally (simplification) means that the light engine takes longer to do its calculations per task, and a larger batch size decreases that time per task correspondingly. This means that the light engine can \*generally\* process more light tasks with a higher maximum batch size than with a lower batch size. ([More Information](https://github.com/Andrews54757/light-batch/blob/fabric_1.16.x/README.md#suspected-mechanism-for-why-it-works) from the LightBatch mod readme)

This behavior is especially useful for 1.14-1.16 technical servers, as when the light thread becomes overloaded by machines such as worldeaters, chunk loading is stalled which can lead to server crashes via watchdog.

Increasing the batch size via these rules can alleviate this issue by increasing the threshold at which the light engine is overloaded (It does not completely prevent light suppression, it is still possible to overload the light thread though it requires more tasks).

This rule changes the maximum light task batch size by Minecraft's LightingProvider's `setTaskBatchSize(int size)` method. This means that no additional mixins are required to support these rules (though needed to get values). The vanilla game uses this method at the start of world load to temporarily set the light batch size to 500 to facilitate in faster world load times, but sets it to 5 afterwords. To preserve that behavior, modified rules will only be applied after world load has been complete.

Three rules are proposed in this pull request:

`lightBatchSizeOverworld`
`lightBatchSizeNether`
`lightBatchSizeEnd`

All have the same function, but apply only to their respective dimensions.

Additional Information:

* PR tested in 1.16.4. (java 1.8)
* Light batch functionality tested and used from 1.15-1.16 with the [LightBatch mod](https://github.com/Andrews54757/light-batch) on multiple technical servers for around 5-6 months without experiencing any problems.
* Accordingly (to @eatmyvenom), paper/spigot has recently changed maximum light batch sizes as well.
* Is a "solution" to related issue https://github.com/jellysquid3/lithium-fabric/issues/37
* Does not break compatibility with other mods


James103:
Does this work with the Phosphor mod?

Andrews54757:
Yes, it adds no additional mixins or access wideners so it will not cause any conflicts.

gnembon:
All is good - have seen that '5' a few times and have been filddling with that myself. 

But you need to provide a VERY strong case to introduce 3 separate rules where:
 - dimensions shoudn't matter - game should function similarily regardless where you are.
 - number of dimensions now is not 3 and not limited. How you would change that in a custom dimension then?
  - why polluting config space with 3 settings that do the same thing?
  
  Otherwise, am postulating one new rule, and call it less technically what's its going (batch size - what's the batch size), call it more what it directly relates to - lightConcurrency? lightEngineConcurrency?

Andrews54757:
@gnembon I also think that this implementation is ideal. I would love it if it was just a single config. Meanwhile, I still want to be able to set the batch sizes for each dimension independently like with the original lightbatch mod, as the dimensions have different rules for the light engine.

This would be trivially easy were this to be implemented as a command, but, I didn't think that such a small feature would be important enough to warrant it. 

My original plan was to add argument functionality to carpet rules, which would probably be of good use to future features as well. However, doing that would have required major changes to the codebase which I was trying to avoid.

So instead of complicating things, I thought that having three separate rules was the simpler idea - after all, it's not entirely unprecedented in carpet with rules such as `portal[GAMEMODE]Delay`.

However, I understand why a minimalist solution would be preferable. I can change it to be a universal config.

P.S. `lightConcurrency` and `lightEngineConcurrency` does sound audibly better than `lightBatchSize`, but it is not correct and naming it as such would be misleading. There is only one light engine thread per dimension, and it cannot do concurrent execution within the thread -- a larger batch size only means that the DynamicGraphMinFixedPoint datastructure will contain more changed states when it recalculations are run on it.

Andrews54757:
For now, I propose following changes: (though it requires more work).

1. Make it so that there is only one config `lightEngineMaxBatchSize` that applies to all dimensions. Useful for server admins because of setdefault behavior.
2. Add a `/light` command (which can be toggled as well) (If dimension arg is omitted player's current dimension is used.)
      - `/light batch ([Value]) ([Dimension])` -> gets/sets the light engine batch sizes for the dimensions independently. Useful for tech servers.
      - `/light status [Dimension]` -> Gets the status of the light thread (busy/not busy, tasks queued). Useful for diagnosing light-suppression.
      - `/light pause [Dimension]` -> Pauses the light engine.
      - `/light stop [Dimension]` -> Stops the light engine (does not preserve queue)
      - `/light run [Dimension]` -> unpauses/runs the light engine.
      - `/light log [Dimension]` -> Subscribe/unsubscribe to light engine log
      - `/light clear [Dimension]` -> Clears queued updates
      - `/light enable [skylight|blocklight] [Dimension]` -> Enables skylight/blocklight engine
      - `/light disable [skylight|blocklight] [Dimension]` -> Disables skylight/blocklight engine
(A lot of the commands here are inspired & will have similar functionality to vktec's old LightYeet mod, except that instead of  
using the temp batch queue, it uses the real light engine queue (aka the mailbox) and doesn't toss critical updates)

gnembon:
sounds good - also having a light engine contol/debug via command sounds fitting here. I would suggest to limit the scope of this PR to that one extra settings rule, and create a separate PR with the `/light` command.

Andrews54757:
@gnembon Ok. This PR is good and ready to merge then.

PR #655, opened 2021-01-03
Added nbt_storage() scapet function

BisUmTo:
Allows you to easly manage vanilla data command storage

### DIFF
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -540,3 +540,8 @@ Available options in the scarpet app space:
  
  Scarpet related properties
  * `scarpet_version` - returns the version of the carpet your scarpet comes with.
+
+## Storage
+
+### `storage()`, `storage(key)`, `storage(key, nbt)`
+Displays or modifies individual storage nbt tags. With no arguments, returns the list of current storages. With specified `key`, returns the `nbt` associated with current `key`, or `null` if storage does not exist. With specified `key` and `nbt`, sets a new `nbt` value, returning previous value associated with the `key`.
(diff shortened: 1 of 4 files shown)


altrisi:
I think you forgot to add the imports in the PR.

gnembon:
all good, just the name 'storage' is very generic and might be missleading. I would suggest making it more..... verbose.
hmm, nbt_storage, tag_storage, nbt_tag_storage, system_tag_storage ? what do you think?

BisUmTo:
`nbt_storage()` would be aligned with the current `'nbt_merge'` entity property and the future `nbt_modify()` function...

In the future I imagine something like:
`nbt_modify(nbt_storage('name'),'Path', 'append', '{nbt:value}')`
`nbt_modify(e~'nbt','Path', 'set', '{nbt:value}')`

gnembon:
ok then - please change to 'nbt_storage'.

couple of last things: 
 - specify in the docs that this is the storage used by command system, shared with datapacks etc. I think its also persistent (might be wrong on that)
  - add the new call to notepad++ and idea keyword list.

docs/scarpet/api/Auxiliary.md
```diff
@@ -540,3 +540,9 @@ Available options in the scarpet app space:
  
  Scarpet related properties
  * `scarpet_version` - returns the version of the carpet your scarpet comes with.
+
+## NBT Storage
+
+### `nbt_storage()`, `nbt_storage(key)`, `nbt_storage(key, nbt)`
+Displays or modifies individual storage NBT tags. With no arguments, returns the list of current NBT storages. With specified `key`, returns the `nbt` associated with current `key`, or `null` if storage does not exist. With specified `key` and `nbt`, sets a new `nbt` value, returning previous value associated with the `key`.
+NOTE: This NBT storage is shared with all the scripts of the entire server and is persistent between restarts and reloads. You can also access this NBT storage with the `/data <get|modify|merge> storage <key> ...` command.
```
gnembon, line 548:
```suggestion
NOTE: This NBT storage is shared with all vanilla datapacks and scripts of the entire server and is persistent between restarts and reloads. You can also access this NBT storage with vanilla `/data <get|modify|merge> storage <key> ...` command.
```

gnembon (approved):

PR #656, opened 2021-01-04
Adding matrix support to scarpet

Ghoulboy78:
This is gonna be so cool!

### DIFF
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -540,3 +540,7 @@ Available options in the scarpet app space:
  
  Scarpet related properties
  * `scarpet_version` - returns the version of the carpet your scarpet comes with.
+
+## Matrix operations
+
+Matrices are important things to have and to be able to manipulate in a programming language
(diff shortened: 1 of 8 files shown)


James103:
Can you remove the first two unrelated commits from this PR, leaving just the last two commits?

Ghoulboy78:
oh wow lol i didnt notice them there

Ghoulboy78:
they dont appear to do anything,

gnembon:
dude pls no.
- Think about the big picture, not just 2d arrays: image support, tensor support. 
- Don't spam namespace with single-use functions (identity) - API design needs to survive a few interations of functinonality. Consult with me when I am back or let me start it and continue with extra beef. Not now though.
- in any case, try using libraries - I woudn't use tensorflow right away, but there are a few smaller ones that seem sufficient. If libraries are not adequate - there should be a good reason to reinvent the wheel.

Closing before it starts to grow - but you can keep it in your pocket and we can go back to that later.

Ghoulboy78:
I couldnt find any libraries online tho mate...

PR #661, opened 2021-01-07
Added two quality of life structure block settings

BeetMacol:
`structureBlockLimit` (int, default: `48`): changes structure block limit of each axis.
`structureBlockIgnoreAir` (bool, default: `false`): changes structure block ignored block to air.

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -518,6 +518,20 @@ public class CarpetSettings
     public static int structureBlockLimit = 48;
 
     @Rule(
+            desc = "Makes air ignored when saving a structure",
+            extra = {"If true, the ignored block will be",
+                    "set to air instead of structure void,",
+                    "and air will not be saved to the NBT file.",
+                    "Remember, that it will also make structure",
+                    "void save to the NBT file.",
+                    "This setting can save a lot of disk space without",
+                    "replacing all of the air with structure void"
+            },
+            category = CREATIVE
+    )
+    public static boolean structureBlockIgnoreAir = false;
+
+    @Rule(
             desc = "Customizable forceload chunk limit",
             options = {"256"},
             category = CREATIVE,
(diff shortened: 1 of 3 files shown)


BeetMacol:
A maximum value for `structureBlockLimit` should be considered.

src/main/java/carpet/CarpetSettings.java
```diff
@@ -509,6 +509,13 @@ public String description() {
     )
     public static int fillLimit = 32768;
 
+    @Rule(
+            desc = "Customizable structure block limit",
+            options = {"48", "124", "256"},
```
James103, line 514: Why did you have the middle option be 124 instead of 128?
```suggestion
            options = {"48", "128", "256"},
```

BeetMacol: Oh, that's a typo, sorry for that.

BeetMacol:
I accidentally left a bug there, and the limit only works after server restart. I will fix it in a few minutes.

BeetMacol:
After a few more tests it works again even though the code is the same. I think, that `@ModifyConstant` replaced the constant with a constant copy of `CarpetSettings.structureBlockLimit` instead of a normal pointer the time it didn't work. I'm not sure why that happened, but I will leave that mixin the way it is, because it works the same way `FillCommandMixin` does in `fillLimit`.

BeetMacol:
Ok, I get it now. Player can't set a higher value, if carpet is not installed on client, because before the player closes the Structure Block GUI, it changes it back to 48 clientside and sends a packet with the limited value. You can still set it with the `/data` command if the mod is only installed on the server.

altrisi:
Why are you merging into `1.17` instead of `master` branch?

Also, about `structureBlockIgnoreAir`, I think it should support more blocks, maybe even specifying any block. (for instance, you may want to ignore cave air instead, or any other block really). It could only suggest only `structure_void` and `air`, but could accept anything. What do you think?

(could be done by something like `Registry.BLOCK.get(new Identifier("id"))`, and validated in a Validator)

BeetMacol:
It's a good idea. Because of issues with the limit rule and the fact I did use the 1.17 branch instead of the main one, I decided to redo the pull request in the near future.

PR #662, opened 2021-01-10
Fixed null acting as a string with matching operator (~) and split()

replaceitem:
Fixes an issue where `null~'string'` is searching for a substring in the actual string version of `null`, which meant that:
`null~'nu'` -> `'nu'`
`null~'l'` -> `'l'`
`null~'null'` -> `'null'`
`null~'hello'` -> `null`
Pretty sure this is not inteded, `null` represents _nothing_, and _nothing_ should not contain anything

### DIFF
--- a/src/main/java/carpet/script/language/DataStructures.java
+++ b/src/main/java/carpet/script/language/DataStructures.java
@@ -9,7 +9,6 @@ import carpet.script.value.LContainerValue;
 import carpet.script.value.LazyListValue;
 import carpet.script.value.ListValue;
 import carpet.script.value.MapValue;
-import carpet.script.value.NullValue;
 import carpet.script.value.NumericValue;
 import carpet.script.value.StringValue;
 import carpet.script.value.Value;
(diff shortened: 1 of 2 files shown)


altrisi:
Yeah, there are many things that just convert `null` to `'null'`.

As another example, I've just found that `split(null)` returns `['n', 'u', 'l', 'l']` instead of something like `null` or `[]`

replaceitem:
Ok, ill just add that here as well then

replaceitem:
Hmmm, but what makes more sense, returning `null` or `[]`?

altrisi:
Oh, I was going to do that in a PR about `split()`. I'd say empty list, since split is expected to return a list, but not sure.

gnembon:
yeah, splitting of `null` should not return anything meaningful. The default variable's behaviour mimics string values which is correct in general, but there are lots of cases where it doesn't make too much sense. Agree to all of the above and probably there is plenty more of cases where that can be improved.

src/main/java/carpet/script/language/DataStructures.java
```diff
@@ -58,11 +59,13 @@ else if (lv.size() == 2 && lv.get(1) instanceof ListValue)
             String hwat;
             if (lv.size() == 1)
             {
+                if(lv.get(0) instanceof NullValue) return new ListValue(new ArrayList<>());
```
gnembon, line 62: better to use lv.get(0).isNull() since that 'may' change in the future (if more `null` behaving values there are).

PR #664, opened 2021-01-10
Add three Structure Block related settings

BeetMacol:
This is a redo of my other pull request (#661) I cancelled.
`structureBlockOutlineDistance` [client-side] (default: `96`) - allows extending the render distance of the structure outline.
`structureBlockIgnored` (default: `minecraft:structure_void`) - customizes the block ignored when saving structure using Structure Block. Added an enchantment proposed by @altrisi in the previous attempt, thanks.
`structureBlockLimit` (default: `48`) - changes structure block limit of each axis (has to be installed on client of the player changing the value).


src/main/resources/carpet.mixins.json
```diff
@@ -4,212 +4,177 @@
...
+    "PistonBlockEntityRenderer_movableTEMixin",
+    "PlayerListHudMixin",
+    "RenderTickCounter_tickSpeedMixin",
     "SoundEngineInt_cleanLogsMixin",
-    "SoundSystem_cleanLogsMixin"
+    "SoundSystem_cleanLogsMixin",
+    "WorldRenderer_creativeNoClipMixin",
+    "WorldRenderer_fogOffMixin",
+    "WorldRenderer_pausedShakeMixin",
+    "WorldRenderer_scarpetRenderMixin"
```
altrisi, line 177: Looks like you sorted the mixins file

src/main/java/carpet/CarpetSettings.java
```diff
@@ -739,4 +741,47 @@ public static void changeSpawnSize(int size)
...
+            if (!Registry.BLOCK.getIds().contains(Identifier.tryParse(newValue))) {
+                Messenger.m(source, "Unknown block '" + newValue + "'.");
+                return null;
+            }
+            return newValue;
+        }
+    }
+    @Rule(
+            desc = "Changes the block ignored by the Structure Block",
+            options = {"minecraft:structure_void", "minecraft:air"},
```
altrisi, line 772: I think you can use non-namespaced ids as long as you are in the `minecraft` namespace, which may look better.

altrisi (commented):
Hi! It's me again!

src/main/resources/carpet.mixins.json
replaceitem: I had that too with the Minecraft Dev idea plugin when generating accessor and invoker mixins

src/main/java/carpet/mixins/UpdateStructureBlockC2SPacketMixin.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Inject(
+            method = "read",
+            at = @At("TAIL")
+    )
+    private void structureBlockLimitsRead(PacketByteBuf buf, CallbackInfo ci) {
+        BlockPos previousOffset = offset;
+        BlockPos previousSize = size;
+        try {
+            // This will throw an exception if carpet is not installed on client
```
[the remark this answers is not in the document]
BeetMacol: > Also Carpet changes should always be behind a default rule value check, in this case both when sending and receiving.

I thought using `try` is easier than using some buf functions to check that. Will change it later.  
I know that but I didn't want to make another carpet setting to allow setting it to more than 127 (byte limit). I could just do that only if `structureBlockLimit` is more than 127, but when user tries to set it to high value, it will bring it back to 0 instead of the limit value.

src/main/resources/carpet.mixins.json
BeetMacol: Didn't notice that. I will change it back in the next commit.

src/main/java/carpet/CarpetSettings.java
BeetMacol: Vanilla auto-completion adds the `minecraft` namespace, and that's why I did it that way.

altrisi: Well, Carpet usually removes the namespace if it equals `minecraft` (at least in Scarpet), that's why I said that. I suppose it doesn't really matter.

See, for example, `carpet.script.value.NBTSerializableValue#L102` or `carpet.script.value.ValueConversions#L176-187`

https://github.com/gnembon/fabric-carpet/blob/a0067d0d6122dcd5394eba1e76f1119773fd45b6/src/main/java/carpet/script/value/NBTSerializableValue.java#L102-L104

https://github.com/gnembon/fabric-carpet/blob/08adaaeffd3e7f18026fa083f1a8f1036090a7a5/src/main/java/carpet/script/value/ValueConversions.java#L176-L187

gnembon:
Haven't looked into the PR yet, but am curious what happens if the connected vanilla client opens a SB that exceeds 48 blocks or a world is opened in vanilla after its been run with carpet and has some SB saved with sizes > 48 blocks?

I know setting it > 48 blocks is voluntary, but we don't want to cause any corruption even in those cases.

BeetMacol:
> what happens if the connected vanilla client opens a SB that exceeds 48 blocks

They see fake values of 48 because the constants in `StructureBlockBlockEntity#fromTag` are not replaced. When they click done or enter, the structure block values will go back to the vanilla limit.  
The vanilla players will see the outline normally the way carpet players see it.

> a world is opened in vanilla after its been run with carpet and has some SB saved with sizes > 48 blocks

In that case the values will change to 48 during world load because of the `fromTag` method i mentioned earlier. They are stored as integers, so it won't cause any harm.

gnembon:
fair enough - makes sense.

src/main/java/carpet/CarpetSettings.java
```diff
@@ -739,4 +741,47 @@ public static void changeSpawnSize(int size)
...
+                    "Setting 'structureBlockIgnored' to air is recommended",
+                    "when saving massive structures.",
+                    "Required on client of player editing the Structure Block.",
+                    "'structureBlockOutlineDistance' may be required for",
+                    "correct rendering of long structures."},
+            options = {"48", "96", "192", "256"},
+            category = {CREATIVE, CLIENT},
+            strict = false
+    )
+    public static int structureBlockLimit = 48;
```
gnembon, line 757: I don't think it makes sense for these value to be less than vanilla, meaning I suggest to add validator making sure its at least 48. Not sure about the upper bound limit - because whatever it would be set to - there will always be folk complaining its not as high as they wanted to be. But at least >=48 in this case.

src/main/java/carpet/CarpetSettings.java
gnembon: well, that's scarpet - where 'minecraft:' namespace is implied. As long as it behaves similarly to other vanilla commands like `/setblock` I am fine with

src/main/java/carpet/CarpetSettings.java
```diff
@@ -739,4 +741,47 @@ public static void changeSpawnSize(int size)
...
+
+    @Rule(
+            desc = "Customizable Structure Block outline render distance",
+            extra = "Required on client to work properly",
+            options = {"48", "96", "192"}, // 48 is the default structure block limit, and 192 is 12 chunks
+            category = {CREATIVE, CLIENT},
+            strict = false,
+            validate = Validator.NONNEGATIVE_NUMBER.class
+    )
+    public static double structureBlockOutlineDistance = 96d;
```
gnembon, line 786: again like with the reach limits - not sure why you would like to suggest the option to shorten that distance if vanilla is 96. I think the ony usecase is to extend that. 

Since its purely cosmetic value I woudn't be limiting it to >=96 here since using 0 may also have some usecases (I don't want to see these lines), but NONNEGATIVE_NUMBER (from built-in validators) can be used here.

src/main/java/carpet/mixins/StructureBlockBlockEntityMixin.java
```diff
@@ -0,0 +1,77 @@
...
+            CarpetSettings.impendingFillSkipUpdates = false;
+        }
+    }
+
+
+    @ModifyConstant(
+            method = "fromTag",
+            constant = @Constant(intValue = 48)
+    )
+    private int positiveLimit(int original) {
```
gnembon, line 45: If there are other mods already that modify that constant - its a guaranteed mod incompatibility. Its fine if you are not aware of any of these, so its just a note, not a request to change.

src/main/java/carpet/mixins/StructureBlockBlockEntityMixin.java
```diff
@@ -0,0 +1,77 @@
...
+    @ModifyArg(
+            method = "saveStructure(Z)Z",
+            at = @At(
+                    value = "INVOKE",
+                    target = "Lnet/minecraft/structure/Structure;saveFromWorld(Lnet/minecraft/world/World;Lnet/minecraft/util/math/BlockPos;Lnet/minecraft/util/math/BlockPos;ZLnet/minecraft/block/Block;)V"
+            ),
+            index = 4
+    )
+    private Block ignoredBlock(Block original) {
+        return Registry.BLOCK.get(Identifier.tryParse(CarpetSettings.structureBlockIgnored));
```
gnembon, line 66: You should statically parse that in the settings whenever that's changed and here just serve that Block instance, otherwise this will be super laggy.

Example would be commandScriptACE that parses the string value and stores ready runPermissionLevel so it doesn't need to be parsed every time script is executed.

gnembon:
Looks good in general - I would only suggest providing reasonable bounds for the new customizable limits as well as accelerate resolution of the ignored block in settings.

src/main/java/carpet/CarpetSettings.java
BeetMacol: Already used the `NONNEGATIVE_NUMBER` validator there.  
I will remove the "46" value proposal.

src/main/java/carpet/mixins/StructureBlockBlockEntityMixin.java
BeetMacol: Well it is unlikely for an other mod to change that exact value as it is only used for structure block reading. I can't see any better solution for that problem so I will leave it as it is for now.

src/main/java/carpet/mixins/StructureBlockBlockEntityMixin.java
BeetMacol: Well, it will only be parsed when the player saves a structure which is not that laggy in my opinion, but I will add that optimisation.

src/main/java/carpet/CarpetSettings.java
```diff
@@ -739,4 +744,61 @@ public static void changeSpawnSize(int size)
...
+
+    public static class StructureBlockIgnoredValidator extends Validator<String> {
+
+        @Override
+        public String validate(ServerCommandSource source, ParsedRule<String> currentRule, String newValue, String string) {
+            if (!Registry.BLOCK.getIds().contains(Identifier.tryParse(newValue))) {
+                Messenger.m(source, "Unknown block '" + newValue + "'.");
+                return null;
+            }
+            structureBlockIgnoredBlock = Registry.BLOCK.get(Identifier.tryParse(newValue));
```
altrisi, line 782: You could simplify the validator to prevent polling the registry twice with something like (note: written in browser and missing imports):

```suggestion
            Optional<Block> ignoredBlock = Registry.BLOCK.getOptional(Identifier.tryParse(newValue));
            if (ignoredBlock.isEmpty()) {
                Messenger.m(source, "Unknown block '" + newValue + "'.");
                return null;
            }
            structureBlockIgnoredBlock = ignoredBlock.get();
```

src/main/java/carpet/CarpetSettings.java
```diff
@@ -739,4 +744,61 @@ public static void changeSpawnSize(int size)
             category = {SURVIVAL, CLIENT}
     )
     public static boolean cleanLogs = false;
+
+    public static class StructureBlockLimitValidator extends Validator<Integer> {
+
+        @Override public Integer validate(ServerCommandSource source, ParsedRule<Integer> currentRule, Integer newValue, String string) {
+            return (newValue >= 46) ? newValue : null;
```
gnembon, line 751: Did I say 46? I definitely meant 48. Can't imagine why that would be lower than that

src/main/java/carpet/CarpetSettings.java
gnembon: I woudn't be worried about polling registries twice since its just check when changing the setting, but your code is just less hacky (using appropriate contract of the registry.

src/main/java/carpet/CarpetSettings.java
BeetMacol: For some reason I keep messing simple integer recently. That's a typo.

src/main/java/carpet/CarpetSettings.java
```diff
@@ -739,4 +745,62 @@ public static void changeSpawnSize(int size)
...
+
+    public static class StructureBlockLimitValidator extends Validator<Integer> {
+
+        @Override public Integer validate(ServerCommandSource source, ParsedRule<Integer> currentRule, Integer newValue, String string) {
+            return (newValue >= 48) ? newValue : null;
+        }
+
+        @Override
+        public String description() {
+            return "You have to choose a value greater or equal to 46";
```
gnembon, line 757:
```suggestion
            return "You have to choose a value greater or equal to 48";
```

gnembon:
looks good so far - will merge/test that providing there is no Mojupdate tomorrow.

gnembon (approved):

Andrews54757:
I've been trying to get a 1.16.5 instance running with my mods today, and for *some* reason when I try to start the game, it would crash and produce this error message:

Screen Shot 2021-02-08 at 11 42 09 PM

Yes, just this, nothing more, not even in the logs/crash reports -- this is the sort of errors you see in a programmers worst nightmare.

After many many hours of debugging and tests, turns out this only happens when you have tweakaroo + carpet 1.4.25 --For some reason carpet 1.4.24 worked fine. Knowing this, I tested with builds from commits between the version to narrow it down to this one.

I was able to get tweakaroo+carpet to work with a build from right before this commit, though it broke with a build from right after. 

Any ideas why this might be? I've looked through the code and tbh can't find where it leads to the crash.

Andrews54757:
I took another look into tweakaroo's code, and it seems that tweakaroo also modifies structure block behavior. 

https://github.com/maruohon/tweakeroo/commit/70af1736e2b25b702196f14fbf1b61554a8f3d03

BeetMacol:
It seems like tweakeroo has TWEAK_STRUCTURE_BLOCK_LIMIT... I will look into it as soon as possible.

BeetMacol:
Yes, I think it is attempting to do the same thing my `structureBlockLimit` does. In that commit description it is said it won't work on dedicated servers. Also the limit can't be extended beyond 256 due to the packet sending the data in a single byte.  
Another problem with that "tweak" is that it doesn't work. At least it didn't when I tried it on 1.16.5.  
I will create an issue on Tweakeroo's GitHub.

BeetMacol:
It is already fixed in the latest Tweakeroo version.

altrisi:
Since this is a pretty common topic right now, lets make some relations here too:

Causes #700, #712, TISUnion/Carpet-TIS-Addition#14 and #714.

BeetMacol:
I'm really sorry for causing that many issues. I just wanted to add some useful features, but I probably didn't do enough testing and it didn't turn out too well.

Fallen-Breath:
The modify constant conflict with tis addon is fine imo, it's other carpet addon's duty to remove the duplicated rule if a rule is added to the main fabric carpet imo. Most the related issues are just this
But the vanilla issue is another thing, a more serious thing which needs to be resolved

BeetMacol:
I fixed the vanilla server incompatibility issue. I cannot see how is the Worldedit incompatibility related to this pull request as it was there since December.

PR #666, opened 2021-01-11
[Suggestion] Allow splitting lists by element value

altrisi:
This PR adds the ability to `split()` to run different functions depending on the value it's trying to split.

By default, it will work as it has done until now, but this PR adds technically three (actually one) new split functions:

- Splitting a list will split it in sublists delimited by the elements which equal the delimiter
- Splitting a `null` value will return an empty list regardless of the delimiter. Until now, it would split the `null` string
- Splitting a map will throw an exception, just like `slice`


gnembon:
good idea - lets wait for after tomorrow in case Mojang decides to throw in a snapshot.

gnembon (approved):

altrisi:
Oh, didn't see this was #666

PR #673, opened 2021-01-14
Fix split

altrisi:
Makes it return a list of lists with a single element each without delimiter, as commented.

~WIP because I haven't tested this yet~. ~Edit: And now that I've tested it it definitely doesn't work yet.~


altrisi:
I think this is ready

gnembon (approved):

PR #675, opened 2021-01-14
Fix NbtIo crash when loading appdata

### LINKED ISSUE #674: Server crash with scarpet
Server exception: https://hastebin.com/ulacahoqob.log
Crash report: [crash-2021-01-14_13.14.21-server.txt](https://github.com/gnembon/fabric-carpet/files/5816249/crash-2021-01-14_13.14.21-server.txt)

*Renamed to .txt because gihub doesn't accept .sc*
Script: [random_tick.txt](https://github.com/gnembon/fabric-carpet/files/5816251/random_tick.txt)

Server version: 20w51a
Mods present:
- fabric-api-0.28.3+1.17.jar
- fabric-carpet-20w51a-1.4.22+v210113.jar
- ViaBackwards-3.3.0-20w51a (1).jar
- ViaFabric [1.16.3] {0.2.16-586.1}.jar
- ViaVersion-3.3.0-20w51a.jar

Steps to reproduce:
- Run _**/script load random_tick**_
- Run _**/random_tick setDelay 60**_
- Run _**/script load random_tick**_ again

Expected result:
 ‎ ‎  ‎ ‎‎ ‎\- Either not load the values back because I messed up or work fine

Actual result:
  ‎ ‎ ‎ ‎ \- **Server closed**

altrisi:
Fixes #674.

It seems ~in recent versions (snapshots)~ NbtIo.read(File) is only present in the client side, so it basically crashes the dedicated server.

This PR is more like a workaround than a full fix, that basically mimics the behaviour of that method and calls other functions that are present in both sides. Obviously finding a proper method to call instead may be desirable, but that at least works for now.

This didn't strike before because that function is only called as a fallback for simple tags, which isn't so common.

Edit: This actually also happens in the 1.16 version, thus changed the base. The branch name is therefore misleading, but was to prevent opening another PR.

### DIFF
--- a/src/main/java/carpet/script/bundled/FileModule.java
+++ b/src/main/java/carpet/script/bundled/FileModule.java
@@ -76,9 +76,14 @@ public class FileModule extends Module
         }
         catch (IOException e)
         {
-            try
+            // Copy of NbtIo.read(File) because that's now client-side only
+            if (!file.exists())
             {
-                return NbtIo.read(file);
+                return null;
+            }
+            try (DataInputStream in = new DataInputStream(new FileInputStream(file)))
+            {
+                return NbtIo.read(in);
             }
             catch (IOException ioException)
             {


altrisi:
Actually, this seems to also happen in 1.16.4, let me change the base.

gnembon (approved):

PR #676, opened 2021-01-14
Extra null handling

altrisi:
This PR adds more handling to `null` values. Basically overrides all methods that return `Value` from the `Value` class.

This is ready, but is a draft because some things may break, so I'd like to get feedback.

This changes:

| Function        | Before  | After  |
|-----------------|---------|--------|
| `slice(null,1)` | `'ull'` | Merged in #684 |
| `-null`         | 0       | Merged in #684 |
| `null+2`        | 2       | `null` |
| `null*2`        | 0       | `null` |
| `null-2`        | -2      | `null` |
| `null/2`        | 0       | `null` |

### DIFF
--- a/src/main/java/carpet/script/value/NullValue.java
+++ b/src/main/java/carpet/script/value/NullValue.java
@@ -41,6 +41,26 @@ public class NullValue extends NumericValue // TODO check nonsingleton code
     }
 
     @Override
+    public Value multiply(Value v) {
+        return Value.NULL;
+    }
+
+    @Override
+    public Value add(Value v) {
+        return Value.NULL;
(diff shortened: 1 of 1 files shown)


gnembon:
only problems I can see with arithm ops is that this doesn't make them commutative, like these operators are assumed to be. 

This kind of Null value treatment would pretty much require that NullValue doesn't subclass from NumericValue, but it its own type - not pretending to be '0'.

altrisi:
Part of this PR has been split into #684. May even close depending on the future.

altrisi:
I think I'll close it for now.

PR #678, opened 2021-01-16
Very WIP: Add base for server-side labels and those themselves

### LINKED ISSUE #570: [Scarpet] Labels don't have an alternative for non-carpet clients
Makes some app's features quite useless without it, and also some look weird (mainly ai_tracker's villager things and box labels).

Probably blocks replacing `/track` with `ai_tracker`.

Idea: Named fake (packet only?) armorstands.
With the tags `{CustomName:(name),CustomNameVisible:1,Invisible:1,Marker:1}` it makes them completely invisible to the player, floating, and with the label at their feet so it is (exactly?) at their coords. Those also have interaction disabled, so they don't obstruct the player click/right click actions.

`CustomName` -> Specifies the name (label) of the armorstand. It is saved/retrieved in JSON, but AFAIK it's a `Text`
`CustomNameVisible:1` -> Makes their nametag/label visible.
`Invisible:1` -> Makes them invisible.
`Marker:1` -> Auto-disables gravity, sets their label at their feet and prevents most interactions with them (there are a few bugs where if present in server could interact with some things, but not if only in the client).

Vanilla command to preview:
```
/summon armor_stand ~ ~ ~ {CustomName:'{"text":"I want to be in Carpet!"}',CustomNameVisible:1,Invisible:1,Marker:1b}
```

Scarpet script to preview:
```py
spawn('armor_stand',pos(player()),'{CustomName:\'{"text":"I wanna be in Carpet!"}\',CustomNameVisible:1,Invisible:1,Marker:1b}')
```

altrisi:
Resolves #570.

The reason I'm doing this is mostly for the `ai_tracker` app to be able to correctly replace the `/track` command at some point without losing functionality in Carpetless clients.

As seen in the title, this is very much WIP, since it isn't even actually applied anywhere, but I've opened this PR so I can get feedback or even help/prs before it's completely finished. The way I'm currently testing this (not included in the PR) is by changing the `create_marker` function in `Auxiliary` to create a `CarpetArmorStandLabel` with `CarpetServer.minecraft_server.getPlayerManager().getPlayerList()` as the list of players. Note that this system is NOT intended (at least in the way it currently works) to replace those markers.

This currently holds:
- A class extending `ArmorStandEntity` that initializes it automatically with some parameters and prevents it from saving or even ticking. It also implements what seems to be a fix for the [MC-135809](https://bugs.mojang.com/browse/MC-135809) bug on those armorstands.
- A mixin into `TACS$EntityTracker` to prevent it from sending our entity's packets to players that aren't declared in the set of players that should receive those, by replacing that set with one that is filtered on added
- Removing them from entity selectors
- A base for "`HandledShape`s", where the idea is to add those into a list in the ShapeDispatcher and manage them from there.
  Adding this also helps for (in a future out of the scope of this PR) sending particles repeatedly for Carpetless clients, since those currently get them only once even when specifying a longer duration

This is currently missing:
- Following ~(or testing if any kind of movement works for those entities, since we cancel syncing them)~
- Reworking `ShapeDispatcher` a bit so only a single entity (`HandledShape` from now on) is generated for multiple Carpetless players
- Replacing them (that thing with ids that says that replaces the label with another text, and probably syncing them, _or did I make that up?_)
- Hooking them into the alternative for labels


altrisi:
Implementation concept wasn't really that good, extremely outdated, will reopen (or redo) if I ever go back to this, though having an actual entity in the server is probably not the way to go.

PR #680, opened 2021-01-18
Use WorldRenderEvents if available

ramidzkh:
Closes https://github.com/grondag/canvas/issues/154


gnembon:
well, not really liking the idea of adding a significant dependency (fabric API) (despite being fabric API, but carpet tries to minimize potential vanilla code modifiers, and FAPI is one of them) just to get rendering in another mod to work.

However I am open to relocate the call to rendering shapes into a different place if that's gonna help and not gonna affect them in a bad way (i.e. screw up water redering even more etc.), or you can call it yourself I guess if carpet is present.

DeadlyMC:
But I believe `modCompileOnly` makes it an optional dependency, so its not necessary for people to have the api when they download carpet, so this should be fine right? 
But could still do the relocate call thing tho

altrisi:
In case this would get in, wouldn't it be better to instead check in FLoader API whether the `fabric-rendering-v1` mod is present and is of at least a certain version, instead of relying in exceptions or whether a class exists?

Also, wouldn't it be possible to only depend on the specific FAPI module at compile time instead of the whole API? (I know it is, just I haven't checked whether `rendering` has deps in other modules)

ramidzkh:
@altrisi All your concerns have been addressed
@gnembon Although it is possible for Canvas to hook this (that's how we did mod compat before WorldRenderEvents 😛), mods should try to co-operate with each other instead of expecting other mods to fix it, especially when something as widespread as FAPI makes it easier

gnembon:
well, mods should first take care of first their own functions, principles and paradigms, and then about compatiblity on other mods, unless they belong to the 'kichen sink' category where compatibility is king.
Carpet principle is to operate as close to vanilla as possible so FAPI is considered as any other external library given that philosophy.

Having said that - seems like everything is working a expected in a sense that stuff seem to work just fine with and without the API so I think it should be fine to incorporate it.

src/main/java/carpet/utils/FabricAPIHooks.java
```diff
@@ -0,0 +1,44 @@
...
+import com.mojang.blaze3d.systems.RenderSystem;
+import net.fabricmc.fabric.api.client.rendering.v1.WorldRenderEvents;
+import net.fabricmc.loader.api.FabricLoader;
+import net.fabricmc.loader.api.SemanticVersion;
+import net.fabricmc.loader.api.Version;
+import net.fabricmc.loader.api.VersionParsingException;
+
+public class FabricAPIHooks {
+
+    public static final boolean WORLD_RENDER_EVENTS = hasMod("fabric-rendering-v1", "1.5.0");
```
gnembon, line 13: technical question - does it need to be bound / harcoded to that specific version, or is works as >= ?

src/main/java/carpet/utils/FabricAPIHooks.java
```diff
@@ -0,0 +1,44 @@
...
+        }
+    }
+
+    private static boolean hasMod(String id, String minimumVersion) {
+        return FabricLoader.getInstance().getModContainer(id).map(m -> {
+            Version version = m.getMetadata().getVersion();
+
+            if (version instanceof SemanticVersion) {
+                try {
+                    return ((SemanticVersion) version).compareTo(SemanticVersion.parse(minimumVersion)) >= 0;
```
gnembon, line 36: Ok - got the answer I think - its '>='

src/main/java/carpet/utils/FabricAPIHooks.java
gnembon: all good, see below

build.gradle
```diff
@@ -21,7 +21,7 @@ dependencies {
 	modCompile "net.fabricmc:fabric-loader:${project.loader_version}"
 
 	// Fabric API. This is technically optional, but you probably want it anyway.
-	modCompileOnly "net.fabricmc.fabric-api:fabric-api:${project.fabric_version}"
+	modCompileOnly "net.fabricmc.fabric-api:fabric-rendering-v1:${project.fabric_rendering_v1}"
```
altrisi, line 24: With this I was referring to something more like:

`fabricApi.module("fabric-rendering-v1", project.fabric_version)`

Since that would allow using a global FAPI version in case at some point hooks with other modules were implemented, so there's no need to track individual versions for each module.

YTG1234 (approved):

BeetMacol (approved):
The code looks OK, didn't test in-game

hYdos (approved):

gnembon (approved):

gnembon:
lets give it a shot

Issue #446, opened 2020-09-01
[Scarpet bug/suggestion] Uncatched exceptions (throw) don't throw a stacktrace


altrisi:
Currently, an unhandled `throw()` call terminates the execution of the script, without giving any extra information than changing the return value. In my opinion, it should throw a proper error (with a stacktrace) instead.
As an example of where this could be useful, while `throw()` may not be used too much, in case the `throw()` may come from a library and the function throwing the error being in an event, it can make the function silently fail, since only the return value will change and nothing will be shown or logged anywhere (except if explicitly logged, but the lib may suppose the exception would be caught). In the same case of an external library, but in this case happening in a command and the exception not being thrown in every call, the end-user may be confused by some random return that may be completely unrelated to the command they were typing.
It would also be a nice way to force displaying a stacktrace without trying to execute incorrect code (this is my case, the rest are hypothetical situations).

gnembon:
well, unhandled throw() warrants an error message, but (scarpet) stacktrace will not be able to be retrieved as its tracked only when programming errors are thrown. Simply when the value is caught at the end - it doesn't carry the scarpet stack information, and passing it with every throw() just in case is not a good idea.

I agree that unhandled throws() should result in an error message. exit() is to do that without the error message

gnembon:
well to have the stack exploded by demans, just cause an error like `cause_crash() -> print()`. Will make sure throw reports an appropriate error.

PR #692, opened 2021-02-03
Exception handling features

altrisi:
Since mods (or even resource/data/whatever packs) can add many things that Scarpet can opt to read, and just returning `null` may not be ideal in many cases, this PR allows some, specified exceptions to be caught. Not every exception is catchable, mostly to prevent people just wrapping everything in a big `try` block. At the same time this PR it also overhauls the current exception system a bit.

What's that overhaul?

Now exceptions have two different values: Their "id" and their message. The id is something that should be kind-of general, in order to allow apps to easily filter exceptions in the code. The message is, well, the message that will be shown in the stacktrace when being unhandled. The id is present in the old `_` variable, and in order to make the message useful, it is passed as `_msg` (feel free to ask for a name change).

Not only that, exception handlers now can (and are encouraged to) filter what exceptions to catch by specifying the second parameter of `try` as the id to match before processing the `try` block. While something similar could be achieved in code by just throwing a new one with the same id and message, filtering them there allows the stacktrace (if needed) to point to the original location that threw it.

How does it work?

First of all, this PR converts the `ThrowStatement` into a subclass of `InternalExpressionException` instead of `ExitStatement`. I don't think the subclass is really needed, but I left it just in case there was something somewhere I didn't see, since I suppose that should prevent code relying on IEEs getting broken with the new catchable ones.

Now, in order for those to have stacktrace and all that info, they need to get converted into some sort of `ExpressionException` in `handleCodeException()`. Therefore, the new exception type `ProcessedThrowException` is a subclass of `ExpressionException` that also keeps the Scarpet information. This is the exception type that will be caught in the `try` blocks, therefore `throw` needs to be a `lazyFunction` (no idea why, else it wouldn't be caught if on the same block and line marking would be wrong), making it be always ready to be caught by `try`.

Note that in order to prevent the "stacktrace" to be shown even when being caught, the message of `ExpressionException` is now lazy. Long story short: It has been a pain to try to locate where were exceptions finally handled, when found I lost it once, and after that just found that would add an extra complexity (passing context, token, etc to `CarpetExpressionException`. Basically at construction, a supplier is created with what would be the message and a dummy message is passed to the `RuntimeException`. Then it overrides `getMessage()` to call the supplier.

This PR also contains a lot (if not all) of the functions that (I think) should be catchable for reasons like distributing and mod/datapack/anything not being present, optional compatibility with some of those, etc.
It also changes some of them to "cleaner" code (using optionals in DefaultedRegistries instead of `v = Registry.BLOCKS.get(id); if(v == Blocks.AIR && id != 'air') throw`).
Also unhandles an exception that shouldn't have ever been handled (`InvalidIdentifierException`), since that is thrown when the identifier (the string like `minecraft:air`) is invalid due to disallowed characters, etc, not when whatever is being searched isn't found.

Exception ids are currently being used from fields in the `ThrowStatement` class in order to make sure all of them are the same for related exceptions.

Docs are updated too.

This works, but may need a bit more testing to make sure nothing is broken.

### DIFF
--- a/src/main/java/carpet/script/Expression.java
+++ b/src/main/java/carpet/script/Expression.java
@@ -19,7 +19,6 @@ import carpet.script.exception.ExpressionException;
 import carpet.script.exception.InternalExpressionException;
 import carpet.script.exception.ResolvedException;
 import carpet.script.exception.ReturnStatement;
-import carpet.script.exception.ThrowStatement;
 import carpet.script.language.Arithmetic;
 import carpet.script.language.ControlFlow;
 import carpet.script.language.DataStructures;
@@ -232,7 +231,7 @@ public class Expression
         if (exc instanceof ExitStatement)
             return exc;
         if (exc instanceof InternalExpressionException)
-            return new ExpressionException(c, e, token, exc.getMessage(), ((InternalExpressionException) exc).stack);
+            return ((InternalExpressionException) exc).promote(c, e, token);
         if (exc instanceof ArithmeticException)
             return new ExpressionException(c, e, token, "Your math is wrong, "+exc.getMessage());
         if (exc instanceof ResolvedException)
@@ -624,9 +623,9 @@ public class Expression
         {
             return exprProvider.get().evalValue(c, expectedType);
         }
-        catch (ContinueStatement | BreakStatement | ReturnStatement | ThrowStatement exc)
+        catch (ContinueStatement | BreakStatement | ReturnStatement exc)
         {
-            throw new ExpressionException(c, this, "Control flow functions, like continue, break, throw or return, should only be used in loops, try blocks, and functions respectively.");
+            throw new ExpressionException(c, this, "Control flow functions, like continue, break or return, should only be used in loops, and functions respectively.");
         }
         catch (ExitStatement exit)
         {
--- a/src/main/java/carpet/script/api/Auxiliary.java
+++ b/src/main/java/carpet/script/api/Auxiliary.java
@@ -20,6 +20,7 @@ import carpet.script.argument.Vector3Argument;
 import carpet.script.exception.ExitStatement;
 import carpet.script.exception.InternalExpressionException;
 import carpet.script.exception.ThrowStatement;
+import carpet.script.exception.Throwables;
 import carpet.script.utils.FixedCommandSource;
 import carpet.script.utils.ScarpetJsonDeserializer;
 import carpet.script.utils.ShapeDispatcher;
@@ -132,7 +133,7 @@ public class Auxiliary {
             Identifier soundName = new Identifier(lv.get(0).evalValue(c).getString());
             Vector3Argument locator = Vector3Argument.findIn(cc, lv, 1);
             if (Registry.SOUND_EVENT.get(soundName) == null)
-                throw new ThrowStatement("No such sound: "+soundName.getPath(), ThrowStatement.UNKNOWN_SOUND);
+                throw new ThrowStatement("No such sound: "+soundName.getPath(), Throwables.UNKNOWN_SOUND);
             float volume = 1.0F;
             float pitch = 1.0F;
             SoundCategory mixer = SoundCategory.MASTER;
@@ -828,7 +829,7 @@ public class Auxiliary {
                     state = ((CarpetScriptHost) c.host).readFileTag(fdesc.getLeft(), fdesc.getRight());
                 } catch (CrashException e)
                 {
-                    throw new ThrowStatement("Couldn't read NBT data", ThrowStatement.NBT_READ_EXCEPTION); 
+                    throw new ThrowStatement("Couldn't read NBT data", Throwables.NBT_READ); 
                 }
 
                 if (state == null) return LazyValue.NULL;
@@ -846,7 +847,7 @@ public class Auxiliary {
                     Throwable exc = e;
                     if(e.getCause() != null)
                         exc = e.getCause();
-                    throw new ThrowStatement("Failed to read JSON file: "+exc.getMessage(), ThrowStatement.JSON_READ_EXCEPTION);
+                    throw new ThrowStatement("Failed to read JSON file: "+exc.getMessage(), Throwables.JSON_READ);
                 }
                 Value parsedJson = gson.fromJson(json, Value.class);
                 if (parsedJson == null)
@@ -940,7 +941,7 @@ public class Auxiliary {
                 state = ((CarpetScriptHost)((CarpetContext)c).host).readFileTag(file, shared);
             } catch (CrashException e)
             {
-                throw new ThrowStatement("Failed to read App data", ThrowStatement.NBT_READ_EXCEPTION);
+                throw new ThrowStatement("Failed to read App data", Throwables.NBT_READ);
             }
             if (state == null)
                 return (cc, tt) -> Value.NULL;
--- a/src/main/java/carpet/script/api/Auxiliary.java
+++ b/src/main/java/carpet/script/api/Auxiliary.java
@@ -19,6 +19,7 @@ import carpet.script.argument.FunctionArgument;
 import carpet.script.argument.Vector3Argument;
 import carpet.script.exception.ExitStatement;
 import carpet.script.exception.InternalExpressionException;
+import carpet.script.exception.ThrowStatement;
 import carpet.script.utils.FixedCommandSource;
 import carpet.script.utils.ScarpetJsonDeserializer;
 import carpet.script.utils.ShapeDispatcher;
@@ -62,6 +63,7 @@ import net.minecraft.text.LiteralText;
 import net.minecraft.text.Text;
 import net.minecraft.util.Identifier;
 import net.minecraft.util.InvalidIdentifierException;
+import net.minecraft.util.crash.CrashException;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.ChunkPos;
 import net.minecraft.util.math.EulerAngle;
@@ -130,7 +132,7 @@ public class Auxiliary {
             Identifier soundName = new Identifier(lv.get(0).evalValue(c).getString());
             Vector3Argument locator = Vector3Argument.findIn(cc, lv, 1);
             if (Registry.SOUND_EVENT.get(soundName) == null)
-                throw new InternalExpressionException("No such sound: "+soundName.getPath());
+                throw new ThrowStatement("No such sound: "+soundName.getPath(), ThrowStatement.UNKNOWN_SOUND);
             float volume = 1.0F;
             float pitch = 1.0F;
             SoundCategory mixer = SoundCategory.MASTER;
@@ -820,7 +822,15 @@ public class Auxiliary {
             Value retVal;
             if (fdesc.getMiddle().equals("nbt"))
             {
-                Tag state = ((CarpetScriptHost) c.host).readFileTag(fdesc.getLeft(), fdesc.getRight());
+                Tag state;
+                try
+                {
+                    state = ((CarpetScriptHost) c.host).readFileTag(fdesc.getLeft(), fdesc.getRight());
+                } catch (CrashException e)
+                {
+                    throw new ThrowStatement("Couldn't read NBT data", ThrowStatement.NBT_READ_EXCEPTION); 
+                }
+
                 if (state == null) return LazyValue.NULL;
                 retVal = new NBTSerializableValue(state);
             }
@@ -833,10 +843,10 @@ public class Auxiliary {
                 }
                 catch (JsonParseException e)
                 {
-                    Throwable exception = e;
+                    Throwable exc = e;
                     if(e.getCause() != null)
-                        exception = e.getCause();
-                    throw new InternalExpressionException("Failed to read JSON file: "+exception.getMessage());
+                        exc = e.getCause();
+                    throw new ThrowStatement("Failed to read JSON file: "+exc.getMessage(), ThrowStatement.JSON_READ_EXCEPTION);
                 }
                 Value parsedJson = gson.fromJson(json, Value.class);
                 if (parsedJson == null)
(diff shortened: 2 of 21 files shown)


gnembon:
I did read the PR, right when it was added, just didn't have time to give it a thought.
Allowing users to handle some system thrown exceptions that are not results of the total fcukup with the code makes sense.
Not sure if having a hierarchy of exceptions like in java or python would be a good place to go (too complex), so suggested flat structure feels right.
Not sure though that having them all separated is the best approach. I would suggest something like a python's ValueError capturing all sorts of wrong value cases, returning `throw('value_error', {'biome' -> 'foobar'})` instead. That should be enough to allow clients to catch them and process them appropriately.
It would makes sense to classify nbt and json errors as something else (data_error, parse_error, 'format_error'?) including details of it in the thrown value.

Value of system exceptions could always be a map object I guess - that would simplify processing.

Also technically with diffenret types of exceptions, it makes sense to allow for multiple catch blocks

`try(expr, exception, catch_expression, exception, catch_expression)`

altrisi:
The reason I opted to go with individual exception names for different things was to stress that the programmer should know, if they are using arguments that may not exist, _what_ will not exist, and anything else failing should not be caught. About python: no idea, don't really know python. But I understand why it could be useful to have some more code-friendly data-structure with the exception. The reason I didn't opt for that was mostly because of already having defined multiple types for multiple data types and because, after all, if the exception is uncaught, (currently) the second value will be the message that will be presented on top of the stacktrace, where I preferred readability instead.

Not sure how should I implement the maps then, I don't really have much reference. If I understood correctly, this would be something like `{required_type -> provided_name}`? Would these be as the message, or as another field?

I may look into inheritance, not sure though. Looks simple-ish, at least not so complicated.

About the multi-catch: That's a great idea I'll have to implement.

Also: Would deprecating `try(expr)`, `try(expr, catch_expr)` and `throw(void)` (those without checking/providing exc type) be a good idea?

altrisi:
Quick preview of hierarchy (actual class, just not implemented yet into exceptions):

```java
public class Throwables {
	private static final Map<Value, Exception> byName = new HashMap<>();
    public static final Exception EXCEPTION               = new Exception("exception", null);
    public static final Exception   VALUE_EXCEPTION       = new Exception("value_exception", EXCEPTION);
    public static final Exception     UNKNOWN_ITEM        = new Exception("unknown_item", VALUE_EXCEPTION);
    public static final Exception     UNKNOWN_BLOCK       = new Exception("unknown_block", VALUE_EXCEPTION);
    public static final Exception     UNKNOWN_BIOME       = new Exception("unknown_biome", VALUE_EXCEPTION);
    public static final Exception     UNKNOWN_SOUND       = new Exception("unknown_sound", VALUE_EXCEPTION);
    public static final Exception     UNKNOWN_PARTICLE    = new Exception("unknown_particle", VALUE_EXCEPTION);
    public static final Exception     UNKNOWN_POI_TYPE    = new Exception("unknown_poi_type", VALUE_EXCEPTION);
    public static final Exception     UNKNOWN_DIMENSION   = new Exception("unknown_dimension", VALUE_EXCEPTION);
    public static final Exception     UNKNOWN_STRUCTURE   = new Exception("unknown_structure", VALUE_EXCEPTION);
    public static final Exception     UNKNOWN_CRITERION   = new Exception("unknown_criterion", VALUE_EXCEPTION);
    public static final Exception   DATA_EXCEPTION        = new Exception("data_exception", EXCEPTION);
    public static final Exception     NBT_READ            = new Exception("nbt_read", DATA_EXCEPTION);
    public static final Exception     JSON_READ           = new Exception("json_read", DATA_EXCEPTION);
    public static final Exception   USER_DEFINED          = new Exception("user_defined_exception", EXCEPTION);
    
    public static class Exception {
    	private final Value name;
    	private final Exception parent;
    	
    	/**
    	 * Creates a new exception without registering it, with an optional parent
    	 * @param name The name/id of the exception
    	 * @param parent An optional Value matching the name of a parent exception.
    	 *               <p>Accepts a {@link NullValue}, but not a <code>null</code>
    	 */
    	public Exception(Value name, Value parent)
    	{
			this.name = name;
			if (!parent.isNull())
			{
				Exception possibleParent = byName.get(parent);
				this.parent = possibleParent == null ? USER_DEFINED : possibleParent;
			} else
				this.parent = USER_DEFINED;
		}
    	
    	/**
    	 * <p>Registers an exception that will be available for lookup and inheritance
    	 * of user exceptions.
    	 * 
    	 * <p>This does NOT expect to allow users to add their own exceptions to the registry.
    	 * @param name This exception's name/id
    	 * @param parent The parent exception, or null if no parent is available
    	 */
    	public Exception(String name, Exception parent)
    	{
    		this.name = StringValue.of(name);
    		this.parent = parent;
    		byName.put(this.name, this);
    	}
    	
    	/**
    	 * <p>Checks whether the given value is an instance of this extension, including checking equality with 
    	 * itself and possible parents.
    	 * @param val The value to check against
    	 * @return Whether or not the given value matches this extension's hierarchy
    	 */
    	public boolean isInstance(Value val) {
    		return name.equals(val) || (parent != null && parent.isInstance(val));
    	}
    }
}

```

altrisi:
I added an exception hierarchy (up to one parent per exception), the multi-catch block and deprecated `try` usages without a filter, since it now would catch more than exceptions just from `throw`. Therefore also deprecated `throw()` without an id.

gnembon:
lol, I was trying to avoid the 'exception explosion', keeping it simple. Having exception hierarchy might be too much and unnecessary. 
For the same reason (and compatiblity) keeping the default type for user thrown exception would be preferable.
I like the idea of user capability to catch stuff, but I don't thing the system needs to be so complex.

src/main/java/carpet/script/Expression.java
```diff
@@ -232,7 +231,7 @@ public static RuntimeException handleCodeException(Context c, RuntimeException e
         if (exc instanceof ExitStatement)
             return exc;
         if (exc instanceof InternalExpressionException)
-            return new ExpressionException(c, e, token, exc.getMessage(), ((InternalExpressionException) exc).stack);
+            return ((InternalExpressionException) exc).promote(c, e, token);
```
gnembon, line 234: good idea

src/main/java/carpet/script/api/Auxiliary.java
```diff
@@ -820,7 +823,15 @@ else if(lv.size()==2)
             Value retVal;
             if (fdesc.getMiddle().equals("nbt"))
             {
-                Tag state = ((CarpetScriptHost) c.host).readFileTag(fdesc.getLeft(), fdesc.getRight());
+                Tag state;
+                try
+                {
+                    state = ((CarpetScriptHost) c.host).readFileTag(fdesc.getLeft(), fdesc.getRight());
+                } catch (CrashException e)
```
gnembon, line 830: How did you get that exception to be thrown here? Ideally all exceptions in this call should have been caught and return null at the end.

gnembon: There are other IOExceptions here in there that should cause the exception to be rethrown, rather than retuning null.

src/main/java/carpet/script/api/Inventories.java
```diff
@@ -119,14 +121,7 @@ public static void apply(Expression expression)
...
             List<Recipe<?>> recipes;
-            try
-            {
-                recipes = ((RecipeManagerInterface) cc.s.getMinecraftServer().getRecipeManager()).getAllMatching(type, new Identifier(recipeName));
-            }
-            catch (InvalidIdentifierException ignored)
-            {
-                return LazyValue.NULL;
-            }
+            recipes = ((RecipeManagerInterface) cc.s.getMinecraftServer().getRecipeManager()).getAllMatching(type, new Identifier(recipeName));
```
gnembon, line 124: not thrown here?

src/main/java/carpet/script/exception/ExpressionException.java
```diff
@@ -8,28 +8,32 @@
...
-        super(makeMessage(c, e, null, message));
-        context = c;
+        this(c, e, null, message);
     }
 
     public ExpressionException(Context c, Expression e, Tokenizer.Token t, String message)
     {
-        super(makeMessage(c, e, t, message));
+        super("Error");
+        lazyStacktrace = () -> makeMessage(c, e, t, message);
```
gnembon, line 29: And specific reason for deferring it?

src/main/java/carpet/script/language/ControlFlow.java
```diff
@@ -42,26 +44,80 @@ public static void apply(Expression expression) // public just to get the javado
 
         expression.addFunction("exit", (lv) -> { throw new ExitStatement(lv.size()==0?Value.NULL:lv.get(0)); });
 
-        expression.addFunction("throw", (lv)-> {throw new ThrowStatement(lv.size()==0?Value.NULL:lv.get(0)); });
+        expression.addLazyFunction("throw", -1, (c, t, lv)-> 
+        {
+            switch (lv.size()) 
+            {
+                case 0:
+                    c.host.issueDeprecation("throw() (without id)");
```
gnembon, line 52: throwing a 'null' by default is actually pretty useful, especially if you don't care what you pass. makes it like a more generalized 'break()'

src/main/java/carpet/script/language/ControlFlow.java
```diff
@@ -42,26 +44,80 @@ public static void apply(Expression expression) // public just to get the javado
...
                 return (c_, t_) -> retval;
             }
-            catch (ThrowStatement ret)
+            catch (ProcessedThrowStatement ret)
             {
+                if (ret.exception.isError())
+                    throw ret;
                 if (lv.size() == 1)
+                {
+                    c.host.issueDeprecation("try(expr) (without filter and catch expression)");
```
gnembon, line 80: this is also fine - what if you want to simply catch a user thrown exception.

src/main/java/carpet/script/language/ControlFlow.java
```diff
@@ -42,26 +44,80 @@ public static void apply(Expression expression) // public just to get the javado
...
                     return (c_, t_) -> Value.NULL;
+                }
+                // This is outside since it is used in try(expr,catch_expr)
                 LazyValue __ = c.getVariable("_");
-                c.setVariable("_", (__c, __t) -> ret.retval.reboundedTo("_"));
-                Value val = lv.get(1).evalValue(c, t);
+                c.setVariable("_", (__c, __t) -> ret.exception.getValue().reboundedTo("_"));
+                LazyValue __msg = c.getVariable("_msg");
+                c.setVariable("_msg", (__c, __t) -> StringValue.of(ret.message).reboundedTo("_msg"));
```
gnembon, line 87: yeah - with value cointaining the information about the exception, there is no need for message (_msg) - the value thrown is the message, which could be anything.

src/main/java/carpet/script/value/FunctionValue.java
```diff
@@ -228,15 +227,6 @@ public LazyValue lazyEval(Context c, Integer type, Expression e, Tokenizer.Token
...
             retVal = returnStatement.retval;
         }
-        catch (ThrowStatement throwStatement) // might not be really necessary
-        {
-            retVal = throwStatement.retval;
-            rethrow = true;
-        }
-        if (rethrow)
-        {
-            throw new ThrowStatement(retVal);
```
gnembon, line 238: this was I guess to add potentially stack information, but never happen.

gnembon:
After reading through the PR, I think it would be even better if only a few types could be returned like value_error, io_error, encoding_error (format_error, etc, for tags and jsons). That would also simplify work with extensions wanting to throw their exceptions for different errors happening.

this way the message is just part of the error thrown, not its own '_msg' thingy.

src/main/java/carpet/script/api/Inventories.java
altrisi: At least not `InvalidIdentifierException`, since that one is for "syntax" errors when creating the `Identifier` instance, not the specific identifier not being present in there (eg having two colons or some other disallowed character).

It returns an empty list if none match (it is a custom implementation `RecipeManager_scarpetMixin`).

Now thinking, maybe `InvalidIdentifierException` would be good to be caught and be rethrown as an `InternalExpressionException`/`ExpressionException` every time when creating an `Identifier`, but it can happen basically anywhere, so it would be a bit tedious to add it to each and every function. Maybe could be added to `handleCodeException()`, like arithmetics.

src/main/java/carpet/script/api/Auxiliary.java
altrisi: `NbtIo#read(DataInput, depth, PositionTracker)` directly throws the crash. It's not when the file doesn't exist or similar, but when it is corrupted, invalid or whatever. Should I catch it in `FileModule` instead? The problem with returning `null` is it cannot distinguish between invalid data and not found files.

src/main/java/carpet/script/exception/ExpressionException.java
altrisi: Well, the `makeMessage()` function calls `errorSnooper.apply()`, which directly notifies the source. I had trouble finding a way to change it not to throw in potentially handled ones, and this worked. There are probably many other ways to do it.

src/main/java/carpet/script/language/ControlFlow.java
altrisi: Will revert that then.

src/main/java/carpet/script/language/ControlFlow.java
altrisi: Ok, then I'll undeprecate (or whatever you call that) it and filter it to user defined if not specified.

src/main/java/carpet/script/value/FunctionValue.java
altrisi: Hmm... Wouldn't it need to catch `ExpressionException` then (don't really know, just asking)? May try to see what can I do with this then.

src/main/java/carpet/script/api/Auxiliary.java
gnembon: ok

src/main/java/carpet/script/exception/ExpressionException.java
gnembon: ok, makes sense

src/main/java/carpet/script/value/FunctionValue.java
gnembon: Yeah, that's fine, just asking.

altrisi:
I've just remember this also actually implements #446

PR #696, opened 2021-02-04
Fix rules-to-wiki and generalize some code for extension's equivalents

altrisi:
Move it later, when the game is ready to be touched.

It previously was in pre-launch because Carpet didn't really touch the game at `onGameStarted`, now it references `Blocks` statically so it basically initializes that and makes a chain reaction until it hits a NPE. This means it'll have to wait for DFU and will take a bit more time, but it should be fine (it's been quite fast on the testrun at https://github.com/altrisi/fabric-carpet/actions/runs/537672403.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -159,7 +159,7 @@ jobs:
             "id": "carpet",
             "version": "1.4.11",
             "entrypoints": {
-              "preLaunch": [ "carpet.utils.RulePrinter" ]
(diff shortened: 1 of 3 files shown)


altrisi:
DON'T MERGE YET. I've just realized something weird with the output (https://github.com/altrisi/fabric-carpet/wiki/Current-Available-Settings)

altrisi:
All done, builds correctly, creates wiki correctly.

gnembon (approved):

PR #704, opened 2021-02-08
Add `pingPlayerListLimit` setting

Andrews54757:
The [ServerListPing Protocol](https://wiki.vg/Server_List_Ping) is used by the client when one is in the multiplayer menu. With it, the client is able to retrieve the server's player number, player limit, protocol version, favicon, and MOTD. One feature of this protocol, is the ability to retrieve a "sample" of up to twelve players currently online on the server. This information can be queried not just from the Minecraft client, but also from [external websites](https://mcsrvstat.us/) and [tools](https://andrews54757.github.io/MCObserver/) regardless of server whitelist status.

On the vanilla client, this info is displayed by hovering over the player number info, like so:

Screen Shot 2021-02-08 at 4 58 18 PM

The current behavior poses two problems:
* Due to privacy concerns, not all server operators may want to broadcast player usernames to the public (addendum: disabling query just for this would be sad).
* For servers that make use of the list, the sample limit of twelve makes some players not show on the list when there are more than twelve players.

To solve these problems, I propose a new rule/setting for carpet called `pingPlayerListLimit`. With this rule, server administrators will be able to set the amount of players shown in the sample. For those who want to hide player-lists, they may set it to 0, while those who want to show more may set it to a higher value.

Tests/Examples:

*With `pingPlayerListLimit` set to 20*
Screen Shot 2021-02-08 at 5 09 21 PM

*With `pingPlayerListLimit` set to 1*
Screen Shot 2021-02-08 at 5 11 29 PM

*With `pingPlayerListLimit` set to 0* (nothing shows)

Screen Shot 2021-02-08 at 5 12 15 PM

**Note: This PR uses a ModifyConstant which means this feature will break if other mods also modify it. However, I don't think there is another easy way to do it besides this.**

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -531,6 +531,14 @@ public class CarpetSettings
     )
     public static int maxEntityCollisions = 0;
 
+    @Rule(
+            desc = "Customizable server list ping (Multiplayer menu) playerlist sample limit, 0 for no limits",
+            options = {"0", "12", "20", "40"},
+            category = CREATIVE,
+            strict = false,
+            validate = Validator.NONNEGATIVE_NUMBER.class
+    )
+    public static int serverListPingPlayerLimit = 0;
     /*
 
     @Rule(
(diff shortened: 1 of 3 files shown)


Andrews54757:
bump?

gnembon (approved):

PR #718, opened 2021-02-16
Fix #712

### LINKED ISSUE #712: Incompatibility with vanilla server when using structure block
## How to reproduce

1. Join a vanilla server, or any server without fabric carpet, with a client with fabric carpet (>=1.4.25) installed on the client
2. Place a structure block, switch to save mode, edit anything e.g. structure size, press done
3. You gets kicked by the server

image

## Analyze

In commit e32b37dcf1b5021bf4c373f138d6a7a943931fc6 from PR #664, a couble of extra int fields are added to the end of UpdateStructureBlockC2SPacket, which inceases the length of the packet, resulting in vanilla server packet logic raising an exception in `net.minecraft.network.DecoderHandler#decode`

BeetMacol:
Fix https://github.com/gnembon/fabric-carpet/issues/712

### DIFF
--- a/src/main/java/carpet/mixins/UpdateStructureBlockC2SPacketMixin.java
+++ b/src/main/java/carpet/mixins/UpdateStructureBlockC2SPacketMixin.java
@@ -1,6 +1,7 @@
 package carpet.mixins;
 
 import carpet.CarpetSettings;
+import carpet.network.CarpetClient;
 import net.minecraft.network.PacketByteBuf;
 import net.minecraft.network.packet.c2s.play.UpdateStructureBlockC2SPacket;
 import net.minecraft.util.math.BlockPos;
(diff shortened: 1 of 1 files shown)


BeetMacol:
It may still kick the player if the server is `<1.4.25` and the client is on a `>=1.4.25` newer version, but should no longer kick the player if the server is vanilla and the client has carpet.

altrisi:
Please. As I already told you in the main PR. Just check if the rule is different from vanilla on both sides (and maybe if over the limit, to save traffic). Rules are synced with clients, they will know if they should send it.

Fallen-Breath:
This workaround not only creates incompatibility between different carpet version but also still leaves a potential furture incompatibility with other mods / external tools that rely on vanilla packet size, since the packet size is still increased in some situation.

I'll suggest to let the client keep the vanilla packet untouched, then send an extra custom carpet structure block packet to the server with structure block info. Updated carpet server can handle this packet well, and other servers will just ignore this packet

gnembon:
Lol, I have been working on the fix myself. I don't see a big deal in getting into the vanilla packet in this case since . Having extra packet is a 100% proper solution, but introduces much more complexity, which is a bad idea in general (not sustainable long term)

https://github.com/gnembon/fabric-carpet/pull/720

Fallen-Breath:
> Lol, I have been working on the fix myself. I don't see a big deal in getting into the vanilla packet in this case since . Having extra packet is a 100% proper solution, but introduces much more complexity, which is a bad idea in general (not sustainable long term)
> 
> #720

I dont think this "complexity" will cause further maintaining issue. My implementation only needs 1 simple injection, and it's a place where minecraft will not likely to have big changes in the future (structure block screen). For the other not mixin codes, those are also not likely to have needs to change in the future, unless minecraft itself refactors the structure block system. But this is the same as the other extra fields workaround

The framework of carpet protocol is already built, why not take good use of it

altrisi:
I think I'll hold my PR seeing how there are already three for this lol.

gnembon:
closing as a simpler version of #720

PR #719, opened 2021-02-16
Struct block protocol

### LINKED ISSUE #712: Incompatibility with vanilla server when using structure block

Fallen-Breath:
A workaround for issue #712

As I suggested in https://github.com/gnembon/fabric-carpet/pull/718#issuecomment-779800171, using a custom carpet packet instead of extending the vanilla structure block packet provides better compatibility to other mods / tools.

## Changes

- Added a `VANILLA_STRUCTURE_BLOCK_LIMIT` field in `CarpetSettings` for rule modified check
- Added a new Carpet packet id `STRUCTURE_BLOCK` with value `2` for structure block update
- Added sending / handling logic towards the carpet structure block update packet
- Added `StructureBlockScreen_structureBlockLimitMixin`, to let the client send a carpet structure block update packet to the server, only if the server is carpet and the value of the rule `structureBlockLimit` has been modified
- Removed the int field writing in `UpdateStructureBlockC2SPacketMixin`, since its funtionalities have been implemented by the  carpet structure block update packet. The `structureBlockLimitsRead` method is reserved for compatibility with client with fabric carpet 1.4.25 ~ 1.4.26

Fixed #712

## Some tests

Already tested the compatibilities with the following setups:

- new client single player
- new client -> new server
- new client -> 1.4.26 server
- new client -> fabric server without mod
- 1.4.26 client -> new server

Where "new" means a client / server with this PR

## Protentional notvanilla

Since the structure block is written twice when the client updates the structure block (one by vanilla packet and one by carpet packet), there might be some extra `getBlockState` and `updateListeners` that will occur

But these extra operations only occur when the carpet rule is modified by the user, and structure blocks are always used in creative servers, so it's fine I guess

### DIFF
--- a/src/main/java/carpet/mixins/StructureBlockScreen_structureBlockLimitMixin.java
+++ b/src/main/java/carpet/mixins/StructureBlockScreen_structureBlockLimitMixin.java
@@ -16,9 +16,9 @@ import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
 @Mixin(StructureBlockScreen.class)
 public abstract class StructureBlockScreen_structureBlockLimitMixin
 {
-	@Shadow @Final private StructureBlockBlockEntity structureBlock;
+    @Shadow @Final private StructureBlockBlockEntity structureBlock;
 
-	@Inject(method = "method_2516", at = @At(value = "TAIL"), locals = LocalCapture.CAPTURE_FAILHARD)
+    @Inject(method = "method_2516", at = @At(value = "TAIL"), locals = LocalCapture.CAPTURE_FAILHARD)
     private void sendCarpetStructureBlockPacket(StructureBlockBlockEntity.Action action, CallbackInfoReturnable<Boolean> cir, BlockPos blockPos, BlockPos blockPos2)
     {
         if (CarpetSettings.structureBlockLimit != CarpetSettings.VANILLA_STRUCTURE_BLOCK_LIMIT)
--- a/src/main/java/carpet/mixins/StructureBlockScreen_structureBlockLimitMixin.java
+++ b/src/main/java/carpet/mixins/StructureBlockScreen_structureBlockLimitMixin.java
@@ -1,7 +1,6 @@
 package carpet.mixins;
 
 import carpet.CarpetSettings;
-import carpet.network.CarpetClient;
 import carpet.network.ClientNetworkHandler;
 import net.minecraft.block.entity.StructureBlockBlockEntity;
 import net.minecraft.client.gui.screen.ingame.StructureBlockScreen;
@@ -22,7 +21,7 @@ public abstract class StructureBlockScreen_structureBlockLimitMixin
 	@Inject(method = "method_2516", at = @At(value = "TAIL"), locals = LocalCapture.CAPTURE_FAILHARD)
     private void sendCarpetStructureBlockPacket(StructureBlockBlockEntity.Action action, CallbackInfoReturnable<Boolean> cir, BlockPos blockPos, BlockPos blockPos2)
     {
-        if (CarpetClient.isCarpet() && CarpetSettings.structureBlockLimit != CarpetSettings.VANILLA_STRUCTURE_BLOCK_LIMIT)
+        if (CarpetSettings.structureBlockLimit != CarpetSettings.VANILLA_STRUCTURE_BLOCK_LIMIT)
         {
             ClientNetworkHandler.sendCarpetStructureBlockPacket(this.structureBlock.getPos(), blockPos, blockPos2);
         }
--- a/src/main/java/carpet/mixins/StructureBlockScreen_structureBlockLimitMixin.java
+++ b/src/main/java/carpet/mixins/StructureBlockScreen_structureBlockLimitMixin.java
@@ -1,5 +1,6 @@
 package carpet.mixins;
 
+import carpet.CarpetSettings;
 import carpet.network.CarpetClient;
 import carpet.network.ClientNetworkHandler;
 import net.minecraft.block.entity.StructureBlockBlockEntity;
@@ -21,7 +22,7 @@ public abstract class StructureBlockScreen_structureBlockLimitMixin
 	@Inject(method = "method_2516", at = @At(value = "TAIL"), locals = LocalCapture.CAPTURE_FAILHARD)
     private void sendCarpetStructureBlockPacket(StructureBlockBlockEntity.Action action, CallbackInfoReturnable<Boolean> cir, BlockPos blockPos, BlockPos blockPos2)
     {
-        if (CarpetClient.isCarpet())
+        if (CarpetClient.isCarpet() && CarpetSettings.structureBlockLimit != CarpetSettings.VANILLA_STRUCTURE_BLOCK_LIMIT)
         {
             ClientNetworkHandler.sendCarpetStructureBlockPacket(this.structureBlock.getPos(), blockPos, blockPos2);
         }
--- a/src/main/java/carpet/mixins/StructureBlockScreen_structureBlockLimitMixin.java
+++ b/src/main/java/carpet/mixins/StructureBlockScreen_structureBlockLimitMixin.java
@@ -1,0 +1,29 @@
+package carpet.mixins;
+
+import carpet.network.CarpetClient;
+import carpet.network.ClientNetworkHandler;
+import net.minecraft.block.entity.StructureBlockBlockEntity;
+import net.minecraft.client.gui.screen.ingame.StructureBlockScreen;
+import net.minecraft.util.math.BlockPos;
+import org.spongepowered.asm.mixin.Final;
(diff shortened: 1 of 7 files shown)


altrisi:
Well, as I commented in #718, there is not really a need for `CarpetClient.isCarpet()` since the rule will only ever be different than vanilla in the client if the server is Carpet, because the SettingsManager is reset when changing servers, and the rule is synced.

Fallen-Breath:
> CarpetClient.isCarpet()

k, yeeted `CarpetClient.isCarpet()`

src/main/java/carpet/mixins/UpdateStructureBlockC2SPacketMixin.java
```diff
@@ -25,27 +25,11 @@
             at = @At("TAIL")
     )
     private void structureBlockLimitsRead(PacketByteBuf buf, CallbackInfo ci) {
+        // fabric-carpet client 1.4.25 ~ 1.4.26 compatibility
```
gnembon, line 28:
```suggestion
        // fabric-carpet client 1.4.25 ~ 1.4.26 compatibility
        // TODO remove at some point with a major MC release as not important
```

src/main/java/carpet/network/CarpetClient.java
```diff
@@ -16,6 +16,7 @@
     public static final int HI = 69;
     public static final int HELLO = 420;
     public static final int DATA = 1;
+    public static final int STRUCTURE_BLOCK = 2;
```
gnembon, line 19: I am suggesting not creating new types, just hook into DATA - there is nothing special about structure blocks here - data handles all sorts of communication, for rules, scarpet etc.

src/main/resources/carpet.mixins.json
```diff
@@ -211,7 +211,9 @@
     "WorldRenderer_scarpetRenderMixin",
 
     "SoundEngineInt_cleanLogsMixin",
-    "SoundSystem_cleanLogsMixin"
+    "SoundSystem_cleanLogsMixin",
+
+    "StructureBlockScreen_structureBlockLimitMixin"
```
gnembon, line 216: Isn't this one supposed to be only a client side mixin - that class doesn't exist on the server only.

gnembon (changes requested):

src/main/java/carpet/mixins/StructureBlockScreen_structureBlockLimitMixin.java
```diff
@@ -0,0 +1,29 @@
...
+
+@Mixin(StructureBlockScreen.class)
+public abstract class StructureBlockScreen_structureBlockLimitMixin
+{
+    @Shadow @Final private StructureBlockBlockEntity structureBlock;
+
+    @Inject(method = "method_2516", at = @At(value = "TAIL"), locals = LocalCapture.CAPTURE_FAILHARD)
+    private void sendCarpetStructureBlockPacket(StructureBlockBlockEntity.Action action, CallbackInfoReturnable<Boolean> cir, BlockPos blockPos, BlockPos blockPos2)
+    {
+        if (CarpetSettings.structureBlockLimit != CarpetSettings.VANILLA_STRUCTURE_BLOCK_LIMIT)
```
gnembon, line 24: we really need to send this is the limit is > 127

gnembon (changes requested):

gnembon:
I have merged my other PR, #720 in order to fix the issue asap - but I started to like this solution. The fact that we are doubling on packets, its fine - I would only request to stay within DATA when handling these.

Have no concerns about extra world manipulation due to that - its what you mention - its only because of creative nature of structure blocks. I hope only packets come in the same ordered they were sent.

src/main/resources/carpet.mixins.json
Fallen-Breath: It does be a client side mixin. It's in the "client" array

Fallen-Breath:
> I hope only packets come in the same ordered they were sent.

This can be guaranteed by the injection point in `StructureBlockScreen`

https://github.com/gnembon/fabric-carpet/blob/a1384c3e97d38db03e2f35072bfcb697ab08d7a5/src/main/java/carpet/mixins/StructureBlockScreen_structureBlockLimitMixin.java#L21

It injects at tail, so on client side the carpet packet is always sent after the vanilla `UpdateStructureBlockC2SPacket` is sent, so the server will always receive the vanilla packet first

For the server processing part, there's an `NetworkThreadUtils.forceMainThread` check at the begining of every packet in `ServerPlayNetworkHandler` to let `this.player.getServerWorld().getServer()` to enqueue the packet processing execution

Same thing happens in `carpet.network.ServerNetworkHandler#onCarpetStructureBlockPacket`, I just directly invoke `player.getServerWorld().getServer().execute`. So the packet processing order should be what we want.

https://github.com/gnembon/fabric-carpet/blob/7979f17458bfb89a27c6c2494379ff468643f457/src/main/java/carpet/network/ServerNetworkHandler.java#L282

src/main/java/carpet/mixins/StructureBlockScreen_structureBlockLimitMixin.java
Fallen-Breath: In current `UpdateStructureBlockC2SPacketMixin`, there isn't any implementation that stops vanilla `net.minecraft.network.packet.c2s.play.UpdateStructureBlockC2SPacket#read` method clamping the byte value in the packet. 

```java
// in net.minecraft.network.packet.c2s.play.UpdateStructureBlockC2SPacket#read
this.offset = new BlockPos(MathHelper.clamp(buf.readByte(), -48, 48), MathHelper.clamp(buf.readByte(), -48, 48), MathHelper.clamp(buf.readByte(), -48, 48));
this.size = new BlockPos(MathHelper.clamp(buf.readByte(), 0, 48), MathHelper.clamp(buf.readByte(), 0, 48), MathHelper.clamp(buf.readByte(), 0, 48));
```

So If this if judgement is changed to testing limit > 127, then some extra mixins are needed to prevent vanilla server to clamp our offset/size value into vanilla 48 range

Fallen-Breath: Some extra mixins can be some `@ModifyConstant` to change those `-48`, `48` into `CarpetSettings.structureBlockLimit`. An example can be found [here](https://github.com/TISUnion/Carpet-TIS-Addition/blob/1.16.5/src/main/java/carpettisaddition/mixins/rule/structureBlockLimit/UpdateStructureBlockC2SPacketMixin.java#L13-L32)

Fallen-Breath: So should I change the if judgement to > 127 and implement these extra mixins?

Fallen-Breath:
Found an issue about this implementation: Minecraft server executes the operation (save / load structures) of structure blocks during the processing of vanilla structure block packets. At that time the structure block is using the size / offset infomation from the vanilla packet. Even though the information of the struct block will be updated to the expected value later, but the actual save/load operations have already been executed with the incorrect values from the vanilla packets

This will affect all struct block operation by player clicking buttons on the client, but not operations triggered by redstone

To solve this, you need to somehow delay the structure block operation when processing vanilla packet until the carpet packet is processed and updated the structure block entity, e.g. enqueue again everything inside the `if (structureBlockBlockEntity.hasStructureName())` in `net.minecraft.server.network.ServerPlayNetworkHandler#onStructureBlockUpdate`. But, that's really too overcomplex

Fallen-Breath:
idk if there are any workaround for this custom packet based `structureBlockLimit` impl. This PR is ok to be closed if there isn't and the workaround about is too complicated

src/main/java/carpet/mixins/StructureBlockScreen_structureBlockLimitMixin.java
gnembon: true - I will need to change that logic I guess

src/main/resources/carpet.mixins.json
gnembon: lol - totally missed that - yeah, its good.

gnembon:
Yeah, makes sense - it will be a big deal to inject that properly - I can see you. And that's right after I started really liking of sending the extra data via carpet channels.

Regretting this, but its so much simpler to just use the packet itself to handle that.

PR #720, opened 2021-02-16
Fix for 712

gnembon:
simple fix for #712

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -819,7 +819,7 @@ public class CarpetSettings
                     "'structureBlockOutlineDistance' may be required for",
                     "correct rendering of long structures."},
             options = {"48", "96", "192", "256"},
-            category = {CREATIVE, CLIENT},
+            category = CREATIVE,
             validate = StructureBlockLimitValidator.class,
             strict = false
     )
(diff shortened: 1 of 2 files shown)


gnembon:
Allows to extend limits of structure block size to int limits, while taking care of vanilla / carpet compatibility issues.

- it only allows to add extra data if singleplayer, or carpet is on the server
- it only reads extra bytes if there is exactly 24 bytes remaining - it would be ignored in case some other mod adds more fields to it.
 - should not fail in case there is complex mod incompatibility - it will just work until 127.
- Doesn't require a new protocol to handle all the rare cases where this would not work

Fallen-Breath:
Extra fields will be sent to server with carpet version < 1.4.25, won't that be an issue? That's an version compatibility thing between carpets. Same issue will still occurs when u are using latest carpet to connect to servers with old servers

altrisi:
Isn't the whole `CarpetClient.isCarpet()` and `minecraft_server != null` redundant? Since the rule will only ever be different than vanilla in the client if the server is Carpet, because the SettingsManager is reset when changing servers, and the rule is synced, or if running in the integrated server.

Also I'd keep the `>0` since another mod might have also mixed into the packet right after Carpet and waiting for more data, therefore making Carpet fail to read (I would also add a rule check in there btw).

**PS**: This is all, of course, if removing the `CLIENT` category from the rule, which I don't think it applies for it.

gnembon:
@Fallen-Breath connecting with earlier carpet servers will be fine, since they won't be able to change the limit to more than 128, since they don't have such a rule. The only problem would be these few broken versions we got in between, which I am not worried about at all.

@altrisi you are correct, if the size is >=128 (and that's where we only need these extra fields) is where all the other conditions are met. I also removed it as a client setting meaning clients won't be able to adjust them when connecting to vanilla servers.

I would suggest keeping '== 24' for a few reasons. If its less than 24 - this means the packet has been modified not by carpet and its not responsibility of carpet to throw any exceptions - let the vanilla packet ingester deal with the problem. If its more than 24 - in that case its possible that someone else injected more to the packet, than just 6 int's from carpet - in this case - these numbers could be added in unspecified order and we shoudn't not read them anyways. Only ==24 is a valid state that we can assume (also not 100% certain) that these are the 6 missing ints.

PR #741, opened 2021-02-24
Update distance_beta.sc

Ghoulboy78:
k, now only the big one (draw_beta) is left...


src/main/resources/assets/carpet/scripts/distance_beta.sc
```diff
@@ -90,9 +89,12 @@ calculate(end)->(
     )
 );
 
+_carpets() -> system_info('world_carpet_rules'):'carpets'=='true';
+
+
 on_player_places_block(player, item_tuple, hand, block) ->
 (
-    if(block==global_assist_block,
+    if(block==global_assist_block && _carpets(),
```
gnembon, line 97: no need to check for carpets, since block could be different and you need to load the app first so that's all fine.

Ghoulboy78: k

PR #742, opened 2021-02-24
Update draw_beta.sc

### LINKED ISSUE #594: Few issues with Scarpet draw and distance apps

### LINKED ISSUE #637: [Suggestion] Abstract core functionality of draw app into a library

Ghoulboy78:
Fixes #594
Fixes #637 

This is the one which is still WIP, the other ones are ready.


Ghoulboy78:
>Note: This does NOT resolve #637, since you cannot import() things from (at least unloaded) non-library rule apps (since when can you import from non-libraries btw? Just found it). The idea in that issue was to move those basic functions into another (library) app, with them returning a list to be used in the draw app, and any other app that would need them.

>I would suggest to keep that separate from this PR, since this already has many things needing review in it blocking it from being merged.

>BTW, you forgot to remove the debug timers (the times can already be gotten by executing the functions with /script in [app] run ...)

This is altrisi's last comment in #618, just keeping here so I don't forget

Ghoulboy78:
Btw, does it rly need to be global app? I have it currently as todo decide in the script, cos im kinda unsure...

DragonEggBedrockBreaking:
> This is the one which is still WIP, the other ones are ready.

Isn't that what draft Pull Requests are for?

DragonEggBedrockBreaking:
> > Isn't that what draft Pull Requests are for?
> 
> Yeah, but there wasn't rly much to change, plus I kinda got distracted...

Yeah, makes sense.

Ghoulboy78:
IMO it's done now that Ive upgraded the algorithms to be the most efficient I can make them:

Ghoulboy78:
IMO it's done, now that the algorithms are as efficient as I can make them:
sphere: O(n^2) (100% efficient)
ball: O(n^3) (100% efficient)
diamond (filled) : O(n^3) (as efficient as scarpet makes it, which idk how to calculate, but I doubt that it's 100%)
diamond (hollow) : O(n^2) (A little less than 100% efficient, cos I think I iterate over some points twice, but still lightning fast)
pyramid (filled) : O(n^3) (100% efficient)
pyramid (hollow) : O(n^2), cos the base is also O(n^2) and is kinda separate, so for all intents and purposes it's still O(n^2) (100% efficient)
Same dealio with cones, not bothered to type that much.
cuboids and cylinders are both similar: Filled is O(n^3), and hollow is O(n^2) across the board (the deal wit hthe closed ends of these guys is the same as the base of the pyramids and cones)

Ghoulboy78:
I improved circles by using Bresenham circle algorithm, see here: https://www.javatpoint.com/computer-graphics-bresenhams-circle-algorithm
and for filled I just saw here: https://stackoverflow.com/questions/10878209/midpoint-circle-algorithm-for-filled-circles
and just figured out the theory.
It's cool that I could just copy some c# code, and most of it still worked, just needed some syntax changes!

altrisi:
You need to register the library as a `builtInScript` library in `CarpetScriptServer` for it to be loadable from `assets` (like `math`)

Ghoulboy78:
oye im dum thx for reminding mate

altrisi:
Small mention that `return(what)` can be replaced by just `what` at the end of functions of the library, to avoid those return calls. Should not be that bad given that the most expensive part is probably computing the shape itself, but it's a small optimization that I guess may make things slightly faster at least in small/simple shapes.

Also personal opinion (feel free to not agree with me): I think having to pass the actual arguments instead of a list of the args in specific order is simpler and easier to read/understand.

Ghoulboy78:
hmm, yeah, that is vey true, but I opted for this cos im a lazyboi and it let me jsut do `call(fun, argslist)`. Also, the only difference is literally ad [ and ], so its not a big deal. As for the return thing, I kinda forgot abt it.

PR #750, opened 2021-02-28
Added player_list option to display_title()

replaceitem:
Adds changing the server's motd with `display_title(p,'motd',text)` as well as sending a list header packet to players with `display_title(player,'header',header,footer)`.

As i mentioned in the docs, the motd would intefere with the `customMOTD` carpet rule. Maybe that could be removed in favor of the scarpet function? Opinions? Not sure about that.

I wasn't quite sure if this should be part of the `display_title` function, but it seemed to fit the most.


gnembon:
not sure if motd should be modifyable via display_title - its completely different location. Also - 21w08a messed up the packets to setup header and footer, so that functions looks much different.

Good idea to get these part of the API - not sure it would be good to add that to display_title.

replaceitem:
Hmm yeah, doesnt really fit, but i also felt like a seperate function is kind of too much, not sure though...

replaceitem:
Or maybe if system_info was able to modify parts things too (not all of them of course, but some) the motd could be part of that

gnembon:
Yeah, that's what I thought about that. Its not that you can't modify it - there is a carpet rule you can trigger. Controlling the tab-header might also be tricky with carpet loggers, since you need to send both header and footer with a request, so that would cancel what was there before.

replaceitem:
Yeah thats a problem, if someome wants a custom header and a logger, they would probably have to recreate the logger in scarpet. That isnt possible for all loggers i think. Thats what i did as well (modifying header any footer is in my scarpet extension)

Or another solution would be to check in the display_title function if a logger is active, and in that case just ignore the specified footer and replace it with whatever logger the player has active (so footer only works without a logger).

replaceitem:
So, i took a look at the PlayerListHeader packet in 20w11a and its exactly the same as in 1.16.5, not sure what you mean whats different.
Also, about the motd, how should that be implemented now? Its quite tricky and i can't really come up with a place for it that i feel like fits...

gnembon:
I suggest dropping entirely 'motd' - its a server setting, not a display area, and can be tweaked using `/script run run('carpet customMotd')`

Regarding footer/header its all good - it is a display area and regardless if loggers use it, you might want to use an app on it and not use loggers. I suggest adding a tri-argument call for `display_title(players, 'player_list', footer)` with the default empty header

docs/scarpet/api/Auxiliary.md
```diff
@@ -277,9 +277,20 @@ Sends the player (or players if `players` is a list) a title of a specific type,
    Executing with those will set the times to the specified ones.
    Note that `actionbar` type doesn't support changing times (vanilla bug, see [MC-106167](https://bugs.mojang.com/browse/MC-106167)).
 
+### `display_title(players, 'motd', motd),`
+
+Changes the motd (message of the day) of the server. The `player` value is ignored for the `'motd'` text.
+Note that when the `customMOTD` carpet rule is active, this will not work and instead use the MOTD from the rule.
+
+### `display_title(players, 'header', header, footer),`
```
gnembon, line 285:
```suggestion
### `display_title(players, 'player_list', header, footer),`
```

src/main/java/carpet/script/api/Auxiliary.java
```diff
@@ -556,8 +560,46 @@ else if (!interactable && targetBlock == null)
                 case "clear":
                     action = Action.CLEAR;
                     break;
+                case "header":
```
gnembon, line 563:
```suggestion
                case "player_list":
```

src/main/java/carpet/script/api/Auxiliary.java
```diff
@@ -556,8 +560,46 @@ else if (!interactable && targetBlock == null)
...
+                    ((PlayerListHeaderS2CPacketMixin) packet).setFooter(footer);
+                    ((PlayerListHeaderS2CPacketMixin) packet).setHeader(header);
+                    AtomicInteger total = new AtomicInteger(0);
+                    targets.forEach(p -> {
+                        p.networkHandler.sendPacket(packet);
+                        total.getAndIncrement();
+                    });
+                    Value ret = NumericValue.of(total.get());
+                    return (cc, tt) -> ret;
+                case "motd":
```
gnembon, line 589: yeet

src/main/java/carpet/script/api/Auxiliary.java
```diff
@@ -556,8 +560,46 @@ else if (!interactable && targetBlock == null)
...
+                        if (pVal instanceof FormattedTextValue)
+                            title = ((FormattedTextValue) pVal).getText();
+                        else
+                            title = new LiteralText(pVal.getString());
+                    } else throw new InternalExpressionException("Third argument of 'display_title' must be present except for 'clear' type");
+                    server.getServerMetadata().setDescription(title);
+                    return LazyValue.TRUE;
                 default:
-                    throw new InternalExpressionException("'display_title' requires 'title', 'subtitle', 'actionbar' or 'clear' as second argument");
+                    throw new InternalExpressionException("'display_title' requires 'title', 'subtitle', 'actionbar', 'clear', 'header' or 'motd' as second argument");
```
gnembon, line 602:
```suggestion
                    throw new InternalExpressionException("'display_title' requires 'title', 'subtitle', 'actionbar', 'player_list' or 'clear' as second argument");
```

altrisi:
Only problem I see with it being in `display_title` is that it's not as intuitive since that function was named based on the different usages of the `/title` vanilla command.

Also (though this is only personal opinion) I don't really like multi-signature functions, I think they loose some flexibility and make them more difficult to work with (I sometimes ask myself why does one function accepts some params when called with a string and doesn't with another, have to make an extra if just to call the same function, etc). It also doesn't help that it requires many more special handling due to the need of two Texts.

Other than that looks good, I think this was needed since long time ago, was using your extension just for this, will be nice to have it in "vanilla" Scarpet!

replaceitem:
Yeah, i just didn't think that something like a set_player_list() function would fit scarpet, so this was the next best alternative i could find

gnembon:
there are 4 vanilla free real-estate that are accessible from the server (actually 5, but sidebar is fully controlled by scoreboard and I would leave it for scoreboards) and having them in one API function makes sense (regardless of the semantics of the 'title'. Also the specificity of player list (need to provide header and footer) makes is a 'special case' but I prefer it then separating these two places with two different API names.

gnembon (approved):

PR #751, opened 2021-02-28
Fixed slice() returning empty string when referring to last character 

replaceitem:
This fixes an issue with slice() which made it return empty strings when the upper bound was one more than the index of the last character, which should be usable for the upper bound since its non inclusive.

| Code                       | Before | Now      |
|----------------------------|--------|----------|
| `slice('abcd',3,4)`        | `''`   | `'d'`    |
| `slice('abcd',1,4)`        | `''`   | `'bcd'`  |
| `slice(split('abcd'),2,4)` | `[]`   | `[c, d]` |

Also fixed two examples of `slice()` in the docs where the upper bound was included for the example

### DIFF
--- a/docs/scarpet/language/Containers.md
+++ b/docs/scarpet/language/Containers.md
@@ -187,10 +187,10 @@ other parts. In that case consecutive calls to `slice` will refer to index `0` t
 cannot go back nor track where they are in the sequence (see examples).
 
 <pre>
-slice([0,1,2,3,4,5], 1, 3)  => [1, 2, 3]
+slice([0,1,2,3,4,5], 1, 3)  => [1, 2]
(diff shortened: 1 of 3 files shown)


gnembon:
This has a weird behaviour when you use multiples of list length, but the intended behaviour limits it from -size to +size so I guess its fine.

gnembon:
Thanks for spotting that

replaceitem:
> This has a weird behaviour when you use multiples of list length, but the intended behaviour limits it from -size to +size so I guess its fine.

Oh right, i noticed that but wasnt sure how this can be fixed, keeping the previous behaviour and fixing upper index...
I guess bigger values arent used for it normally anyway

PR #753, opened 2021-03-05
Optimizing scoreboard API with scoreboard_property

### LINKED ISSUE #672: scoreboard_add should be really scoreboard_set

### LINKED ISSUE #731: Unable to change display name of scoreboard objectives without using "run('scoreboard ...')" function
Currently, if you want to change the display name of a scoreboard objective, you have to do the following:
`run('scoreboard objectives modify test displayname {"text":"NAME"}');`

Even though Scarpet has its own functions in the `scoreboard` family to handle many other scoreboard use cases, it is this use case that is not handled, which is changing the display name for the function. I suggest adding a Scarpet function for that, named `scoreboard_display_name(objective, display_name)` or something similar, which takes in the objective name and either a `String` or `FormattedText` for the new display name, and changes the display name of the objective. For example, the above could instead be written as follows:
`scoreboard_display_name('test', 'NAME');`.

Another example:
Before: `run('scoreboard objectives modify test displayname {"text":"NAME", "bold":"true"}');`
After: `scoreboard_display_name('test', format('b NAME'));`

replaceitem:
Fixes #731 
Resolves #672 

Adding `scoreboard_property` to change and read various properties of scoreboard objectives:

`scoreboard_property(objective, property)`
`scoreboard_property(objective, property, value)`

* `criterion`
* `display_name` (Formatted text supported)
* `display_slot`: When reading, returns list of slots, when modifying, displays the objective in the specified slot
* `rendertype`: Either `'integer'` or `'hearts'`, defaults to `'integer'` if invalid value specified
* `remove`: Removes the objective, specifying a `value` doesn't matter

This deprecates the part of `scoreboard_add` which allowed changing scoreboard criterions. 

The `scoreboard_remove` function is also deprecated. The two parts of `scoreboard_remove` were:

* Removing an objective. This is now done with `scoreboard_property(objective,'remove')`.
* Resetting a score. This PR also adds resetting a score with `scoreboard(objective, key, null)`.

### DIFF
--- a/docs/scarpet/api/Scoreboard.md
+++ b/docs/scarpet/api/Scoreboard.md
@@ -20,8 +20,6 @@ scoreboard_add('lvl','level')
 
 ### `scoreboard_remove(objective)` `scoreboard_remove(objective, key)`
 
-NOTE: This function is deprecated, use `scoreboard(objective, key, null)` and `scoreboard_property(objective, 'remove')` instead.
-
 Removes an entire objective, or an entry in the scoreboard associated with the key. 
 Returns `true` if objective has existed and has been removed, or previous
 value of the scoreboard if players score is removed. Returns `null` if objective didn't exist, or a key was missing
@@ -40,7 +38,6 @@ Reads a property of an `objective` or sets it to a `value` if specified. Availab
 * `display_name` (Formatted text supported)
 * `display_slot`: When reading, returns list of slots, when modifying, displays the objective in the specified slot
 * `rendertype`: Either `'integer'` or `'hearts'`, defaults to `'integer'` if invalid value specified
-* `remove`: Removes the objective, specifying a `value` doesn't matter
 
 # Team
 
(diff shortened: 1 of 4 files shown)


docs/scarpet/api/Scoreboard.md
```diff
@@ -21,6 +20,8 @@ scoreboard_add('lvl','level')
 
 ### `scoreboard_remove(objective)` `scoreboard_remove(objective, key)`
 
+NOTE: This function is deprecated, use `scoreboard(objective, key, null)` and `scoreboard_property(objective, 'remove')` instead.
```
gnembon, line 23: its far less intuitive to remove the objective by modifying its property, or remove entry by setting it to null.
Suggesting removing _property(..., 'remove') and just keep scoreboard_remove as is.

src/main/java/carpet/script/api/Scoreboards.java
```diff
@@ -150,6 +165,75 @@ public static void apply(Expression expression)
...
+
+                    List<Value> slots = new ArrayList<>();
+                    for(int i = 0; i < 19; i++) {
+                        if (scoreboard.getObjectiveForSlot(i) == objective) {
+                            String slotName = Scoreboard.getDisplaySlotName(i);
+                            slots.add(StringValue.of(slotName));
+                        }
+                    }
+                    return (_c, _t) -> ListValue.wrap(slots);
+                case "rendertype":
```
gnembon, line 221: rendertype is inconsistently not using snake_case like other options. Suggest unifying one way or the other.

docs/scarpet/api/Scoreboard.md
replaceitem: alright makes sense

gnembon (approved):

PR #756, opened 2021-03-08
Exceptions - rework

gnembon:
@altrisi your PR is read-only with github, so I needed to make another one. Feel free to compare. Didn't edit the docs yet.

### DIFF
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -328,8 +328,8 @@ write is beneficial for writing speed. To send multiple packs of data, either pr
 third argument.
 
 Throws:
-- `nbt_read_exception`: When failed to read NBT file
-- `json_read_exception`: When failed to read JSON file. The `_msg` variable will contain details about the problem
+- `nbt_read_error`: When failed to read NBT file.
+- `json_read_error`: When failed to read JSON file. The exception data will contain details about the problem.
 
 <pre>
 write_file('foo', 'shared_text, ['one', 'two']);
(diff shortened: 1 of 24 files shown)


altrisi:
Any specific reason not to allow custom exception types? I think those were nice for things like libraries, for example, so they could have more specific catch filters. (if it's because of complexity I get it)

As a side note it's a bit sad to have lost the way to throw an error without actually doing something incorrect, but I guess that wasn't really the most important feature of the system.

Also `byId` would probably be a better name for `byValue`, since that now stores by id string.

gnembon:
Yeah, allowing for custom exception types could be useful in libraries, or larger apps, even if its only just one level deeper that the existing hierarchy. I will add it back, just need to make sure its not muddling too much the system. You are correct - simplicity is what I would want to keep here.

For throwing an error that will not get caught, that's what exit is for. What do you think about it?

gnembon:
and yeah, to satisfy https://github.com/gnembon/fabric-carpet/issues/446 stack of `throw` should be accessible in catch statements, maybe as `_stack` or `_trace`.

gnembon:
@altrisi I added the full trace information as you requested, including stack, local peekaboo and pointer to the error in code. I will now add back the ability to through direct children of the build-in exceptions in case programmer wants to distinguish several cases, but that would be limited to only direct children of buit-in exceptions.

altrisi:
Oh, that issue (IIRC, it's a bit old) wasn't about getting the stack programmatically, but about unhandled throw exceptions to show it like regular IEEs, it was when throw just worked as a exit when unhandled.

I guess it's a nice addition anyway, but not sure about where you would use it though.

gnembon:
it woudl be for debugging.

PR #772, opened 2021-03-21
Annotated Scarpet

### LINKED ISSUE #723: [Question] (Scarpet) Would a PR doing this (extended) be accepted?
I've kinda had nothing to do for a while, and confused myself with what I needed to have for tests in my (yet useless) ScarpetAppTester. I thought that for some reason I needed to know the whole list of functions in Scarpet (I don't, that program if finished will check user-defined functions, not native ones), and then started thinking at ways of improving the current way of registering functions while also keeping them in a list.

Even though I realized I didn't need that at all, I liked the idea I had of annotating public methods and inferring all arguments and things from there. My current idea is to also allow to declare different types in the different arguments of those methods and make the parser automatically convert the values (and handle the errors when incorrect) to ease coding new functions and deduplicate code (mostly casting and instanceofs, also parameter count checking, value unpacking, etc).

I already have quite some base working, but I'd like to know whether the PR with that would be accepted before coding everything (although I enjoyed doing so and may do it anyway at some point).
Currently, it can already create fixed and variable argument functions passing Values and waiting for LazyValues. But the idea is to be able to pass specific values, unwrapped values (like `Entity`, `String`, `Double`, `Text`, etc), or even extra parsed ones (like `ServerPlayerEntity`), passing the `Context` if needed (currently the petition is registered but it isn't really passed), accepting Values as return and converting them to LazyValues (since that's actually quite common), and maybe passing LazyValues too if requested instead of evaling every parameter pre-execution. The first four already have some kind-of a ~base~ something started.

Preview of the current functionality (code still a bit messy) (after watching it again, not that much useful to watch):

https://streamable.com/kfbwjl

Example code for defining functions with this:

Currently working:
```java
@LazyFunction(maxParams = 6)
public LazyValue semi_multiparams(Value fixed, Value... values) {
	String str = fixed.getString();
	for (Value val : values)
		str += val.getString();
	Value retval = StringValue.of(str);
	return (c,t) -> retval;
}
```

Concept for the final version (kind-of reimplementation of `display_title`):
```java
@LazyFunction(maxParams = 6)
public Boolean display_title(ServerPlayerEntity target, String actionString, Text content, int... times) {
    TitleS2CPacket.Action action;
    switch (actionString.toLowerCase(Locale.ROOT))
    {
        case "title":
            action = Action.TITLE;
            break;
        case "subtitle":
            action = Action.SUBTITLE;
            break;
        case "actionbar":
            action = Action.ACTIONBAR;
            break;
        default:
            throw new InternalExpressionException("'display_title' requires 'title', 'subtitle' or 'actionbar' as second argument");
    }
    if (times.length == 3)
    {
        target.networkHandler.sendPacket(new TitleS2CPacket(Action.TIMES, null, times[0], times[1], times[2]));
    }
    target.networkHandler.sendPacket(new TitleS2CPacket(action, content));
    return true;
}
```
`static` modifier isn't really needed, I may just even remove it for the sake of simplicity by requiring to pass an instance of the class on parse. _Changed_

Now I'm also thinking about allowing Enums in parameters too for predefined options, such as in this case, a custom enum with a variable with the Action itself, with values being the options supported.

altrisi:
# Annotated Scarpet

## Testing in extensions now

You can use Jitpack to test how it would be like in your project. Remember that until released in Carpet it may change, and users won't be able to use it unless they download an artifact of the PR.
Check https://jitpack.io/#altrisi/fabric-carpet/annotated-scarpet-SNAPSHOT for a guide in how to set it up (change `implementation` to `modImplementation`).

Any feedback is appreciated!

## PR body again

This PR is the implementation of a way to register Scarpet functions easily and without lots of duplicated code for things like checking, casting and converting parameters back and forth. That also includes converting return to a `Value` or `LazyValue`.

All you need to do is create a method annotated with the `@ScarpetFunction` annotation with any parameters you need in their Java form, register your function class, and that's it!* Annotated Scarpet will handle converting the arguments to the function and its return value into the proper arguments, you don't even need to know how Scarpet works internally, ideal for other mod developers that just want to integrate a Scarpet function into their API!

Yes, this is an actual Scarpet function!

```java
@ScarpetFunction(maxParams = 10)
public String joinStrings(String... strings)
{
    return String.join(",", strings);
}
```

Includes List and MapValue converting to the specific generics!
```java
@ScarpetFunction
public void extinguish(@Param.AllowSingleton List<ServerPlayerEntity> players)
{
    players.forEach(p -> p.extinguish());
}
```
Did you see that annotation? There are many annotation to modify the behaviour and parsing of arguments, this one allowing to pass a single element of the type to the function instead of a list. Don't want that? Just don't use the annotation, or choose another one!
Also void returns are converted to `Value.NULL` directly!

*Exception is the case you want to use a type which is not registered, in which case you must register it in its proper registry, check javadocs and thrown IllegalArgumentException.

### Slightly more technical details

ALL of this is considered _Public API_. Meaning that once released/merged, public methods should NOT be touched in uncompatible ways (changing method signatures, changing the behaviour of methods, the contracts of interfaces, etc). It's extensively documented to prevent so, the idea is to have a base other devs can rely on being present, and I tried to make it as complete as possible, including registries for the small but useful converters like `SimpleTypeConverter` or `ValueCaster`, and even complex ones like custom `ValueConverter` implementation factories using `@Param.Custom`. Of course it can be extended, but the idea is not to break existing behaviours. Classes are also properly final or package-private to prevent extension and visibility of those that are not meant to be part of the public API. Therefore specifically, absolutely everything visible outside of the `carpet.script.annotation` is meant to be public API, except for `AnnotationParser#apply`, which needs to be public in order to be called from `Expression`, and the `Locator.Locators` class, which is a limitation of me wanting annotation holders to be interfaces for some reason and the visibility of it is anyway limited (and documented saying it's not part of the public API).

Created functions are a kind-of combination of the current ContextFunction, TypedContextFunction and the old LazyFunction: Arguments are evaled and unpacked directly to prevent overcomplicating and potentially making the API slower, working in that regard like a ContextFunction, but you can specify your own `Context.Type` in an annotation parameter, and the API supports return types of LazyValue if desired.

The API will check any passed `Value` and throw an `InternalExpressionException` if they don't match to the required arguments (as of the contract of `ValueConverter` of returning `null` if it is incorrect). It will also build an (IMO nice) error message with the usage report of the function using the name of the declared arguments' types, declared in `ValueConverter#getTypeName`. The error looks like `Incorrect argument passed to display_title2 function. Usage: display_title2(player or list of players, string, optional content, integer...)`

Therefore, in order to declare optional parameters, you have to use a parameter of Java's `Optional` type, which actually makes sense, and allows the converter to follow the contract and also asserts the programmer is checking that the optional argument is actually present before using it.

My example reference usage that I've been using all this time is a port of the `display_title` in `FunctionsTestClass`, which has the exact same behaviour while IMO being easier to read and simpler to maintain.

The Annotated Scarpet system uses the new (-ish, Java 7) blazing fast MethodHandles (in some situations they can even compile to a single `invokeVirtual`), even though they are slightly more complex to work with. Comparing them to the overhead that reflection's ` Method#invoke` has, the difference is substantial, since reflection checks access at invoke while MethodHandles do at create time, plus other things. There are a few more slight details about this in `AnnotationParser`.

Everything else is basically documented all over the code, not gonna repeat something I hope you'll read, reviewing tips below.

### Benefits in Carpet

In my opinion, this PR offers a benefit for Carpet for multiple reasons:
- Will reduce code duplication just checking, casting and converting things, with all it means
  That means centralizing the conversion process of those, allowing to fix conversion bugs codebase-wise in a simple change and potentially optimizing the conversion process for every method without having to change it everywhere or remember what's the best way
- Will make creation of new functions faster and easier, since once a converter is registered you can just use the type as a parameter
- Will prevent common missed checks in Scarpet functions and help debugging Scarpet code
  Why? First one is because of things that are usually skipped: Argument count checking. It's enforced in this system: if your function is effectively varargs it will require you to set the maximum number of params for it.
  Second one is because currently many Scarpet functions don't check extra arguments are of the required type when operation doesn't need them, potentially accepting incorrect arguments and not realizing. That makes debugging code that _sometimes works_ well harder to debug, since it fails when you change an argument that is _not_ the incorrect one
- Allows calling Scarpet functions directly from Java code, with the more useful Java argument and return types, so they can be used for other things or as part of other methods.

### Carpet changes

This PR tries not to touch existing Carpet at all! Of course, some exceptions had to be made, you cannot add something to Carpet without touching it.

Carpet changes in this PR are:
- A hook into `CarpetSettings#onGameStarted` to call a new `parseFunctionClasses` method in `CarpetScriptServer`
- Such `parseFunctionClasses` method in `CarpetScriptServer`, which calls `AnnotationParser` for all current existing existing API classes (this is currently a no-op, since there are no functions declared in those classes)
- A call to `AnnotationParser.apply` in `CarpetExpression`, to apply functions parsed by it to expressions
- This PR also makes `Fluff.AbstractLazyFunction#unpackArgs` call `#unpackLazy`, since that section was duplicated

A port of the `display_title` is available in the `FunctionsTestClass` class, which is currently unreferenced and therefore not added to the language, but it's ready to be moved into `Auxiliary` if wanted (note that it currently changes behaviour to reset on `null or empty string` to only on `null`).

### Systems out of the scope of this PR

Some systems were initially scoped for this PR, however those are no longer in here. Those are Argument Locators.

What are those? Well, basically arguments that called the locators in the `Argument` package. Since the original intention has always been not to touch Carpet and the way I ended up implementing this wasn't directly compatible with current locators, I decided to leave them out.

However, all the wrappers, annotations and parameters for them are already available and ready for whenever they are implemented. I will probably implement some of them in a later PR, already have a working implementation for `BlockArgument` locator.

Major problem was changing locators to use iterators instead of the list directly, given that `ValueConverter`s only get that in order to know what should they be using. It's easy for Block (compat == iterator at offset and keep offset), but more complex for others, since they check the size of the list, which iterators cannot do.

### Remaining Questions

Most important is: What default `SimpleTypeConverter`s should be added?!

I've asked this question several times, but never got an answer! I think it's important to come with a comprehensive set of built-in converters not only to remove the hassle later when developing methods, but also so there are no collisions if one/multiple extensions create their converter then Carpet makes its own (more than one converter for same class = startup crash).

Same for `OutputConverter`s: What Java types are usually converted in functions' return types into `Value`s/`LazyValue`s?

Another question is should I rename `TheLazyT` to `ContextType`, but I guess the answer is yes...

### Reviewing Tips

Welp, this is obviously not the easiest PR to review, but not the hardest! It's mostly comments, which explains quite well what everything is doing. However, in the PR files are not ordered, therefore here are my recommendations to review it if you want it in order of Interface then class (guide supposing you are reading in Github):

Start reviewing at the beginning until you get to `FunctionsTestClass`. Then go to the bottom of the review tab (`ValueConverter`) and review each remaining class going up.

Yeah, that's it really. There are not that many special things.

Last note: Even if everything is rebased on top of the latest commit, this PR is IMO candidate for squashing. 58 commits are many...

<details>
<summary>Old description, with the checkboxes and things</summary>

## Old PR body

Closes #723.

The PR is here to start getting some reviews and opinions if possible, since, even if it's not 100% ready (checklist below), the major part of its functionality is. Only really missing is cancelling functions if parameters are invalid, register a few more built-in converters (_help wanted!_ what should be added?), locators, optional parameters and a few extra checks, but the code is almost ready to accept all of those.

This PR tries not to touch the rest of the Carpet code base at all, the only current changes in this PR are just to initialize the test functions class (to be removed) and an `apply()` function to add the generated functions.

Won't add much more content to the body of this PR, since the code is fully commented (_maaybe even a tad overcommented_).

Summary of working things at the time of first submitting this PR:
- [x] Registering functions from actual Java methods
- [x] Converting the function's output to a respective `LazyValue`
- [x] Converting the Scarpet input to specific values
- [x] Converting Scarpet values in parameters into regular Java types
- [x] Complex converters: Converting lists to their generics
- [x] List converting: Optionally allowing a loose value to make a singleton from it
- [x] Complex converters: Converting maps to their generics
- [x] Getting a `LazyValue` as a parameter
- [x] Variable Scarpet argument parameters and parameters that consume a specific number, but more than one
- [x] Getting the `Context` of the expression in the method
- [x] Strict parameters: Requiring exactly a `StringValue` for a String parameter, `Text` (with shallowness), or boolean (with shallowness too)
- [x] Map converting: Allowing a "key-value" pair mode: from either an actual map, a list type [key, val,key2,val2] or that inline
- [x] ~An heuristic Value auto-registerer~ (this was only done as an entertainment and will be removed)

What's missing in my opinion (comment your ideas!):
- [x] (Optionally) support for primitive in varargs. Currently there's a `ClassCastException` since it needs to be a primitive array, which cannot be cast to `Object[]`
      ~Doing this would require casting varargs from primitives to boxed (already required) to primitives again (declared varargs type) to boxed once again (to cast to `Object[]`) and to primitives one last time (to pass them in the varargs). If someone has a better idea please tell me and I'll implement it, else I'm considering to just throw if varargs is primitive.~ Implemented it differently.
- [x] Checking if lv size is greater than maximum declared params and throwing IEE
- [x] Checks. Anything can currently be null (see below), and if it's null, things may fail.
- [x] Checking whether passed parameters are incorrect and throwing with the required parameter types
       `ValueConverter`s already (most) return `null` if they can't provide a valid value, so it's just a matter to add some `null` checks at the top
- ~Locators~ (**out of scope**)
       The infrastructure is already mostly there, it's just a matter of adding the logic to produce them and change the "`Argument` API" to work with iterators instead of lists (keeping old behaviour for compat). This is the only change made needed to be made to current Carpet code other than calling the `apply` function.
- [x] Optional parameters
       Those are probably just going to be some wrapped ValueConverters that will have a pair of extra checks to allow for the iterator to be empty.. What I'm not sure is: Should them be declared in an annotation and return `null`, or should them use Java's `Optional<Type>`?
- [x] More descriptive names for types in `SimpleTypeConverter`s
- [x] Checking that things are correct (non-null, for the contract of ValueConverter) in list and map converters
- [x] (maybe) try to improve performance a little more. Last 0% scientific benchmark, as mentioned in #723, was of about 7/11 runs per tick with an equivalent to the `display_title` function (still around 10k), although that was when the code was a lot less optimized. _Implemented MethodHandles. Performance is currently even better than current in some methods_
- [x] The Lazy T
- [x] More API for extensions to register converters better than SimpleTypeConverter without resorting to StrictConverter
- [x] Some TODOs I may have missed
- [x] Changing code formatting to Carpet's (`{` in new lines, tabs to spaces,...)
- [x] **NEW**: ~Handle `FunctionUnpackedArgumentsValue` if possible, or change everything to not support LazyValues, which would be sad~ LazyValues are now only supported in return values.
- [x] **EXTRA** (possibility): Allow choosing TheLazyT to use when evaluating arguments (yes, I know it's ContextType, I'll still call it TheLazyT)

I'll try not to force-push this PR often, but I may in order to make sure it works with recent changes to Carpet.

Please feel free to comment anything! I want to make it great, would also love input from Scarpet extension creators! (may ping them at some point)

There are some example functions in `FunctionsTestClass`, although those don't really harness all the power of this. May add more in the future, although those test things will be removed before I mark this PR as ready.

PS: This PR is definitely candidate to squashing. ~This PR is currently on top of 47f9efb0f264c45439aee94b623da99edbc7741e~

And for one last time, a comparison with a function (included in the test class) with an actual existing functions, to see the difference (exactly the same functionality):

<details>
<summary>See the sample</summary>

Yeah, those have been removed from here so my edit window is smaller.

### Annotated Scarpet:

```java
// Check FunctionsTestClass.java
```
### Current Scarpet for comparison:
```java
// Check Carpet's Auxiliary.java
```

</details>

</details>


gnembon:
I assume you understand that this would be optional to use. I would not imagine converting functions like "map" to this scheme, but for API that could be very conventient.

altrisi:
Yes, I do (already mentioned it in the issue). I'm trying to touch the current Carpet infrastructure as less as possible (this is actually built on top of it). 

It's just a way to make it easier to declare simpler functions.

Actually it would even allow someone who doesn't know about the Scarpet internals at all (except for this annotation and parse function ofc) to easily make their own custom APIs.

altrisi:
Just checked it (for the sake of doing it, nothing else) and the conversion would be quite simple, though mostly useless since it wouldn't use almost any features (would just autocast the list to AbstractListValue and convert the result to a lazyvalue), making it almost a copy-paste of the current one with a bit more overhead.

altrisi:
Amazingly, with the change to using MethodHandles, the performance of the `display_title2` function included in the test class is even faster than the current one! From very unscientific testing, around 18k (this) vs 15k (current) runs per tick.

Ghoulboy78:
> * [ ]  Changing code formatting to Carpet's (`{` in new lines, tabs to spaces,...)

Idk how necessary that is tbh... I mean it's not like it's that confusing as long as you can easily see what each part of the code does. If gnembon were that fussed he would never have let my first ever pr to fc go through (just look at the brackets in DrawCommand.java, they're kind of mixed and matched all over the place, cos half the code I wrote and half gnembon wrote)

gnembon:
@Ghoulboy78 maybe I am not annoying enough then. I am all for Jakarta standards and am diligently changing it myself wherever the IDE gives me something else
https://commons.apache.org/proper/commons-net/code-standards.html

Ghoulboy78:
@gnembon and yet you r ok with that kind of code when I write it... I have legit never written my brackets the way you do, and u never seem to give a damn. That style of bracketing screws my brain, not least because normally IDEA formats code like that, but alsocos that's how I normally format.

Ghoulboy78:
Also isn't coding and formatting supposed to be descriptive, (so presenting the options for how to format), as opposed to prescriptive (which is saying that FORMAT YOUR CODE LIKE THIS!!)?
Imo that makes more sense.

altrisi:
I think this should still work after the refactor...

altrisi:
Help how should I handle the new `FunctionUnpackedArgumentValue`?

I have no idea on how to make it work with this system without unpacking the arguments, since it would remove many functionality of this (basically any handling of LazyValues, including using as params or as return value).

altrisi:
So... What built-in SimpleTypeConverters should be added? What are the most used Java types in Scarpet functions (Carpet or extensions) that require a flat 1 Scarpet Value -> 1 Type conversion (supporting also other static variables around)?

Same for OutputConverters: What java types are usually converted in functions' return types into Values/LazyValues?

altrisi:
Should I rename TheLazyT to ContextType now that I know what it is? (and that lazy things are almost not present in the system)

altrisi:
Finally Ready for reviews!

It has actually been for quite a bit of time, just was lazy to reformat it and change the PR body.

All info has been added to the PR body.

gnembon:
Great work - this will be quite handy for the future stuff for sure.

First a few comments about the description and things mentioned there:
 - standard argument locators and output wrappers - yeah - didn't think yet about them, yet another thing that needs to be properly fleshed out.
 - example of display title - something that does change significantly in 1.17, but will check how hard would it be to change it to 1.1 compatible code.
 
 It will probably take a several days to review it, but I am looking into it.
 
 Now with compile time optimizations finally in place, which I always wanted to add, I think it will be less traumatic time for scarpet internals, and good time to set some standards.
 
 Great job.

src/main/java/carpet/script/CarpetScriptServer.java
```diff
@@ -481,4 +495,25 @@ public void reAddCommands()
     {
         modules.values().forEach(host -> host.addAppCommands(s -> {}));
     }
+    
+    public static void parseFunctionClasses()
```
gnembon, line 499: What's your take on extensions and parsing their expression classes.

src/main/java/carpet/script/Fluff.java
```diff
@@ -125,15 +125,7 @@ public boolean numParamsVaries() {
...
-            List<Value> args = new ArrayList<>();
-            for (LazyValue lv : lzargs)
-            {
-                Value arg = lv.evalValue(c, contextType);
-                if (arg instanceof FunctionUnpackedArgumentsValue)
-                    args.addAll(((ListValue) arg).getItems());
-                else
-                    args.add(arg);
-            }
+            List<Value> args = unpackLazy(lzargs, c, contextType);
```
gnembon, line 128: sole reason for me to inline it was to save on an extra function call in that crucial spot that is passed for every system call. Probably with zero impact and I shouldn't be so particular about it. Thoughts?

altrisi: If I've learnt something from being in CaffeineMC's discord for so long, is that if it gets called frequently enough the JVM (or JIT or whatever) will inline it (would have to check it's under the bytecode limit, but a static method invocation doesn't seem that bad anyway IMO).

src/main/java/carpet/script/CarpetScriptServer.java
altrisi: Well (if I understood the question correctly), this one is to parse Carpet and base Scarpet function classes (to have all calls in a Scarpet place and not everything in CarpetServer), while extensions have `AnnotationParser#parseFunctionClass`.

If you mean the `apply(Expression expr)` method currently used, that is not really included in this system, but the old `scarpetApi` method in CarpetExtension is still there for those.

src/main/java/carpet/script/Fluff.java
gnembon: ok then. Didn't know how deep jit gets. Thanks for clarification.

src/main/java/carpet/script/CarpetScriptServer.java
gnembon: well, its more about if that's more 'standard' way of extending scarpet, then having an automated treatment of their API call classes without explicit calls to parseFunctionClass or apply() could be preferrable. But that is outside of the scope of this PR

src/main/java/carpet/script/annotation/AnnotationParser.java
```diff
@@ -0,0 +1,318 @@
...
+            if (isEffectivelyVarArgs)
+            {
+                if (lv.size() < minParams)
+                    throw new InternalExpressionException(name + " expected at least " + minParams + " arguments, got " + lv.size() + ". " + getUsage());
+                if (lv.size() > maxParams)
+                    throw new InternalExpressionException( name + " expected up to " + maxParams + " arguments, got " + lv.size() + ". " + getUsage());
+            } else
+            {
+                if (lv.size() != minParams) // min == max if not varargs. We do this cause we don't use function's arg checking to be able to return lazy
+                    throw new InternalExpressionException("Function " + name + " expected " + minParams + " arguments, got " + lv.size() + ". " + getUsage());
```
gnembon, line 238: 'most' error messages (I like to be 'all') when indicating a faulty method, use 'quotes' to highlight the faulty API call.
Consider `Function incorrectly expected ....` vs. `Function 'incorrectly' expected ... `
```suggestion
                    throw new InternalExpressionException("Function '" + name + "' expected " + minParams + " arguments, got " + lv.size() + ". " + getUsage());
```

src/main/java/carpet/script/annotation/AnnotationParser.java
```diff
@@ -0,0 +1,318 @@
...
+                }
+                params[methodParamCount - 1] = primitiveVarArgs ? ArrayUtils.toPrimitive(varArgs) : varArgs; // Copies the array
+            }
+            return params;
+        }
+
+        private String getUsage()
+        {
+            // Possibility: More descriptive messages using param.getName()? Would need changing gradle setup to keep those
+            StringBuilder builder = new StringBuilder("Usage: ");
```
gnembon, line 300:
```suggestion
            StringBuilder builder = new StringBuilder("Usage: '");
```

src/main/java/carpet/script/annotation/AnnotationParser.java
```diff
@@ -0,0 +1,318 @@
...
+            builder.append(valueConverters.stream().map(ValueConverter::getTypeName).filter(Objects::nonNull).collect(Collectors.joining(", ")));
+            if (varArgsConverter != null)
+            {
+                builder.append(", ");
+                builder.append(varArgsConverter.getTypeName());
+                builder.append("s...)");
+            } else
+            {
+                builder.append(')');
+            }
```
gnembon, line 312:
```suggestion
            }
            builder.append("'");
```

src/main/java/carpet/script/annotation/ListConverter.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+        return list;
+    }
+
+    private List<T> convertSingleton(Value val)
+    {
+        T converted = itemConverter.convert(val);
+        if (converted == null)
+            return null;
+        return Collections.singletonList(converted);
```
gnembon, line 58: Should singletons be returned as normal ArrayLists as well? Aren't they non-mutable? I have caught myself returning .asList() for things like block positions and someone complained you can't alter the size of these. For that reason I replaced all listvalue construtors to use new ArrayList.

src/main/java/carpet/script/annotation/Locator.java
```diff
@@ -0,0 +1,270 @@
...
+                return "block";
+            }
+
+            @Override
+            public R checkAndConvert(Iterator<Value> valueIterator, Context context, Context.Type theLazyT)
+            {
+                BlockArgument locator = null; // BlockArgument.findIn((CarpetContext)context, valueIterator (requires changing to Value), 0,
+                                              // acceptString, optional, anyString);
+                // return (R) (returnBlockValue ? locator.block : locator);
+                throw new NotImplementedException("Locator.Block still requires adapting BlockArgument to accept iterators (which is actually simple)");
```
gnembon, line 172: Oh, so its only a wrapper essentially. I assume you would want to implement that

src/main/java/carpet/script/annotation/Locator.java
```diff
@@ -0,0 +1,270 @@
...
+{
+    /**
+     * <p>Represents that the annotated argument must be gotten by passing the arguments in there into a {@link BlockArgument} locator.</p>
+     * 
+     * <p>Must be used in either {@link BlockValue} or {@link BlockArgument} parameters</p>
+     */
+    @Documented
+    @Retention(RUNTIME)
+    @Target({ PARAMETER, TYPE_USE })
+    @Deprecated // Not implemented, although implementation available
```
gnembon, line 41: Why are you deprecating it? Are you hoping it will be merged with the general block locator using that instead?

src/main/java/carpet/script/annotation/OutputConverter.java
```diff
@@ -0,0 +1,129 @@
...
+ * @see #registerToValue(Class, Function)
+ * @param <T> The type to convert from into a {@link LazyValue}
+ */
+public final class OutputConverter<T>
+{
+    private static final Map<Class<?>, OutputConverter<?>> byResult = new HashMap<>();
+    private static final OutputConverter<Value> VALUE = new OutputConverter<>(v -> (c, t) -> v);
+    static
+    {
+        register(LazyValue.class, Function.identity()); // Primitives are handled. Things are boxed in the process anyway, therefore
```
gnembon, line 41: This functionality could be merged with ValueConversions::guess since they share similar functionality, just the guess requires world handler to convert some runtime types, but a map based converter is better than a series of if-elses.

src/main/java/carpet/script/annotation/ValueCaster.java
```diff
@@ -0,0 +1,101 @@
...
+    static
+    {
+        register(Value.class, "value");
+        register(BlockValue.class, "block");
+        register(EntityValue.class, "entity");
+        register(FormattedTextValue.class, "formatted text");
+        register(FunctionValue.class, "function");
+        register(ListValue.class, "list");
+        register(MapValue.class, "map");
+        register(AbstractListValue.class, "list or similar"); // For LazyListValue basically? Not sure what should use this
```
gnembon, line 42: maps and lists are them also. Its for iterators like block iterators

src/main/java/carpet/script/annotation/OutputConverter.java
altrisi: Hm... Problem is `ValueConversions#guess` uses `instanceof`, which is more flexible than the class directly like here (currently thinking of having `#guess` call `Object#getClass` and search that way in the map, which wouldn't work for interfaces or subclasses).

The reason it works here is because the class used is the declared one (in the method's return type), so even if a subclass or implementation is given it will still search for the former. Can't think of any other way to do it (other than passing the class as another parameter maybe, which I'm not sure if it's worth it).

src/main/java/carpet/script/annotation/ValueCaster.java
altrisi: Hm... Should the built-in list and map converter check for those instead then (if so, what checks would them have, cause it seems it could cause a bit of chaos in some situations)?

Or should I just keep them in here so functions can decide to receive them manually (aka using `AbstractListValue` as a parameter type)?

src/main/java/carpet/script/annotation/AnnotationParser.java
altrisi: Whoops, didn't notice this was a click to fix, implemented it from the IDE (same for the other two, also found a few more).

src/main/java/carpet/script/CarpetScriptServer.java
altrisi: Hm... I guess maybe an entrypoint could be added for Scarpet if that's what you mean? (like the startup entrypoint, but listing Scarpet function classes instead). And possibly have it be an interface with a default no-op `apply` method so extensions can add methods that are not compatible with this system if they wish.

I thought of `parseFunctionClass` as the public API for this because I considered it would be ok as a way to, in the end, communicate what are the classes that have to be checked, and because it's the same as how SettingsManager parses rules. Could definitely have some more abstraction if desired.

altrisi: Though then you still need a way to register your own values or type conversions.

altrisi:
Made a change to `SimpleTypeConverter` to not depend on `ValueCaster`, that saves having to register both a value and a type when the extension has values that just wrap another object that you want (also javadoc'ed its registry).

PR #781, opened 2021-03-30
Make mixins not be required by default

vktec:
This improves compatibility with other mods, as those mods can override carpet's mixins when necessary.

### DIFF
--- a/src/main/resources/carpet.mixins.json
+++ b/src/main/resources/carpet.mixins.json
@@ -216,8 +216,5 @@
     "SoundSystem_cleanLogsMixin"
   ],
   "server": [
-  ],
-  "injectors": {
-    "defaultRequire": 1
-  }
+  ]
 }


vktec:
From Mumfrey, creator of Mixin:
> require is supposed to be a precision tool, you're supposed to set it on configs that are critical or injectors which shouldn't be overridden

Mixin is designed in such a way that conflicting mixins can use priorities to gracefully replace each other. By setting mixins to be required by default, mods prevent this from happening, meaning compatibility has to be explicitly added for every conflicting mod, even if those mods could coexist happily, or could add compatibility themselves.

This mixin config originates in the Fabric example mod. In the same conversation as I quoted from above, Mumfrey also said:
> yeah for the example mod really there should be a required config and an optional one though
> then modders can choose which config to add their mixins to

gnembon:
well, the main problem with optional mixins is that if you care about your mod working fine across mojang updates, you have to set it to require when working on it, since that warns you about potential issues where the code you target against no longer matches. So far we have been targetting these mod conflicts one by one, releasing the mixin require as needed. 

Also if something doesn't get triggered on a random level, then not only stuff gets hidden, but potentially starts doing really bad things. I prefer to get a hard stop on that, rather than having a change of messing something big times.

vktec:
> you have to set it to require when working on it

This is what `expect` is for, which is enabled by default. I accidentally made the entire mixin config non-required, which makes expect not work correctly - have force pushed to fix that.

vktec:
As for the issue of conflicts potentially breaking things on a more subtle level, I would suggest explicitly setting required on mixins where that could be an issue. Many carpet options use only a single mixin, which if disabled would simply disable the rule.

It's also worth noting that any mixin conflicts will be shown in the log as warnings, meaning it's at least quite easy to debug if things do go wrong.

gnembon:
I  honestly prefer to look at each potential conflict and decide what to do with that. In some cases a different mixin can be used to alow both to function, in other cases I just assert it is crucial for carpet etc.

PR #798, opened 2021-04-06
Allow apps to specify mod version requirements

altrisi:
Note: Doesn't have docs since I'm not really sure where to specify this/I was too lazy to check (at least yet, feel free to add them).

Resolves #644.

Allows Scarpet apps to specify dependencies on mods, including specific mod versions (using a `>=` comparison).

Syntax is basically the following:

```py
__config() -> {
    'requires' -> {
        'carpet' -> '1.2.3',
        'minecraft' -> '1.16',
        'java' -> 11,
        'carpet-extra' -> '*',
        ...
    },
    ...
};
```

That means: `Carpet at least 1.2.3, Minecraft at least 1.16, Java at least 11, any Carpet Extra version`.

This basically supports requiring any arbitrary mod to be present, since it uses FLoader's (not FAPI) APIs to get whether the mod is present and what version is it running, plus parsing the requirement as a semantic version and comparing it properly.

Therefore:
- If a specific version is specified, it will require the mod of _at least_ that version to be present
- If an `*` is specified, it will allow _any_ version of the mod as long as it is present
- Currently, there's no other checks. I was going to use version dependency checking logic in FLoader, but it seems that isn't part of the public API, so it may change in a future (unlikely, but not using it just in case)

My opinion to possible Whys:
- Why allowing any mod? App may depend on specific blocks, fluids, entities, functionalities, not only as a complement
- Why not making sure requirement is Carpet when checking dev env? Extensions may use a dev env and require themselves
- Why did you add a `return false` somewhere around the code that isn't related to this? I saw it, had to fix it. Though removing apps in the loading sequence may need to pop the profiler.
- Why not moving Ghoulboy's branch just mentioned in the discussion? That made me remember I also had this locally, which was mostly finished, just stopped cause I wasn't sure if I should have added the operators manually.

Questions:
- Should it be in `readConfig()` instead? In such case, should it throw from there instead of return false in case arguments are correct since else it would read "failed to read config"?
- Should the major or equal version comparison require a prefix in case I/anyone feels like adding more at some point?
- Does the name of the function make sense?

### DIFF
--- a/docs/scarpet/api/Overview.md
+++ b/docs/scarpet/api/Overview.md
@@ -76,7 +76,7 @@ will warn and prevent such apps from loading with an error message. If `allow_co
       'requires' -> _() -> (
           d = convert_date(unix_time());
           if(d:6 == 5 && d:2 == 13, 
-            'Its Friday, 13th' // Will throw this if Friday 13th, will load else since functions return false by default
+            'Its Friday, 13th' // Will throw this if Friday 13th, will load else since functions return `null` by default
           )
     }
     ```
--- a/docs/scarpet/api/Overview.md
+++ b/docs/scarpet/api/Overview.md
@@ -76,7 +76,7 @@ will warn and prevent such apps from loading with an error message. If `allow_co
       'requires' -> _() -> (
           d = convert_date(unix_time());
           if(d:6 == 5 && d:2 == 13, 
-            'Its Friday, 13th' // Will throw this if Friday 13th, will load else since `if` functions returns `null` by default
+            'Its Friday, 13th' // Will throw this if Friday 13th, will load else since `if` function returns `null` by default
           )
     }
     ```
--- a/docs/scarpet/api/Overview.md
+++ b/docs/scarpet/api/Overview.md
@@ -76,7 +76,7 @@ will warn and prevent such apps from loading with an error message. If `allow_co
       'requires' -> _() -> (
           d = convert_date(unix_time());
           if(d:6 == 5 && d:2 == 13, 
-            'Its Friday, 13th' // Will throw this if Friday 13th, will load else since functions return `null` by default
+            'Its Friday, 13th' // Will throw this if Friday 13th, will load else since `if` functions returns `null` by default
           )
     }
     ```
--- a/docs/scarpet/api/Overview.md
+++ b/docs/scarpet/api/Overview.md
@@ -59,11 +59,15 @@ will warn and prevent such apps from loading with an error message. If `allow_co
 *   `'requires'` - defines either a map of mod dependencies in Fabric's mod.json style, or a function to be executed. If it's a map, it will only
     allow the app to load if all of the mods specified in the map meet the version criteria. If it's a function, it will prevent the app from 
     loading if the function does not execute to `false`, displaying whatever is returned to the user.
+    
+    Available prefixes for the version comparison are `>=`, `<=`, `>`, `<`, `~`, `^` and `=` (default if none specified), based in the spec 
+    at [NPM docs about SemVer ranges](https://docs.npmjs.com/cli/v6/using-npm/semver#ranges)
     ```
     __config() -> {
       'requires' -> {
         'carpet' -> '>=1.4.33', // Will require Carpet with a version >= 1.4.32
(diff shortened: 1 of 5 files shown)


gnembon:
Good idea. I mean you can do that yourself when loading the app using system_info in the static sense, but that would not prevent the app to be loaded.
It would be good to support things like https://iscompatible.readthedocs.io/en/latest/

Can actually do that allowing 'requires' to be a function - if present, could be run asserting whether the app should be loaded or fail, technically allowing whatever you want. A false result would allow the app, since this way you can pass the message why not (maybe little opposite of intuition, but good to have a reason to display.
```
__config() -> {
   'requires' -> _() -> (d = convert_date(unix_time());  if(d:6==5 && d:2==13,'Its Friday, 13th' )),
}
```
`system_info('server_mods')` would return the map of loaded mods and its versions, same as mod loader returns.

In terms of questions, 
1) When they fail - not really that important as long as the app is removed.
2) not sure what you mean. `11` would mean essentailly '==11', or something like that, right
3) naming wise its fine by me.

altrisi:
Would it be too much to add a SemanticVersionValue for this in order to allow properly comparing versions?

gnembon:
I would be super careful not to add too much into Value.compareTo nor add to scarpet tools that are a bit excessive. Therefore I like the idea of a few simple version checks you can do with the stuff you proposed, defaulting into a function in case you need something more complex. In short - not feeling that is something that would be that useful.

altrisi:
Note: This is not 100% ready yet, currently this defaults to `>=` comparison and that's the only check it can do. I'll add a few prefixes to a map or something later, since Fabric's SemanticVersionPredicateParser isn't public API (though if you prefer I could use it, since it's unlikely to be changed).

altrisi:
Added a few prefixes (committing in a sec), though it lacks what I consider the most important functionality of Fabric's: Wildcards. But it's the best we can currently do, since there is no public API constructor for semantic versions with wildcards.

Once that commit is pushed, this is ready for review again. I'll add docs soon enough.

docs/scarpet/api/Overview.md
```diff
@@ -56,6 +56,30 @@ conflicts and ambiguities between different paths of execution. While ambiguous
...
+        'chat-up' -> '*' // Will require any version of the chat-up mod
+      }
+    }
+    ```
+    ```
+    __config() -> {
+      'requires' -> _() -> (
+          d = convert_date(unix_time());
+          if(d:6 == 5 && d:2 == 13, 
+            'Its Friday, 13th' // Will throw this if Friday 13th, will load else since functions return false by default
```
gnembon, line 79:
```suggestion
            'Its Friday, 13th' // Will throw this if Friday 13th, will load else since functions return `null` by default
```

docs/scarpet/api/Overview.md
```diff
@@ -56,6 +56,30 @@ conflicts and ambiguities between different paths of execution. While ambiguous
...
+        'chat-up' -> '*' // Will require any version of the chat-up mod
+      }
+    }
+    ```
+    ```
+    __config() -> {
+      'requires' -> _() -> (
+          d = convert_date(unix_time());
+          if(d:6 == 5 && d:2 == 13, 
+            'Its Friday, 13th' // Will throw this if Friday 13th, will load else since functions return `null` by default
```
gnembon, line 79:
```suggestion
            'Its Friday, 13th' // Will throw this if Friday 13th, will load else since `if` functions returns `null` by default
```

docs/scarpet/api/Overview.md
```diff
@@ -56,6 +56,30 @@ conflicts and ambiguities between different paths of execution. While ambiguous
...
+        'chat-up' -> '*' // Will require any version of the chat-up mod
+      }
+    }
+    ```
+    ```
+    __config() -> {
+      'requires' -> _() -> (
+          d = convert_date(unix_time());
+          if(d:6 == 5 && d:2 == 13, 
+            'Its Friday, 13th' // Will throw this if Friday 13th, will load else since `if` functions returns `null` by default
```
gnembon, line 79:
```suggestion
            'Its Friday, 13th' // Will throw this if Friday 13th, will load else since `if` function returns `null` by default
```

PR #799, opened 2021-04-06
Remove limitation of rule apps of having the minimum permission level set to the rule's value

altrisi:
Since apps can now provide their own command usage validator, this PR removes the limitation that rule apps have to only be able to be ran with the permissions specified in their rule's field.

Why? For instance, an app may only want to remove commands based on different criteria, plus currently if the app isn't set to a valid permission level (e.g. the app isn't only a command but has some extra features configured in the rule) it will just return that it can't ever be executed, when the app probably could decide whether it can actually be executed (`SettingsManager#canUseCommand` returns `false` by default if object doesn't meet any criteria, which makes sense, but doesn't work as expected for this). Also it's spread through half of the code, though that part may be wanted to be kept in order to allow extra command validators from code other than this, in which case this PR should be closed and the field restriction just be killed directly.

BTW I'd personally use a predicate there, though that's more a matter of taste.

### DIFF
--- a/src/main/java/carpet/commands/ScriptCommand.java
+++ b/src/main/java/carpet/commands/ScriptCommand.java
@@ -246,13 +246,13 @@ public class ScriptCommand
                         suggests( (cc, bb) -> suggestMatching(CarpetServer.scriptServer.listAvailableModules(true),bb)).
                         executes((cc) ->
                         {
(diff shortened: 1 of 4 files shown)


altrisi:
Conflicts! Yay!

Will reopen when I fix them, if I don't forget.

altrisi:
Done

PR #804, opened 2021-04-10
Chain the last EE constructor

altrisi:
Does that by making the `lazyMessage` supplier already present in EE be passed directly in the last constructor.

Follows #801, wanted to make it a single PR but I didn't have time to.


altrisi:
Thinking about this again, it would partially kill my use case, I'll just mix into both. It even makes sense to have two separate constructors for each exception, that way I can get the message without the stacktrace.

PR #819, opened 2021-04-18
Make `player_list` take arguments one at a time and make it compatible with loggers

### LINKED ISSUE #818: Suggestion for `display_title` arguments before release
I got late to the PR... Hopefully I don't for the release.

My idea is to have two different arguments: `tab_header` and `tab_footer`. Those will set a variable somewhere in Carpet that will be used whenever sending the tab list packet. Loggers will be added with these variables, and as such you can have some custom headers or footers while having some footer loggers.

It would also make `display_title` always follow the same format, allowing for changing the position of the display by just a variable instead of an if changing the entire signature. It would also be able to reuse most code that gets the Text instead of that much special handling.

Update: PR now attached. The final behaviour is explained there.

altrisi:
Implements (and GH keyword) Closes #818.

This PR changes the `display_title()` signature for things in the `player_list` to `display_title(targets, position, text)`, making it the same as the other `display_title()` arguments.

Not only that, it also changes the internals so that those headers and footers are stored in `HUDController` instead of just being sent once. HUDController will then make sure that those values are sent to the client even if it's sending other loggers at the same time.

Current behaviour is to display the Scarpet footer above loggers, can easily be changed by moving the call it has in `HUDController#updateHud`.

`CarpetServer` will reset the Scarpet headers and footers when the world changes, but they will be stored for the player while the server is running (if SPE keeps same hashcode and equals after relog will also keep them after relog, else it won't).

Also adds a `force` boolean parameter to `HUDController#updateHud` so Scarpet headers and footers can be updated instantly when the function is called.

Old versions are not deprecated since they have not been released yet and my idea would be to get them before release.

### DIFF
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -265,7 +265,7 @@ Example usages:
   // not a problem in apps
 </pre>
 
-### `display_title(players, type, title?, fadeInTicks?, stayTicks?, fadeOutTicks),`
+### `display_title(players, type, text?, fadeInTicks?, stayTicks?, fadeOutTicks),`
 
 Sends the player (or players if `players` is a list) a title of a specific type, with optionally some times.
  * `players` is either an online player or a list of players. When sending a single player, it will throw if the player is invalid or offline.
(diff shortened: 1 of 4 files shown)


gnembon:
What about that @replaceitem, since that was your PR earlier?

One thing that makes me cringe is the fact that we are mixing here carpet and scarpet, introducing something that cannot be kept if these were separate entities. 

On the upside, it makes more sense. as it behaves like other display text locations, so I thing it would be good to do it this way.

replaceitem:
It makes sense to do it, as long as scarpet is part of carpet, yeah.

gnembon:
could always change that if that's ever not the case. Lets run with it then.

gnembon (approved):

PR #820, opened 2021-04-18
Make `get()` always return a list if requested item ends with `[]`

### LINKED ISSUE #746: Scarpet: get(block, "Items[]") sometime is a list, sometime is a scalar
## The problem

Look at this snippet:

```it=get(bd=block_data(b), 'Items[]'))```

Dependent upon how many items are in the block tile, the following code will return either 
1. `null`,
2. `[{item1...},{item2...}]`
3. `{item1}`

This is annoying, because the result cannot be passed to `map()`, `filter()` and so on directly, without doing some extra fluff code to check type and retype. 

The `null` is reasonable, however, the case 3: `{item1}` is questionable. 

### Make `get(b, 'Items[]')` always return a list, even if it's a 1-item list. 

This is in my opinion the correct behavior and the correct bugfix.

I have been told that this behavior of demoting 1-item lists to scalars is inherited from Vanilla `data` queries. Still this appears inconsistent/puzzling/annoying. 

---

### Make `map` and similar functions accept scalars as input

```map({item1}, ...)``` would then be equivalent to ```map([{item1}], ...)```. There is some concerns regarding loosening the typing of the language. I don't think type checking is critical in Scarpet, but one may think otherwise and make a convincing argument. 

---

### Do nothing 

With no changes to Scarpet, the case can be handled with the following snippet:

```if(null != (it=get(bd=block_data(b), 'Items[]')), if(!has(it, 0), it=l(it)); map(it, ...)```
 (I have been told that `type()` could also work here). 

It is not nice or logical, but it works just fine. 

----

## Duplicate issue

I have been told that this is a preferable place to file this issue, compared to the scarpet repo. Feel free to close whichever is a duplicate w.r.t. https://github.com/gnembon/scarpet/issues/185

### PR DESCRIPTION
**Pull request by altrisi**
Resolves #746.

Basically that. If the requested address ends with `[]` it will create either a singleton with the element.

PENDING TESTING

Can this break apps relying on the old functionality?
I guess maybe? But most should work since the check that seems to be done usually is to convert it to a list if it's not, and this will just return lists in more cases. And it still returns null for no items.

### DIFF
--- a/src/main/java/carpet/script/value/NBTSerializableValue.java
+++ b/src/main/java/carpet/script/value/NBTSerializableValue.java
@@ -569,13 +569,14 @@ public class NBTSerializableValue extends Value implements ContainerValueInterfa
     @Override
     public Value get(Value value)
     {
-        NbtPathArgumentType.NbtPath path = cachePath(value.getString());
+        String valString = value.getString();
+        NbtPathArgumentType.NbtPath path = cachePath(valString);
         try
         {
             List<Tag> tags = path.get(getTag());
             if (tags.size()==0)
                 return Value.NULL;
-            if (tags.size()==1)
+            if (tags.size()==1 && !valString.endsWith("[]"))
                 return NBTSerializableValue.decodeTag(tags.get(0));
             return ListValue.wrap(tags.stream().map(NBTSerializableValue::decodeTag).collect(Collectors.toList()));
         }

### CODE REVIEW

File: src/main/java/carpet/script/language/DataStructures.java [hunk 1]
```diff
@@ -238,9 +238,12 @@ else if (lv.size() == 2)
             for (int i = 1; i < lv.size(); i++)
             {
                 if (!(container instanceof ContainerValueInterface)) return (cc, tt) -> Value.NULL;
-                container = ((ContainerValueInterface) container).get(lv.get(i).evalValue(c));
```
gnembon, line 241: Please don't fix in here, but in NbtValue where it implements .get - otherwise this code will be executing every time you need for example to get the value of a list like `lst:1`. This issue affects ONLY nbts

gnembon: Also this will then also affect all other usages of CVI.get in all other cases. Check NBTSerializableValue#580

altrisi:
Will change that then reopen. I'm not very familiar with that part of the code, sorry.

altrisi:
Changed. Pending testing though.

gnembon:
Should work just fine, what can go wrong?

PR #823, opened 2021-04-19
Adding getTextByValue method to replace duplicate code

replaceitem:
I noticed there are a lot of instances where there is a `value instanceof FormattedTextValue` check, so this static method in the `FormattedTextValue` class replaces all of those.

In the EntityValue modifying customName code, i also swapped the LiteralText constructor with this, which makes it possible to use formatted text here too, useful side effect.

### DIFF
--- a/src/main/java/carpet/script/value/EntityValue.java
+++ b/src/main/java/carpet/script/value/EntityValue.java
@@ -1106,7 +1106,7 @@ public class EntityValue extends Value
                 v = ((ListValue) v).getItems().get(0);
             }
             e.setCustomNameVisible(showName);
-            e.setCustomName(new LiteralText(v.getString()));
+            e.setCustomName(FormattedTextValue.getTextByValue(v));
         });
 
         put("persistence", (e, v) ->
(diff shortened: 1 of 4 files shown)


src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -1106,7 +1106,7 @@ private static void updateVelocity(Entity e)
                 v = ((ListValue) v).getItems().get(0);
             }
             e.setCustomNameVisible(showName);
-            e.setCustomName(new LiteralText(v.getString()));
+            e.setCustomName(FormattedTextValue.getTextByValue(v));
```
replaceitem, line 1109: Adds functionality to use formatted text for modifying entity names

altrisi:
Maybe change the one in display title too and change `soundsTrue` to something like
```java
soundsTrue = pVal instanceof FormattedTextValue ? title.asString().isEmpty() : pVal.getBoolean()
```

replaceitem:
Yep makes sense will add that too

gnembon:
best would be to properly define what a getBoolean() is for a Text, that is not just !asString().isEmpty() - which would be quite inefficient. Current definition of getBoolean is incorrect cause  `bool(format('w ')) == true` right now

gnembon (approved):

PR #824, opened 2021-04-20
Adding query player language

### LINKED ISSUE #787: [Scarpet Suggestion] Querying player language
It would be nice, and it is sent in `ClientSettingsC2SPacket`. It is not really saved anywhere, but the packet is passed to the `ServerPlayerEntity` in `setClientSettings`.

I'm a total noob in Mixin, so I'll probably not try to implement this at least Soon™.

Ghoulboy78:
Fixes #787 
- [x] Todo test

### DIFF
--- a/src/main/java/carpet/mixins/ServerPlayerEntity_scarpetEventMixin.java
+++ b/src/main/java/carpet/mixins/ServerPlayerEntity_scarpetEventMixin.java
@@ -46,8 +46,6 @@ public abstract class ServerPlayerEntity_scarpetEventMixin extends PlayerEntity 
 
     @Shadow public boolean notInAnyWorld;
 
-    @Shadow public int pingMilliseconds;
-
     @Redirect(method = "consumeItem", at = @At(
             value = "INVOKE",
             target = "Lnet/minecraft/entity/player/PlayerEntity;consumeItem()V"
@@ -132,7 +130,7 @@ public abstract class ServerPlayerEntity_scarpetEventMixin extends PlayerEntity 
     }
 
     //getting player language
-    private String language = null;
+    private String language;
 
     public String getLanguage(){
         return this.language;
--- a/src/main/java/carpet/mixins/ServerPlayerEntity_scarpetEventMixin.java
+++ b/src/main/java/carpet/mixins/ServerPlayerEntity_scarpetEventMixin.java
@@ -1,5 +1,6 @@
 package carpet.mixins;
 
+import carpet.fakes.ClientSettingsC2SPacketInterface;
 import carpet.fakes.EntityInterface;
 import carpet.fakes.ServerPlayerEntityInterface;
 import carpet.script.EntityEventsGroup;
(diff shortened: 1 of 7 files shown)


Ghoulboy78:
Ok, this works well. It's a bit of a mixing horror show, because not only do I have to mix into the `setClientSettings` method in `ServerPlayerEntity` as mentioned [here](https://github.com/gnembon/fabric-carpet/issues/787#issue-848408862) by altrisi, I *also* have to mix into `ClientSettingsC2SPacket` as the language is private. Not sure if that means it should stay that way or what tho tbh, cos on the one hand gnembon mentioned here : 
> some players may want to remain anonymous. That's why there is 'ping' for players, but not 'ip'. Language stays in the gray zone between these two.
I understand this and respect that viewpoint, but then gnembon said this:
> I am on the fence of the language thingy, but on the 'its fine' side.
So I think we should have a poll. 👍 if you think we should keep this and allow player's language to be seen by server, 👎 if you think not.

DragonEggBedrockBreaking:
I agreed, because language reveals little about someone, e.g. if it is a common language (e..g spanish), even the country is ambiguous (spain, argentina, chile, etc.), and if it is specific to s relaitevely small country (e.g. bulgaria), it could still be someone just practising the language or smth. You have no way to know their actual location through their language, only what language they know. The only potential issue I can think of is someone being racist to people who speak a certain language, but that is a separate issue, and there are other equally bad things that can be done without knowing the language, so this point is anyway quite invalid.

gnembon:
lol even vanilla doesn't really use it for anything

Ghoulboy78:
I have made the poll in carpet mod channel in Sci discord, so anyone there can vote. After a week, I will either close or leave this pr open, and let gnembon know that this pr is ok to accept or something, idrk. People may want to maintain privacy, in which case they should be allowed to decide. https://discord.com/channels/211786369951989762/573613501164159016/834157181229203516

gnembon:
community hath spoken, go for it.

src/main/java/carpet/mixins/ServerPlayerEntity_scarpetEventMixin.java
```diff
@@ -44,6 +46,8 @@ public ServerPlayerEntity_scarpetEventMixin(World world, BlockPos blockPos, floa
 
     @Shadow public boolean notInAnyWorld;
 
+    @Shadow public int pingMilliseconds;
```
gnembon, line 49: what do you need it for?

src/main/java/carpet/mixins/ServerPlayerEntity_scarpetEventMixin.java
```diff
@@ -126,4 +130,16 @@ public boolean isInvalidEntityObject()
     {
         return isInvalidReference;
     }
+
+    //getting player language
+    private String language = null;
```
gnembon, line 135: you don't need to == null here (its gonna be added to all constructors if you do it this way)

src/main/java/carpet/mixins/ServerPlayerEntity_scarpetEventMixin.java
Ghoulboy78: Umm, Im not sure how it got there...

src/main/java/carpet/mixins/ServerPlayerEntity_scarpetEventMixin.java
Ghoulboy78: ye I noticed, just wasn't 100% sure when that method got called cos as I mentioned I just coded this up in school without being able to test, so I didn't want NPE

Ghoulboy78:
ok then this is complete now @gnembon

gnembon (approved):

PR #826, opened 2021-04-20
Adding /script download

### LINKED ISSUE #713: Scarpet Request | /script get
## The Idea
I had an idea for a way for Carpet users to quickly and easily download and install scripts to their world.  An new choice for the `/script` command: `get`.

## The Implementation
`/script get` requires two arguments, a category and a name.  The category refers to the folders in the Scarpet Github repository and the name is the name of the desired script.  If the specified script exists, it will automatically be downloaded For example: the command `/script get survival angel_block.sc`  will download the `angel_block` script from the Scarpet repo and put it in the `scripts` folder for the current world, creating the folder if it doesn't already exist.  The command would only work for scripts in the repo to limit the chances for malicious use.

## The Challenges
The main sticking point with this suggestion would be the need to add some sort of web handler to access the Scarpet repo.
Additionally, I'm not sure if auto-complete would be possible without repeated queries to the repo.

### Special Thanks
Special thanks to Kollum in the Scicraft Discord for helping refine the idea when explaining it to them.

Ghoulboy78:
Very much a WIP thing, but a fun project nontheless
Resolves #713 

Tasks: 
- [x] Getting the actual code
- [x] Saving it to file properly
- [x] Proper suggestion/autocompletion
- [x] Testing saving to global folder on servers (Resolution: It doesn't work)
- [x] Adding detailed docs

### DIFF
--- a/docs/scarpet/api/ScriptCommand.md
+++ b/docs/scarpet/api/ScriptCommand.md
@@ -167,3 +167,10 @@ This would slow down the computation of fib(40) from a minute to two, but allows
 and be responsive to commands, using about half of each tick to advance the computation. Obviously depending on the 
 problem, and available hardware, certain things can take more or less time to execute, so portioning of work with 
 calling `gametick` should be balanced in each case separately
+
+# `/script download` command
+
+`/script download <global or local?> <path>` command allows to download and save scripts directly from [scarpet app store](https://www.github.com/gnembon/scarpet).
+You can choose whether to save the script in the global config scripts folder or the local per-world script folder, (todo check for server saving scenario...)
+and it will be saved there, e.g `/script download local survival/angel_block.sc` would save the `angel_block.sc` file in https://github.com/gnembon/scarpet/blob/master/programs/survival/angel_block.sc
+to a folder called `survival` in the world's scripts folder.
--- a/docs/scarpet/api/ScriptCommand.md
+++ b/docs/scarpet/api/ScriptCommand.md
@@ -171,6 +171,7 @@ calling `gametick` should be balanced in each case separately
 # `/script download` command
 
 `/script download <global or local?> <path>` command allows to download and save scripts directly from [scarpet app store](https://www.github.com/gnembon/scarpet).
-You can choose whether to save the script in the global config scripts folder or the local per-world script folder, (todo check for server saving scenario...)
-and it will be saved there, e.g `/script download local survival/angel_block.sc` would save the `angel_block.sc` file in https://github.com/gnembon/scarpet/blob/master/programs/survival/angel_block.sc
-to a folder called `survival` in the world's scripts folder.
+You can choose whether to save the script in the global config scripts folder or the local per-world script folder, 
+(on servers it always saves in the scripts folder) and it will be saved there, e.g `/script download local survival/angel_block.sc`
+would save the `angel_block.sc` file [here](https://github.com/gnembon/scarpet/blob/master/programs/survival/angel_block.sc)
+to a folder called `appstore/survival` in the world's scripts folder.
--- a/src/main/java/carpet/commands/ScriptCommand.java
+++ b/src/main/java/carpet/commands/ScriptCommand.java
@@ -304,14 +304,13 @@ public class ScriptCommand
                                                 StringArgumentType.getString(cc, "call")
                                         )?1:0))));
 
-        LiteralArgumentBuilder<ServerCommandSource> d = literal("download").then(literal("global").then(argument("path", StringArgumentType.greedyString()).
+        LiteralArgumentBuilder<ServerCommandSource> d = literal("download").requires((player) -> SettingsManager.canUseCommand(player, CarpetSettings.commandScriptACE)).then(literal("global").then(argument("path", StringArgumentType.greedyString()).
                 suggests((cc, bb) -> null).//todo (in this pr) suggestion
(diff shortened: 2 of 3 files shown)


gnembon:
now tab completion ....

Ghoulboy78:
Well first of all there is issue of actually getting it to get the correct code...

Ghoulboy78:
That was just cos I wanted to quickly switch branches and fix the other ones... I don't like keeping stashed changes in case smth happens and I lose them.

Ghoulboy78:
I also noticed this while researching: https://github.com/hub4j/github-api

Ghoulboy78:
ok wtf? This is ridiculous. My IDE shows no errors, but this stupid thing does?

Ghoulboy78:
oh wait, NOW it shows errors. Stupid piece of crap wouldn't respond for about a minute cos it was indexing.

Ghoulboy78:
yo gnembon what's the deal with global carpet config scripts folder? like where does that get saved?

Ghoulboy78:
Right now if it finds a file already in that location, it will throw an error and not do anything. But I thought about it and there is an interesting thought. Currently, survival/angel_block.sc would be put into a folder called survival, and in that would be created angel_block.sc. So it is unlikely that people who don't know what they're doing have put it there. Not to mention, that perhaps the file could have been updated since you last downloaded, in which case you could update? Idk it's just of commenting out a couple lines, but it changes quite a lot.

gnembon:
you can put it even deeper, like /survival/angel_block would go into /config/carpet/scripts/appstore/surivival/angel_block.sc etc
And don't worry about updating, worry more about autocomplete

Ghoulboy78:
ok, was just wondering cos it's like two lines of code, but potentially changing quite a bit, whereas the autocomplete is more of a coding challenge than a design challenge. The above issue is, as mentioned, like 2 lines literally.

Ghoulboy78:
Admittedly my solution is kinda ugly, cos it basically reads the html response doc, and checks to see the tag which is used to store the file names in one long-ass regex Pattern object. It is vulnerable to breakage if github ever decides to redesign its web browser, so perhaps a better solution is needed, but I couldn't find it after quite a while of searching

altrisi:
You could use Github API for suggestions, for example the `/repos/{owner}/{repo}/contents/{path}, as described in 
https://docs.github.com/en/rest/reference/repos#contents

Ghoulboy78:
yeah, but that's all for javascript, not java. so idk, this seems to be the only option

altrisi:
Ehhm... No?

Github API is http, it's just they also have a wrapper in JS for convenience.

Ghoulboy78:
ah, so is there a way for me to manipulate in java?

altrisi:
Yes, just call it like any other GET (web) request (I added an example URL to the comment above before seeing your last message)

Ghoulboy78:
Ah thanks

Ghoulboy78:
But wait @altrisi how do I read that?

altrisi:
It's JSON, just pass the string to GSON.

Ghoulboy78:
yeah I noticed, I was jsut having trouble cos I was an idiot and I tried to parse as JsonObject not JsonArray first

altrisi:
I'd cache the suggestions in order not to reach ratelimiting (cache directories, then files for each directory when requested, not refreshed until at least 5-10 hours). Else it may cause increased network traffic and could stop working/ban the server's IP if someone abuses the suggestion provider, which IIRC would send the requests each time a single character changes.

Ghoulboy78:
hmm, ye, that sounds like a smart idea.

altrisi:
Maybe through guava's CacheBuilder, since it's already included in minecraft.

Ghoulboy78:
hmm, I was thinking of a different solution which is basically to store the repo's files and stuff statically, so upon server creation (and restarts ofc), and then if the query throws an error it means that the repo's structure changed, so we just re-query and then resend. to the user the updated options. No caching, no nothing needed.

Ghoulboy78:
yo @gnembon should I add an option to query from any repo or just scarpet one? cos the code is rather generalized so I think it may technically work on any repo, to in practice it's only been tested on scarpet for the time being

Ghoulboy78:
Here is a video showcasing the suggestion thing working:https://www.youtube.com/watch?v=tQ3sAj6hYDI&list=PLAMmEWZKDdq4Xtl2aMyRfeVoZ5Zw7mV-6

Ghoulboy78:
Ok, I've gotten the script downloading and saving to work, now I've got to test saving to the global scripts folder on a server, and it should be all good to merge. I'm not sure about whether the current docs are enough, perhaps you wanna add more?

gnembon:
For what its setup to, works great. Its useless pretty much with local world location (primary use I guess)
Capture

Its a minor thing though, so if you don't see that in a few hours, I will just merge it and fix it myself.

gnembon:
Capture
Also suggestions get pretty wonky. No pressure though - will fix it myself if you don't get that to work today.

Ghoulboy78:
I do not see the issue with the global scripts folder. Weird. I did, however, fix the second issue, by simply not allowing a folder name to match the regex pattern : `\.+\.sc\`. Kind of a hack fix, I know, but also the only way I know to fix it tbh without getting too complicated. Also, I ensured that those two are the only types of names you could get, so it works ig.

PR #829, opened 2021-04-23
Adding b64 and json decoder and encoder functions

### LINKED ISSUE #811: scarpet should have native support for json and b64 de/serializers
maybe rename for consistency?
parse_nbt -> decode_nbt

then 

decode_b64(string) -> string
encode_b64(string) -> string
decode_json(string) -> value
encode_json(value) -> string

Ghoulboy78:
Fixes #811 
Todo:

- [x] Testing properly i.e actually checking if this works. I has no reason not to as far as I can see, but I haven't run it yet
- [x] Docs


altrisi:
Instead of a new function in ValueConversions you could just use `ScarpetJsonDeserializer`, or reuse the existing GSON instance which has JSON -> Value already prepared (I think it was somewhere around FileModule, but that has changed recently.

Ghoulboy78:
That sounds like a very good idea, Imma check it out.

src/main/java/carpet/script/language/ControlFlow.java
```diff
@@ -145,5 +151,23 @@ public static void apply(Expression expression) // public just to get the javado
                 return (c_, t_) -> retval;
             }
         });
+
+        expression.addUnaryFunction("encode_b64", v -> StringValue.of(Base64.getEncoder().encodeToString(v.getString().getBytes())));
+        expression.addUnaryFunction("decode_b64", v -> {
+            try {
+                return StringValue.of(new String(Base64.getDecoder().decode(v.getString()), StandardCharsets.ISO_8859_1));//using this charset cos it's the one used in decoding function
+            } catch (IllegalArgumentException iae){
+                throw new InternalExpressionException("Invalid b64 string: " + v.getString());
```
gnembon, line 160: throw a user-catchable error rather here, there is already a json_error, just add base64_error

Ghoulboy78: wdym user catchable?

src/main/java/carpet/script/language/ControlFlow.java
```diff
@@ -145,5 +151,23 @@ public static void apply(Expression expression) // public just to get the javado
                 return (c_, t_) -> retval;
             }
         });
+
+        expression.addUnaryFunction("encode_b64", v -> StringValue.of(Base64.getEncoder().encodeToString(v.getString().getBytes())));
```
gnembon, line 155: ControlFlow is for controlling flow of programs not for IO utils. Use Sys or DataStructures

src/main/java/carpet/script/language/ControlFlow.java
```diff
@@ -145,5 +151,23 @@ public static void apply(Expression expression) // public just to get the javado
...
+        expression.addUnaryFunction("decode_b64", v -> {
+            try {
+                return StringValue.of(new String(Base64.getDecoder().decode(v.getString()), StandardCharsets.ISO_8859_1));//using this charset cos it's the one used in decoding function
```
gnembon, line 158: Why not UTF-8?

src/main/java/carpet/script/language/ControlFlow.java
altrisi: ThrowStatement, from #692 + an edit I can't find the number rn. You have examples there.

(aka exceptions that can be caught by the user in a `try` block)

src/main/java/carpet/script/language/ControlFlow.java
Ghoulboy78: k

src/main/java/carpet/script/language/ControlFlow.java
Ghoulboy78: as I mentioned in the comment. that's the charset used in the `Base64.getDecoder.decode(String s)` function, so it makes sense to re-encode with the same charset.

src/main/java/carpet/script/language/ControlFlow.java
Ghoulboy78: oh ok

src/main/java/carpet/script/language/ControlFlow.java
gnembon: well entire scarpet base is in UTF-8 and you are potentially introducing a string with a different base

gnembon: kk, gotcha

gnembon: but you might need to test with some wierd emojis and stuff

gnembon:
> ScarpetJsonDeserializer

Yeah, if you move the GSON static to ScarpetJsonDeserializer and use it in both places

Ghoulboy78:
Ok this pr is ready to merge as well, I've tested and it works fine @gnembon

altrisi:
Mind adding the exceptions that are thrown to the docs of each function, and to the list of exceptions?

Ghoulboy78:
Sure thing mate

Ghoulboy78:
done

docs/scarpet/language/LoopsAndHigherOrderFunctions.md
```diff
@@ -146,3 +146,12 @@ it will be used from now on as a new value for the accumulator.
 reduce([1,2,3,4],_a+_,0)  => 10
 reduce([1,2,3,4],_a*_,1)  => 24
 </pre>
```
gnembon, line 148: Why in loops 🤦

PR #838, opened 2021-05-03
Adding Javadocs to CounterCommand.java and relevant files

Ghoulboy78:
This will involve updating:

- [x] All of CounterCommand.java
- [x] All of HopperCounter.java
- [x] Relevant parts of WoolTool.java
- [x] All of IngredientInterface.java
- [x] All of RecipeManagerInterface.java
- [x] All of HopperBlockEntiyMixin.java

### DIFF
--- a/src/main/java/carpet/helpers/HopperCounter.java
+++ b/src/main/java/carpet/helpers/HopperCounter.java
@@ -240,7 +240,7 @@ public class HopperCounter
     }
 
     /**
-     * Maps a number of items to a number of blocks, so we can get their colours easily with the
+     * Maps items that don't get a good block to reference for colour, or those that colour is wrong to a number of blocks, so we can get their colours easily with the
      * {@link Block#getDefaultMaterialColor()} method as these items have those same colours.
      */
     private static final ImmutableMap<Item, Block> DEFAULTS = new ImmutableMap.Builder<Item, Block>()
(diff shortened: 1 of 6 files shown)


src/main/java/carpet/helpers/HopperCounter.java
```diff
@@ -184,6 +239,10 @@ public static int appropriateColor(int color)
         return (r << 16) + (g << 8) + b;
     }
 
+    /**
+     * Maps a number of items to a number of blocks, so we can get their colours easily with the
```
gnembon, line 243:
```suggestion
     * Maps items that don't get a good block to reference for colour, or those that colour is wrong to a number of blocks, so we can get their colours easily with the
```

src/main/java/carpet/helpers/HopperCounter.java
```diff
@@ -292,6 +354,10 @@ else if (Registry.BLOCK.containsId(id))
         return null;
     }
 
+    /**
+     * Guesses the item's colour from the item itself. It first calls {@link HopperCounter#fromItem} to see if it has a
+     * valid colour there, if not just makes a guess, and if that fails just returns {@link HopperCounter#WHITE}
```
gnembon, line 359:
```suggestion
     * valid colour there, if not just makes a guess, and if that fails just returns null
```

src/main/java/carpet/mixins/HopperBlockEntityMixin.java
```diff
@@ -6,6 +6,7 @@
 import net.minecraft.block.entity.HopperBlockEntity;
 import net.minecraft.block.entity.LootableContainerBlockEntity;
 import net.minecraft.item.ItemStack;
+import net.minecraft.server.MinecraftServer;
```
gnembon, line 9: this shouldn't be there, right?

PR #840, opened 2021-05-04
Fixed a really annoying typo

DillonB07:
Just fixed a typo 'Hwat?' to 'What?'
Really small, but it annoys me(and probably many others) every time I open the page.


altrisi:
I feel like that's WAI

gnembon:
Hwat did you say?

PR #845, opened 2021-05-06
Fixing #843

### LINKED ISSUE #843: -x % x returns -0, which is apparently less than and equal to 0
When taking the modulus of a negative number, a result that should be 0 appears to be stored as -0, which scarpet reports as being equal to 0 **_and_** less than 0. This does not happen with positive numbers.
`3 % 3 < 0 => false`
`-3 % 3 == 0 => true`
`-3 % 3 < 0 => true`
`number(-3 % 3) < 0 => false`
Though the third output is the incorrect one, it would appear that, for now, using `number()` can alleviate this in scripts. A lower level fix would be appreciated, though.

Ghoulboy78:
Fixes #843 
The way I fix it is by ensuring that if -0.0D is inputted as the numeric value, it will automatically by put to positive 0.0D. The code looks confusing cos 0.0D returns equal to -0.0D,  but also returns less than 0.0D cos of the weirdness of IEEE 754 standard used for doubles in java.

### DIFF
--- a/src/main/java/carpet/script/value/NumericValue.java
+++ b/src/main/java/carpet/script/value/NumericValue.java
@@ -201,11 +201,12 @@ public class NumericValue extends Value
 
     public NumericValue(double value)
     {
-        this(value, (long) value);
+        this.value = value == 0.0D ? 0.0D : value;//to stop it from accidentally being -0.0D
+        this.longValue = (long) value;
     }
     private NumericValue(double value, Long longValue)
     {
-        this.value = value == 0.0D ? 0.0D : value;//to stop it from accidentally being -0.0D,
+        this.value = value == 0.0D ? 0.0D : value;//to stop it from accidentally being -0.0D
         this.longValue = longValue;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/carpet/script/value/NumericValue.java
```diff
@@ -220,7 +220,7 @@ public NumericValue(String value)
             }
             catch (ArithmeticException ignored) {}
         }
-        this.value = decimal.doubleValue();
+        this.value = decimal.doubleValue() == 0.0D ? 0.0D : decimal.doubleValue();//to stop it from accidentally being -0.0D,
```
Ghoulboy78, line 223: I didnt want to repeat this code but it wouldn't let me do otherwise

src/main/java/carpet/script/value/NumericValue.java
```diff
@@ -201,11 +201,11 @@ public boolean equals(Object o)
 
     public NumericValue(double value)
     {
-        this.value = value;
+        this(value, (long) value);
```
gnembon, line 204: why?

gnembon:
I would honestly just fix it in place when its caused Can you cause -0 anywhere using normal arithmetic or only calling '%' which needs a rework anyways?

src/main/java/carpet/script/value/NumericValue.java
Ghoulboy78: So then I can avoid repeating the if-else statement over and over ig? if u want I can remove ofc

Ghoulboy78:
I don't know if it can be caused anwhere, which is why I am adding it on the constructor of NumericValue so it never returns -0.0D to essentially fix it anywhere in the code where this bug may be lurking around

src/main/java/carpet/script/value/NumericValue.java
Ghoulboy78: actually nah Imma just remove it and add the if else thing there to. Literally the same thing so who cares

PR #847, opened 2021-05-07
Add the option to track and undo block state changes

Xendergo:
This would add the ability for scarpet scripts to easily track changes they make to the world, and undo them, or save them, using `track_changes()`, `undo_changes()`, and `commit_changes()`.

This would be especially useful for creative building scripts that apply modifications to an existing build. For example, a script to automatically do landscaping could save all the changes it makes, and if you decide there's too much tall grass or something, the script could undo, and go again with the new parameters.

### DIFF
--- a/docs/scarpet/api/BlocksAndWorldAccess.md
+++ b/docs/scarpet/api/BlocksAndWorldAccess.md
@@ -247,7 +247,7 @@ NB: It can thunder without there being a thunderstorm, there has to be both rain
 With two args, sets the weather to `type` for `ticks` ticks.
 
 ### `track_changes()`
-Keep track of any changes to the block states of blocks made by your scarpet script so they can be undone later if neccesary using `undo_changes()`. All changes tracked are undone when the script unloads, or if the server closes (unless you close a server with the x button, or it crashes). Does nothing if changes are already being tracked.
+Keep track of any changes to the block states of blocks made by your scarpet script so they can be undone later if neccesary using `undo_changes()`. All changes tracked are undone when the script unloads, or if the server closes (unless you close a server with the x button, or it crashes). Does nothing if changes are already being tracked. Only changes made using `set()` or `place_item()` are tracked.
 
 Returns a boolean for whether changes were being tracked before
 
--- a/docs/scarpet/api/BlocksAndWorldAccess.md
+++ b/docs/scarpet/api/BlocksAndWorldAccess.md
@@ -246,6 +246,57 @@ NB: It can thunder without there being a thunderstorm, there has to be both rain
 
 With two args, sets the weather to `type` for `ticks` ticks.
 
+### `track_changes()`
+Keep track of any changes to the block states of blocks made by your scarpet script so they can be undone later if neccesary using `undo_changes()`. All changes tracked are undone when the script unloads. Does nothing if changes are already being tracked.
+
+Returns a boolean for whether changes were being tracked before
+
+Only works in globally scoped apps
+
+```
+__config() -> {
+  'scope' -> 'global',
+  'commands' -> {
+    'block_one <block>' -> _(block) -> ( global_block1 = block; make_block(); ),
+    'block_two <block>' -> _(block) -> ( global_block2 = block; make_block(); ),
+    'ratio <ratio>' -> _(ratio) -> ( global_ratio = ratio; make_block(); ),
+    'commit' -> _() -> commit_changes()
+  },
+  'arguments' -> {
+    'block' -> { 'type' -> 'block' },
+    'ratio' -> { 'type' -> 'float', 'min' -> 0, 'max' -> 1 }
+  }
+};
+
+global_block1 = 'stone_bricks';
+global_block2 = 'cracked_stone_bricks';
+global_ratio = 0.5;
+
+make_block() -> (
+  undo_changes();
+  track_changes();
+  for (rect(l(0, 64, 0), l(5, 5, 5)),
+    if (rand(1) > global_ratio,
+      set(_, global_block1);
+    ,
+      set(_, global_block2);
+    );
+  );
+);
+
+make_block();
+```
+
+### `undo_changes()`
+Undo all the changes tracked by `track_changes()`. Stops tracking changes to block states
+
+Returns the amount of changes undone
+
+### `commit_changes()`
+Commit all changes tracked by `track_changes()` so they can't be undone. Stops tracking changes to block states
+
+Returns the amount of changes committed
+
 ## Block and World querying
 
 ### `pos(block), pos(entity)`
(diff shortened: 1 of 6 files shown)


docs/scarpet/api/BlocksAndWorldAccess.md
```diff
@@ -246,6 +246,57 @@ NB: It can thunder without there being a thunderstorm, there has to be both rain
 
 With two args, sets the weather to `type` for `ticks` ticks.
 
+### `track_changes()`
+Keep track of any changes to the block states of blocks made by your scarpet script so they can be undone later if neccesary using `undo_changes()`. All changes tracked are undone when the script unloads, or if the server closes (unless you close a server with the x button, or it crashes). Does nothing if changes are already being tracked.
+
+Returns a boolean for whether changes were being tracked before
+
+Only works in globally scoped apps
```
gnembon, line 254: I don't see a reason why this needs to be a thing. Explain

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -776,12 +777,18 @@ else if (args.get(0) instanceof MapValue)
 
             if (sourceBlockState == targetBlockState && data == null)
                 return Value.FALSE;
+
```
gnembon, line 780: what are these spaces for?

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -877,14 +886,14 @@ else if (item instanceof TridentItem || item instanceof SwordItem)
...
-                if (how < 0 || (tag != null && EnchantmentHelper.getLevel(Enchantments.SILK_TOUCH, tool) > 0))
+                if (fortuneLevel < 0 || (tag != null && EnchantmentHelper.getLevel(Enchantments.SILK_TOUCH, tool) > 0))
                 {
                     Block.dropStack(world, where, new ItemStack(state.getBlock()));
                 }
                 else
                 {
-                    if (how > 0)
-                        tool.addEnchantment(Enchantments.FORTUNE, (int) how);
+                    if (fortuneLevel > 0)
```
gnembon, line 895: honestly you would undo the blocks but can't undo the entities. Consider not track changes for harvest or destroy for that reason.

docs/scarpet/api/BlocksAndWorldAccess.md
Xendergo: For the returns thing, I didn't really see a reason not to, I'd be happy to remove it

For the globally scoped thing, I just couldn't figure out how to get it to work right for player scoped apps, and I figured it didn't matter much anyways, since the new features are more world focused. Also, if it would work in player scoped apps, it'd probably work such that each player would keep track of changes individually and undo individually which I think would be strange and not very useful behavior.

src/main/java/carpet/script/api/WorldAccess.java
Xendergo: If I'm not mistaken, those are places where I had code for debugging, but didn't bother or forgot to remove the line breaks when I removed the code, since it seemed like a natural place to have them.

docs/scarpet/api/BlocksAndWorldAccess.md
```diff
@@ -246,6 +246,57 @@ NB: It can thunder without there being a thunderstorm, there has to be both rain
 
 With two args, sets the weather to `type` for `ticks` ticks.
 
+### `track_changes()`
+Keep track of any changes to the block states of blocks made by your scarpet script so they can be undone later if neccesary using `undo_changes()`. All changes tracked are undone when the script unloads, or if the server closes (unless you close a server with the x button, or it crashes). Does nothing if changes are already being tracked.
```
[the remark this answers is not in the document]
Xendergo: Yeah definitely

src/main/java/carpet/script/api/WorldAccess.java
Xendergo: Yeah good idea, I'm pretty sure those functions are meant for different things anyways

docs/scarpet/api/BlocksAndWorldAccess.md
gnembon: consider a usecase - a veinminer. Its a typical player scoped usecase. All modified blocks are stored with the player script host, so not affecting other players. Makes perfect sense and the way you have implemented it - with storing the undo data in players' script host the support for multiple users comes automatic, doesn't it?

Xendergo: Well sort of, I noticed while writing the example that running the commit command wouldn't work properly, and the changes would undo anyways, and I thought it was some sort of strange side effect with how player scoping was implemented. Rereading the documentation, it sounds like this is because the data wasn't being copied correctly. Ima try and give it another go.

Xendergo: Ok, it seems to be working now for player scope

src/main/java/carpet/script/CarpetScriptHost.java
```diff
@@ -98,11 +99,12 @@ else if (parent != null)
         appArgTypes = argTypes;
         this.commandValidator = commandValidator;
         this.isRuleApp = isRuleApp;
+        this.blockChanges = blockChanges;
     }
 
     public static CarpetScriptHost create(CarpetScriptServer scriptServer, Module module, boolean perPlayer, ServerCommandSource source, Function<ServerCommandSource, Boolean> commandValidator, boolean isRuleApp)
     {
-        CarpetScriptHost host = new CarpetScriptHost(scriptServer, module, perPlayer, null, Collections.emptyMap(), new HashMap<>(), commandValidator, isRuleApp);
+        CarpetScriptHost host = new CarpetScriptHost(scriptServer, module, perPlayer, null, Collections.emptyMap(), new HashMap<>(), commandValidator, isRuleApp, new ArrayList<>());
```
gnembon, line 107: not sure you need that at all

src/main/java/carpet/script/ScriptHost.java
```diff
@@ -136,18 +136,26 @@ public void SaveBlockStateChange(World world, BlockPos blockPos) {
...
             CompoundTag tag = entity == null ? null : entity.toTag(new CompoundTag());
+            SavedBlockChange change = new SavedBlockChange(world, blockPos, world.getBlockState(blockPos), tag);
 
             synchronized (blockChanges) {
-                blockChanges.add(new SavedBlockChange(world, blockPos, world.getBlockState(blockPos), tag));
+                blockChanges.add(change);
+            }
+
+            // Copy changes made in the global state to user hosts
+            for (ScriptHost host : userHosts.values()) {
```
gnembon, line 146: why, the idea of user hosts is that other apps are not affected

gnembon:
not sure what you have done in https://github.com/gnembon/fabric-carpet/pull/847/commits/8a59b9e347b4da1ee02bb536e6ddda584e5ecb2a

Your previous code was should be working perfectly with player apps, not its messier. What the point of updating other player apps from global nor copy from global. Global app should not be accessible for player scope apps

Xendergo:
Honestly you're probably right.

The only downside I see would be that libraries don't work particularly well in scarpet, since to use one, you would have to either import the script and get everyone who wants to use it to download the library too, or to copy/paste the library into your own code. However, that would be a completely different problem

gnembon:
yeah, this a good example of something you could just do in the app to individual specification.

PR #848, opened 2021-05-07
Modify look

### LINKED ISSUE #846: [Suggestion] Add `modify(e, 'look', vector)`
Title says it all, it means we can just assign the entity to look at a vector. Should there be a requirement for the vector to be normalized or not tho? Idk, doesnt rly change much, just a call to normalize the vector tbh if thats necessary.

I would make the pr myself but I already have like 2 open and plus I don't know how to do the whole fiddling with yaw and pitch thing

replaceitem:
Added ability to modify entity look vector

Resolves #846

### DIFF
--- a/docs/scarpet/api/Entities.md
+++ b/docs/scarpet/api/Entities.md
@@ -589,6 +589,12 @@ Changes entity's location in the specified direction.
 
 Changes entity's pitch or yaw angle.
 
+### `modify(e, 'look', x, y, z), modify(e, 'look', l(x,y,z) )`
+
+Sets entity's 3d vector where the entity is looking.
+For cases where the vector has a length of 0, yaw and pitch won't get changed.
+When pointing straight up or down, yaw will stay the same.
+
 ### `modify(e, 'head_yaw', angle)`, `modify(e, 'body_yaw', angle)`
 
 For living entities, controls their head and body yaw angle.
--- a/docs/scarpet/api/Entities.md
+++ b/docs/scarpet/api/Entities.md
@@ -589,7 +589,7 @@ Changes entity's location in the specified direction.
 
 Changes entity's pitch or yaw angle.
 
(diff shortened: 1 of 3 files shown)


src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -1034,7 +1034,21 @@ private static void updateVelocity(Entity e, double scale)
...
         });
 
-        //"look"
+        put("look", (e, v) -> {
+            if (!(v instanceof ListValue)) throw new InternalExpressionException("Expected a list of 3 parameters as a second argument");
+            List<Value> vec = ((ListValue)v).getItems();
+            float x = NumericValue.asNumber(vec.get(0)).getFloat();
+            float y = NumericValue.asNumber(vec.get(1)).getFloat();
+            float z = NumericValue.asNumber(vec.get(2)).getFloat();
+            float l = MathHelper.sqrt(x*x + y*y + z*z);
```
gnembon, line 1043: I would protect from zero division here. Also yaw is undef afaik if you look straight up or down - in that case just make sure its ... zero?

replaceitem: Yeah  i was wondering what a vector that cant properly define pitch or yaw should do. Same case for looking straight down or up, then the yaw would always have the same value. I was thinking about checking for those cases and keeping player's yaw, or just set yaw to the same value, as it is now.

And yeah docs, completely forgot that

docs/scarpet/api/Entities.md
```diff
@@ -589,6 +589,12 @@ Changes entity's location in the specified direction.
 
 Changes entity's pitch or yaw angle.
 
+### `modify(e, 'look', x, y, z), modify(e, 'look', l(x,y,z) )`
```
gnembon, line 592:
```suggestion
### `modify(e, 'look', x, y, z), modify(e, 'look', [x,y,z] )`
```

gnembon (approved):

PR #851, opened 2021-05-10
Switch out l() syntax to [] in docs

replaceitem:
Replacing all cases where `l()` is still used for constructing lists with `[]`, exept of course those that explain the function version of it.

Thought this was needed since in 25f5a3771ded7a2bd27019a6cc88d59c6b327330 the `l()` got changed too, and i noticed the other modify() lists were also still `l()` so i searched through all docs and replaced it

### DIFF
--- a/docs/scarpet/api/BlocksAndWorldAccess.md
+++ b/docs/scarpet/api/BlocksAndWorldAccess.md
@@ -2,7 +2,7 @@
 
 ## Specifying blocks
 
-### `block(x, y, z)`, `block(l(x,y,z))`, `block(state)`
+### `block(x, y, z)`, `block([x,y,z])`, `block(state)`
 
 Returns either a block from specified location, or block with a specific state (as used by `/setblock` command), 
 so allowing for block properties, block entity data etc. Blocks otherwise can be referenced everywhere by its simple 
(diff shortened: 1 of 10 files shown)


gnembon:
all makes sense. Thanks :)

docs/scarpet/api/BlocksAndWorldAccess.md
```diff
@@ -91,7 +91,7 @@ generator. The following code causes a cascading effect as blocks placed on chun
 loaded to full, thus generated:
 
 <pre>
-__config() -> m(l('scope', 'global'));
+__config() -> m(['scope', 'global']);
```
gnembon, line 94: might as well apply map syntax, so  `-> {'scope' -> 'global'};`

docs/scarpet/api/BlocksAndWorldAccess.md
```diff
@@ -104,7 +104,7 @@ __on_chunk_generated(x, z) -> (
 The following addition resolves this issue, by not allowing block updates past chunk borders:
 
 <pre>
-__config() -> m(l('scope', 'global'));
+__config() -> m(['scope', 'global']);
```
gnembon, line 107:
```suggestion
__config() -> {'scope' -> 'global'};
```

PR #856, opened 2021-05-12
Java 16 build compatibility

altrisi:
Well, Minecraft is now Java 16...

Meaning Carpet will at some point need to be compatible with it when building too.

This PR changes the actions to run using Java 16 instead of 8 (there are no problems with compiling for older java versions until 1.17 releases, since target version is still 8) and updates the buildscript to use Gradle 7.0 (needed for Java 16 compatibility) and Loom 0.6 (needed for compatibility with Gradle 7), including a few fixes for removed features.

Also updates `setup-java` action step to v2 while at it, and makes the ci builds run when any file is changed, not only in src, just cause why not, that way there is always a green check if it works.

~Still a draft since I've just done this fast in the Github web app and I have to test it.~ Tested and it works, both ci and releases (once it's merged to 1.17 branch ofc).

### DIFF
--- a/.github/workflows/build-on-pr.yml
+++ b/.github/workflows/build-on-pr.yml
@@ -15,7 +15,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 1.8
+    - name: Set up JDK 16
       uses: actions/setup-java@v1
       with:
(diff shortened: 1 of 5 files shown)


gnembon:
missed your pr earlier - got to the same results pretty much

PR #859, opened 2021-05-13
Use primitive specific Predicate instead of Boolean Function

altrisi:
A bit more compact and prevents autoboxing. Was mostly bored.

Also (completely unrelated to this PR), Carpet for releases could still be built using Loom 0.7, since that's supposedly more stable and has support for building using Java 8 through 16 (rn Carpet requires Java 16 to be built since Loom 0.8 targets it).

Edit: `test.test()` may not be the most expressive name for the function, yeah, but at that time I just changed the types.

### DIFF
--- a/src/main/java/carpet/script/CarpetScriptHost.java
+++ b/src/main/java/carpet/script/CarpetScriptHost.java
@@ -56,7 +56,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.function.BiPredicate;
 import java.util.function.Consumer;
-import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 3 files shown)


gnembon:
loom 8 was recommended by the fabric peeps. Since I am not a framework type'a guy, I would go with the flo.

gnembon:
Yeah Predicate / Function same thing. Likely saving a cycle or two by using a predicate here. since null is unused.

altrisi:
> loom 8 was recommended by the fabric peeps. Since I am not a framework type'a guy, I would go with the flo.

Yeah, 0.8 is the only one that works for the new snapshot, but modmuss mentioned in Fabricord#loom that for 1.16 Loom 0.7 is the recommended. Just mentioning so releases for 1.16 don't have the "may be unstable" tag and can also be built with lower Java, for those who haven't updated yet, or for those who have problems with 0.8 (there are a few known bugs in some setups).

altrisi:
Actually, I guess as long as it works it should be fine and may be better for the IDE to not have to be switching the entire toolchain when changing branches. If it fails for someone I guess they can downgrade loom manually, it's just a number.

gnembon:
Yeah, I need to switch pretty much every commit to keep 1.17 in sync.

gnembon (approved):

PR #860, opened 2021-05-13
Remove the number_classes set in favour of ClassUtils

altrisi:
That was a thing of mine when fixing `getBoolValue` for Scarpet rules.

This new way removes the set in favour of using Apache's ClassUtils `primitiveToWrapper` method, which gets the wrapper class, that actually extends `Number`.

Technically fixes rules (from extensions maybe?) using wrappers as their numbers. However other parts checking specifically for primitives aren't changed (e.g. checks for `boolean.class`) (would it be worth setting `type` to the wrapper and only check equality with wrappers? Not sure)

### DIFF
--- a/src/main/java/carpet/settings/ParsedRule.java
+++ b/src/main/java/carpet/settings/ParsedRule.java
@@ -3,7 +3,6 @@ package carpet.settings;
 import carpet.utils.Translations;
 import carpet.utils.Messenger;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Sets;
 import net.minecraft.server.command.ServerCommandSource;
 
 import java.lang.reflect.Constructor;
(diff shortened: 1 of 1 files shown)


gnembon:
well, the previous method seems much more intuitive of what is doing. `primitiveToWrapper` is a confusing name to say the least. WDYM a rule is using a wrapper as the number? Can you provide an example?

altrisi:
Doesn't happen in Carpet, but given SettingsManager is public API-ish, someone who is new in Java could do something like:

```java
@Rule(...)
public Integer something = 5;
```

A very edge case, but I guess could happen. Tbh I was mostly bored when I found this again and remembered of `ClassUtils`.

Now seeing it again, it may be more intuitive, true.

gnembon (approved):

PR #865, opened 2021-05-17
Fix dedicated server crash with hopper counters

### LINKED ISSUE #864: [Bug]In the server ,using /counter will kill the server,when hopperCounter counts bone_meal
In the server ,using /counter will kill the server,when hopperCounter counts bone_meal
Release version：1.14.33+
[2021-05-17-7.log](https://github.com/gnembon/fabric-carpet/files/6491644/2021-05-17-7.log)

altrisi:
Fixes #864.

The new color thing was using a method that is only present in the client (`containsId`), so I changed it to `getOrEmpty().isPresent()`. Also saved the result of `getOrEmpty` so there's no need to poll the registry twice.

### DIFF
--- a/src/main/java/carpet/helpers/HopperCounter.java
+++ b/src/main/java/carpet/helpers/HopperCounter.java
@@ -40,6 +40,7 @@ import java.util.EnumMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


gnembon:
Thanks for helping - didn't get to this one.

PR #872, opened 2021-05-21
Allow `/player spawn` spawn offline-mode player in online server; Code clean-up in `/player spawn` command

keuin:
First I did a clean-up in method `PlayerCommand.spawn` and `EntityPlayerMPFake.createFake`, making them cleaner and easy-to-read.

Then I added a carpet configuration named `fallBackToSpawningOfflinePlayers` which allows `/player spawn <playerName>` to spawn offline-mode players if there is no online-mode player has the name `<playerName>`, which allows you spawning fake players with more freedom on their names. If online-mode player with the specified ID exists, `/player spawn` will spawn an online-mode player, which behaves identical with previous versions.


altrisi:
Opinion: If it works, I don't think the rule is necessary and Carpet should allow spawning those players by default. Haven't checked the code yet.

keuin:
The switch is set to `false` and thus won't affect any feature by default.

gnembon:
sorry but code refactors are super unfun with two parallel versions of carpet running for two minecraft versions.

also ssuggesting using more 'noob friendly' name for 'fallBackToSpawningOfflinePlayers'

src/main/java/carpet/CarpetSettings.java
```diff
@@ -28,17 +28,7 @@
...
-import static carpet.settings.RuleCategory.CREATIVE;
-import static carpet.settings.RuleCategory.EXPERIMENTAL;
-import static carpet.settings.RuleCategory.FEATURE;
-import static carpet.settings.RuleCategory.OPTIMIZATION;
-import static carpet.settings.RuleCategory.SURVIVAL;
-import static carpet.settings.RuleCategory.TNT;
-import static carpet.settings.RuleCategory.DISPENSER;
-import static carpet.settings.RuleCategory.SCARPET;
-import static carpet.settings.RuleCategory.CLIENT;
+import static carpet.settings.RuleCategory.*;
```
gnembon, line 31: please no

gnembon:
don't import stars
don't reorder imports
don't refactor just to make it clearer
 - these last two are super important to track history changes even in between snapshots where only one branch is really important. otherwise tracking it gets unnecessarily complicated.

src/main/java/carpet/CarpetSettings.java
keuin: I expanded the wildcard imports and reverted the reordering, keeping intact with your main stream.

keuin:
Code changes are minimized and non-essential code refactors are reverted. Sorry for my casual modifications.

The method `cantSpawn` is inlined to method `EntityPlayerMPFake.createFake` because `cantSpawn` takes duplicate pre-condition checking and offline fake players can't be spawned unless the checking is removed.

keuin:
As for the switch naming, I struggled to make some:

- `spawnOfflineFakePlayerIfUnavailable`
- `spawnOfflinePlayerInOnlineModeIfUnavailable`
- `allowSpawningOfflineFakePlayerInOnlineMode`

I prefer the second one. It is a little bit longer, but it is clearer.

src/main/java/carpet/patches/EntityPlayerMPFake.java
```diff
@@ -11,45 +11,87 @@
...
+        PlayerManager playerManager = server.getPlayerManager();
+
+        // if this player already logged in
+        if (playerManager.getPlayer(playerName) != null)
+        {
+            throw new FakePlayerSpawnException("r Player ", "rb " + playerName, "r  is already logged on");
+        }
+
+        // check player name length
+        // player with a name longer than 16 bytes will kick out all players
```
gnembon, line 52: not true - works in SP up to 40 chars

src/main/java/carpet/patches/EntityPlayerMPFake.java
```diff
@@ -11,45 +11,87 @@
...
+        // if the player with this playerName does not exist
+        if (gameprofile == null)
+        {
+            if (!fallBackToSpawningOfflinePlayer) {
+                throw new FakePlayerSpawnException("r Player " + playerName + " is either banned by Mojang, or " +
+                        "auth servers are down. Banned players can only be summoned in Singleplayer " +
+                        "and in servers in off-line mode.");
+            } else {
+                // spawn offline player
+                gameprofile = new GameProfile(PlayerEntity.getOfflinePlayerUuid(playerName), playerName);
```
gnembon, line 68: isn't that THE only line that really needed to be added to have the mvp in this situation?

src/main/java/carpet/commands/PlayerCommand.java
```diff
@@ -178,37 +176,6 @@ private static boolean cantReMove(CommandContext<ServerCommandSource> context)
...
-        PlayerEntity player = manager.getPlayer(playerName);
-        if (player != null)
-        {
-            Messenger.m(context.getSource(), "r Player ", "rb " + playerName, "r  is already logged on");
-            return true;
-        }
-        GameProfile profile = server.getUserCache().findByName(playerName);
-        if (profile == null)
-        {
-            Messenger.m(context.getSource(), "r Player "+playerName+" is either banned by Mojang, or auth servers are down. " +
```
gnembon, line 195: conditionally allow it here, by the rule.

src/main/java/carpet/patches/EntityPlayerMPFake.java
keuin: No, it isn't.
In file `PlayerCommand.java`, method `cantSpawn` checks the specified fake player name:

```java
    private static boolean cantSpawn(CommandContext<ServerCommandSource> context)
    {
        String playerName = StringArgumentType.getString(context, "player");
        MinecraftServer server = context.getSource().getMinecraftServer();
        PlayerManager manager = server.getPlayerManager();
        PlayerEntity player = manager.getPlayer(playerName);
        if (player != null)
        {
            Messenger.m(context.getSource(), "r Player ", "rb " + playerName, "r  is already logged on");
            return true;
        }
        GameProfile profile = server.getUserCache().findByName(playerName);
        if (profile == null)
        {
            Messenger.m(context.getSource(), "r Player "+playerName+" is either banned by Mojang, or auth servers are down. " +
                    "Banned players can only be summoned in Singleplayer and in servers in off-line mode.");
            return true;
        }
// ...
```

And if the online-mode player with specified name does not exist, this method will return `true`, which makes `spawn` method in `PlayerCommand.java` returns immediately:

```java
    private static int spawn(CommandContext<ServerCommandSource> context) throws CommandSyntaxException
    {
        if (cantSpawn(context)) return 0;
// ...
```

If only the line you mentioned was added, that line will never be reached.

gnembon:
regarding namng, some more general wording might be more useful and potentially may extend on other usecases, like `strictOnlineMode` albeit I am not set on that either.

I have a feeling that this change can just be done in two if statements, one in 'cantSpawn` and the other in profile creations in constructor, unless I am wrong.

src/main/java/carpet/patches/EntityPlayerMPFake.java
keuin: Well, I saw that another PR is focusing on this problem, I will remove it from this PR.

src/main/java/carpet/patches/EntityPlayerMPFake.java
gnembon: well, I added the other comment above later
Allow it there (by the new rule): https://github.com/gnembon/fabric-carpet/pull/872/files#r637356921
and make it happen here, right?

keuin: Yeah I re-implemented this feature in that way.

keuin:
This feature is implemented without touching any other logic. I've tested it and it works in SMP and SSP.
However, I suggest cleaning method `createFake` and `cantSpawn` because currently they are fragile.

altrisi:
Still don't get why should this be under a rule, since it's not like it's changing vanilla behaviour (more than fake players themselves are changing it), just upgrading Carpet behaviour. But I guess it's up to you.

I think such `strictOnlineMode` will actually confuse players and make them think they can use it to toggle online mode on and off during runtime.

keuin:
People might accidentally spawn offline mode fake players which are not expected. A switch will give them control of the result. If they have not read the changelog, users may be confused.
I am not much inspired about naming, feel free to change it.

gnembon:
maybe name is not that important, but I agree with keuin - as an admin, you may not want to choose one behaviour or the other. If you are using a certain player on and on for specific task, and with poor connection or something you may by acccident spawn a different player. Making sure they are all validated could be a tool an admin may want to use. Open for suggestions for a better, yet still short and to the point rule name for it.

DragonEggBedrockBreaking:
I have some ideas:
`allowOfflineFakePlayers`
`allowSpawningOfflinePlayers`
`allowOfflinePLayerFallback`

gnembon:
shorter, to the point, I like them

src/main/java/carpet/CarpetSettings.java
```diff
@@ -472,6 +472,9 @@ public String description() {
     @Rule(desc = "Enables /player command to control/spawn players", category = COMMAND)
     public static String commandPlayer = "ops";
 
+    @Rule(desc = "Spawn offline players in online mode if online-mode player with specified name does not exist", category = COMMAND)
+    public static boolean strictOnlineMode = true;
```
gnembon, line 476:
```suggestion
    public static boolean allowSpawningOfflinePlayers = true;
```

src/main/java/carpet/commands/PlayerCommand.java
```diff
@@ -192,9 +192,14 @@ private static boolean cantSpawn(CommandContext<ServerCommandSource> context)
         GameProfile profile = server.getUserCache().findByName(playerName);
         if (profile == null)
         {
-            Messenger.m(context.getSource(), "r Player "+playerName+" is either banned by Mojang, or auth servers are down. " +
-                    "Banned players can only be summoned in Singleplayer and in servers in off-line mode.");
-            return true;
+            if (CarpetSettings.strictOnlineMode)
```
gnembon, line 195:
```suggestion
            if (!CarpetSettings.allowSpawningOfflinePlayers)
```

src/main/java/carpet/patches/EntityPlayerMPFake.java
```diff
@@ -45,7 +45,15 @@ public static EntityPlayerMPFake createFake(String username, MinecraftServer ser
         finally {
             UserCache.setUseRemote(server.isDedicated() && server.isOnlineMode());
         }
-        if (gameprofile == null) return null;
+        if (gameprofile == null)
+        {
+            if (CarpetSettings.strictOnlineMode)
```
gnembon, line 50:
```suggestion
            if (!CarpetSettings.allowSpawningOfflinePlayers)
```

gnembon (approved):

DragonEggBedrockBreaking:
> shorter, to the point, I like them

yay, I did something useful \o/

PR #873, opened 2021-05-21
Fix fake player name too long (length > 16) cause server kick all player

Merrg1n:
In online server, the length of fake player name is restricted by Mojang's auth server, so when you spawn a fake player with name length > 16, carpet will report "cannot spawn in online mode" instead of "name is too long". But in offline server, client can spawn fake player which name's length > 16 and < 41, that cause server kick out all client and refuse client to connect server.

Here is a video about this problem: https://streamable.com/y3dcnn

### DIFF
--- a/src/main/java/carpet/commands/PlayerCommand.java
+++ b/src/main/java/carpet/commands/PlayerCommand.java
@@ -267,7 +267,7 @@ public class PlayerCommand
         }
         catch (CommandSyntaxException ignored) {}
         String playerName = StringArgumentType.getString(context, "player");
-        if (playerName.length()>40)
+        if (playerName.length()>16)
         {
             Messenger.m(context.getSource(), "rb Player name: "+playerName+" is too long");
             return 0;


gnembon:
I can spawn a player that has more than 16 characters. If players with more than 16 characters makes problems in some scenarios, target these issues

Merrg1n:
In `net.minecraft.network.packet.s2c.play.PlayerListS2CPacket` class, method `write` construct packet without check the length of user name.

Merrg1n:
My yarn version is `1.16.5+build.1-v2`.

In class `net.minecraft.network.packet.s2c.play.PlayerListS2CPacket`, method `write` constructs packets without checking the length of user name.

```java
public void write(PacketByteBuf buf) throws IOException {
// L109-117
        while(true) {
            while(var2.hasNext()) {
                PlayerListS2CPacket.Entry entry = (PlayerListS2CPacket.Entry)var2.next();
                switch(this.action) {
                case ADD_PLAYER:
                    buf.writeUuid(entry.getProfile().getId());
                    buf.writeString(entry.getProfile().getName());
                    buf.writeVarInt(entry.getProfile().getProperties().size());
                    Iterator var4 = entry.getProfile().getProperties().values().iterator();
// ...
```

But the method `read`, which decodes server packets, will check the length of user name.

```java
public void read(PacketByteBuf buf) throws IOException {
// L59-63
         switch(this.action) {
         case ADD_PLAYER:
            gameProfile = new GameProfile(buf.readUuid(), buf.readString(16));
            int l = buf.readVarInt();
            int m = 0;
// ...
```

Merrg1n:
My yarn version is `1.16.5+build.1-v2`.

In class `net.minecraft.network.packet.s2c.play.PlayerListS2CPacket`, method `write` constructs packets without checking the length of user name.

```java
public void write(PacketByteBuf buf) throws IOException {
//L109-117
        while(true) {
            while(var2.hasNext()) {
                PlayerListS2CPacket.Entry entry = (PlayerListS2CPacket.Entry)var2.next();
                switch(this.action) {
                case ADD_PLAYER:
                    buf.writeUuid(entry.getProfile().getId());
                    buf.writeString(entry.getProfile().getName());
                    buf.writeVarInt(entry.getProfile().getProperties().size());
                    Iterator var4 = entry.getProfile().getProperties().values().iterator();
//......
```

But the method `read`, which decodes server packets, will check the length of user name.

```java
public void read(PacketByteBuf buf) throws IOException {
//L59-63
         switch(this.action) {
         case ADD_PLAYER:
            gameProfile = new GameProfile(buf.readUuid(), buf.readString(16));
            int l = buf.readVarInt();
            int m = 0;
//......
```

Merrg1n:
> I can spawn a player that has more than 16 characters. If players with more than 16 characters makes problems in some scenarios, target these issues

I can only reproduce this problem in multiplayer, singleplayer doesn't seem to have this restriction.

altrisi:
I was going to say a check for client environment could be added, however that would still kick people connecting to LAN worlds.

gnembon:
Well, don't need to check for environment - if the server is integrated and port is not open, I would still allow to 40 chars, otherwise (dedicated or integrated with ports open) should limit to 16. Now - you can start an integrated server, and then open it for others, but I woudn't worry about that.

gnembon:
cramming all that logic is probably not worth to do that here in github. Will do that independently

gnembon:
https://github.com/gnembon/fabric-carpet/commit/f2e759130a7c20123f37b583a1830edf64044a44

PR #877, opened 2021-05-23
Added rule `updateSuppressionBlock`

fxmorin:
Pull request for https://github.com/fxmorin/UpdateSuppressionBlock


src/main/java/carpet/mixins/BarrierBlock_updateSuppressionBlockMixin.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Random;
+
+@Mixin(BarrierBlock.class)
+public class BarrierBlock_updateSuppressionBlockMixin extends Block {
+    private boolean shouldPower = false;
+
+    public BarrierBlock_updateSuppressionBlockMixin(Settings settings) { super(settings); }
+
+    @Override
+    public int getWeakRedstonePower(BlockState state, BlockView world, BlockPos pos, Direction direction) {
```
gnembon, line 25: Why you are changing this behaviour unconditinally?

fxmorin: getWeakRedstonePower does not exist in the BarrierBlock so I need to override it. shouldPower can only be changed within functions that have rule checks. Therefore it should keep its normal behavior. Unless you meant something else?

PR #880, opened 2021-05-24
Use primitive doubles in `NumericValue` and allow adding exhaustion with floats

### LINKED ISSUE #879: `/script run modify(p, 'add_exhaustion', value)` casts `value` to an integer before applying the exhaustion
https://github.com/gnembon/fabric-carpet/blob/1b39feaab7a19fed659542a41d6ce95b089dd347/src/main/java/carpet/script/value/EntityValue.java#L1460-L1462

The above code means that `/script run modify(p, 'add_exhaustion', 1.5)` will actually add 1 exhaustion instead of 1.5, and `modify(p, 'add_exhaustion', value` where `value < 1` will do nothing as
```java
(int) NumericValue.asNumber(v).getLong()
```
rounds the given value to the nearest integer towards zero, truncating any decimals, then
```java
((PlayerEntity) e).getHungerManager().addExhaustion(0); // see above
```
does nothing to change the player's exhaustion level.

altrisi:
Title.

Resolves #879.

Changes `NumericValue` to hold a primitive double, I guess should help with heavy loads with numbers by avoiding autoboxing and autounboxing (there were no `null` checks or assignations to that variable) (haven't tested it, at least yet, just no compile errors).

### DIFF
--- a/src/main/java/carpet/script/value/EntityValue.java
+++ b/src/main/java/carpet/script/value/EntityValue.java
@@ -1,6 +1,5 @@
 package carpet.script.value;
 
-import carpet.CarpetServer;
 import carpet.fakes.BrainInterface;
 import carpet.fakes.EntityInterface;
 import carpet.fakes.ItemEntityInterface;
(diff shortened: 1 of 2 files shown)


gnembon:
Yeah, that should work. yolomerge

PR #881, opened 2021-05-24
Deprecate old boolean constructor and move existing usages

altrisi:
Shouldn't change behaviour since `BooleanValue` extends `NumericValue`.

I'm currently in the process of finding whether other extensions are using the constructor in order to make it protected (for boolean and null constructors) instead of deprecated (couldn't those use the numeric constructor with `0`?).
Update: Used by Carpet TIS Addition


altrisi:
Current users of the constructor:
- Carpet TIS Addition

gnembon:
Now lets pray to the auto-merging gods that 1.17 branch won't go berserk because of that.

PR #882, opened 2021-05-24
Replace usages of deprecated ListValue::fromItemStack and remove it

altrisi:
It has been deprecated since over 6 months ago (9e8556f9d9504651df0022a57edddd2487dcb48b). Couldn't really find any usage outside of Carpet.


gnembon:
same mergeyolo

PR #884, opened 2021-05-24
Implement `@Locator.Block`

altrisi:
Does so by making the end function use an iterator, that for current functions comes from a `params.listIterator(offset)` call in order to keep current Scarpet functions working.

Small note: `BlockArgument.findInValues(...)` seems to do exactly the same as `findIn` (assuming it's from before the lazy refactor). I think it could be dropped. Edit: Nothing seems to call it.

About the other locators: I don't really know how to implement them (at least yet) since they require list length, which an iterator can't provide.

### DIFF
--- a/src/main/java/carpet/script/annotation/Locator.java
+++ b/src/main/java/carpet/script/annotation/Locator.java
@@ -8,6 +8,7 @@ import java.util.Iterator;
 
 import org.apache.commons.lang3.NotImplementedException;
 
+import carpet.script.CarpetContext;
 import carpet.script.Context;
 import carpet.script.argument.Argument;
 import carpet.script.argument.BlockArgument;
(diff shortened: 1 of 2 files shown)


altrisi:
Added a few more return types to allow using the locator in a more direct way. Now it can also be used in BlockPos and BlockState, and will call those methods from BlockValue. Therefore it supports those plus BlockArgument and BlockValue.

src/main/java/carpet/script/annotation/Locator.java
```diff
@@ -140,7 +141,7 @@ private Locators() {}
 
         private static class BlockLocator<R> extends AbstractLocator<R>
         {
-            private final boolean returnBlockValue;
+            private final java.util.function.Function<BlockArgument, R> returnFunction;
```
gnembon, line 144: any reason can't import java.util.function.Function?

altrisi: `Locator.Function`, since it technically is just `Function`. Not sure if I really tried to manually import though, but I think it didn't let me.

Maybe I could give it its own `FunctionalInterface` to make it easier to look at.

gnembon (approved):

PR #891, opened 2021-05-29
Implement `@Locator.Function` and `@Locator.Vec3d`

altrisi:
### `@Locator.Function`
By basically consuming the whole iterator, since that's what the locator was going to do anyway.

Small thing: Isn't the `offset` value of LocatorArgument incorrect in case of varargs? It isn't really used anyway since everything is consumed and passed to the args, but just mentioning.

### `@Locator.Vec3d`
Easier than expected, just had to change the size check for an `iterator.hasNext()`, since it's in the exact position.

### DIFF
--- a/src/main/java/carpet/script/annotation/Locator.java
+++ b/src/main/java/carpet/script/annotation/Locator.java
@@ -8,6 +8,8 @@ import java.util.Iterator;
 
 import org.apache.commons.lang3.NotImplementedException;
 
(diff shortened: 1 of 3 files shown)


gnembon:
re: small things. Possibly, likely, just funcions are so far always defined at the end and only one of them. Migth be an issue for functions accepting multiple funcitons - in that case we would need to restrict it to one value - either name of a named zero-argument function, or a lambda definition.

PR #894, opened 2021-05-30
Add Movable Amethyst Rule

hugeblank:
As discussed in Episode 79 of ilmangos peaceful challenge. 
He said he'd do it, I figured I could beat him to the punch. :)

Context: https://youtu.be/0-fAKpr_J8k?t=1180

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -315,6 +315,9 @@ public class CarpetSettings
     )
     public static boolean hopperCounters = false;
 
+    @Rule( desc = "Allows Budding Amethyst blocks to be pushed by pistons", category = FEATURE )
+    public static boolean pushableAmethyst = false;
+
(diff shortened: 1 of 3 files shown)


src/main/java/carpet/CarpetSettings.java
```diff
@@ -315,6 +315,9 @@ public String description() {
     )
     public static boolean hopperCounters = false;
 
+    @Rule( desc = "Allows Budding Amethyst blocks to be pushed by pistons", category = FEATURE )
+    public static boolean pushableAmethyst = false;
```
DragonEggBedrockBreaking, line 319: Wouldn't `movableAmethyst` be more consistent, considering carpet's `movableBlockEntities` rule?

hugeblank: Agreed, good thinking. I'll get on renaming

gnembon:
Suggest extending to work with silk touch as well. This will give a choice. Since these two mechanics are not colliding suggesting implementing under the same 'true' value.

gnembon:
snappsshot today - will add that manually

PR #913, opened 2021-06-11
Re-implement `fastRedstoneDust` Optimization in 1.16.5+

SirDavidLudwig:
This PR re-introduces the `fastRedstoneDust` optimization in 1.16.5+ that seems to have been removed after the changes to redstone dust in snapshot 20w18a.

My assumption for the removal of this optimization was due to the changes of the vanilla redstone wire code. These changes did not seem as significant as they appeared to be as they consisted primarily of code reorganization. I reworked the `RedstoneWireBlockMixin` class around these changes so that it functions in the same way as the previous successful implementation while keeping the changes to a minimum.

Any feedback is greatly appreciated, and hopefully we can get this optimization back into the Carpet mod!

### DIFF
--- a/src/main/java/carpet/mixins/RedstoneWireBlockMixin.java
+++ b/src/main/java/carpet/mixins/RedstoneWireBlockMixin.java
@@ -3,6 +3,7 @@ package carpet.mixins;
 import carpet.CarpetSettings;
 import carpet.fakes.RedstoneWireBlockInterface;
 import carpet.helpers.RedstoneWireTurbo;
+import com.google.common.collect.Sets;
 import net.minecraft.block.Block;
 import net.minecraft.block.BlockState;
 import net.minecraft.block.RedstoneWireBlock;
@@ -27,22 +28,12 @@ import static net.minecraft.block.RedstoneWireBlock.POWER;
 
 @Mixin(RedstoneWireBlock.class)
 public abstract class RedstoneWireBlockMixin implements RedstoneWireBlockInterface {
-    /*
-
-    needs reworking
 
     @Shadow
-    private boolean wiresGivePower;
+    private void update(World world_1, BlockPos blockPos_1, BlockState blockState_1) { }
 
     @Shadow
-    @Final
-    private Set<BlockPos> affectedNeighbors;
-
-    @Shadow
-    private int increasePower(int int_1, BlockState blockState_1) { return 0; }
-
-    @Shadow
-    private BlockState update(World world_1, BlockPos blockPos_1, BlockState blockState_1) { return null; }
+    private int getReceivedRedstonePower(World world, BlockPos pos) { return 0; }
 
     @Override
     @Accessor("wiresGivePower")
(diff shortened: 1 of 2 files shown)


DragonEggBedrockBreaking:
Since this rule has the potential to fix locational contraptions, as stated in the rule description, then why not just use [the alternate current mod](https://github.com/SpaceWalkerRS/alternate-current) to go the full way, if this anyway causes issues?

SirDavidLudwig:
@DragonEggBedrockBreaking Oh, I was actually not aware of that mod. I have just been missing this optimization which is why I looked into reimplementing it.

altrisi:
This would need a thorough review if getting back to Carpet and quite some testing to make sure it doesn't break anything (IMO, as this would get auto-enabled if someone still has the rule in their config file, potentially breaking fragile mechanisms).
What are your preliminary test results? What did you test? Did anything fail or looked different than vanilla?

src/main/java/carpet/mixins/RedstoneWireBlockMixin.java
```diff
@@ -64,96 +55,73 @@ private void onRedstoneWireBlockCTOR(Block.Settings settings, CallbackInfo ci) {
...
+    @Inject(method = "update", at = @At("HEAD"), cancellable = true)
+    private void updateLogicAlternative(World world, BlockPos pos, BlockState state, CallbackInfo cir) {
+        if (CarpetSettings.fastRedstoneDust) {
+            updateLogicPublic(world, pos, state);
+            cir.cancel();
+        }
     }
 
     @Override
     public BlockState updateLogicPublic(World world_1, BlockPos blockPos_1, BlockState blockState_1) {
```
gnembon, line 79: This method looks pretty much identical to vanilla right now, is that true? Then what's the reason for mixin into 'update'

gnembon:
Thanks for looking into it. Indeed the changes don't seem to be drastic, besides one method which seems to come from vanilla.

gnembon:
A few initial tests suggest it works just fine. as it used to. I only have question about that method that looks like it doesn't need to be modified, unless I am missing something - which is possibly the case.

There is no rush of releasing - it looks good to me - can be disabled, but if you want to do some extra testing, feel free.

src/main/java/carpet/mixins/RedstoneWireBlockMixin.java
```diff
@@ -64,96 +55,73 @@ private void onRedstoneWireBlockCTOR(Block.Settings settings, CallbackInfo ci) {
...
-            if (world_1.getBlockState(blockPos_1) == blockState_2) {
+        int i = this.getReceivedRedstonePower(world_1, blockPos_1);
+        BlockState blockState = blockState_1;
+        if (blockState_1.get(POWER) != i) {
+            blockState_1 = blockState_1.with(POWER, i);
+            if (world_1.getBlockState(blockPos_1) == blockState) {
                 world_1.setBlockState(blockPos_1, blockState_1, 2);
             }
 
             if (!CarpetSettings.fastRedstoneDust) {
```
gnembon, line 88: Ok - its easy to miss what's not highlighted. Yeah looks good to me.

I will let the challengers test it further but it looks to me like updated theo's redstone.

SirDavidLudwig: Yes, the update method is only slightly modified, in this case to return the updated state. I found it to be the easiest way to get back to a successful implementation as without returning the updated state seems to be a bit more involved and would require modifying the redstone turbo class which I wanted to avoid as to not break anything.

gnembon: makes sense

PR #914, opened 2021-06-11
Made deepslate generate below y16

replaceitem:
Instead of only generating below y0, deepslate generates below y16 (where deepslate blobs are found too)
This is to make renewable deepslate work in standard 1.17 worlds as well, without extended world height below y0

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -769,7 +769,8 @@ public class CarpetSettings
     public static boolean renewableBlackstone = false;
 
     @Rule(
-            desc = "Lava and water generate deepslate and cobbled deepslate instead below y16",
+            desc = "Lava and water generate deepslate and cobbled deepslate instead below Y16",
+            extra = "This rule may change Y value to 0 with 1.18",
             category = FEATURE
     )
     public static boolean renewableDeepslate = false;
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -769,7 +769,7 @@ public class CarpetSettings
     public static boolean renewableBlackstone = false;
 
     @Rule(
-            desc = "Lava and water generate deepslate and cobbled deepslate instead below y0",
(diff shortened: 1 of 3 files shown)


altrisi:
I wonder if there's a function somewhere to make it rely on whether deepslate can generate in the world, instead of hardcoding it (I guess there is somewhere, hopefully deepslate gen is not that hardcoded). Would prevent future breakages and maintain consistency if using a/the datapack (does the caves and cliffs datapack change this height? Never played with it).

gnembon:
well, you have to enable it explicitly so hard to tell if will break stuff. You could check the 'deepslate_enabled' option in worldgen settings, but from my initial checking it was quite well hidden in the dimension config.
Checking for blobs is also something that will change since in 1.18, deepslate blobs will only appear below y0.

gnembon:
technically, you can use the preview datapack with your world to get oveworld generate below 0. My only concern is that if you have a typical flatworld and build stuff on the surface, that would convert all cobble generators to deepslate generators., but the same would happen after update since then the flat worlds have their flats at Y-64, not Y0
Not sure what's the best in this case.

DragonEggBedrockBreaking:
In terms of flat worlds, remember that this is a toggleable feature, so you can just *not enable it* on your flat worlds, avoiding the issue.

replaceitem:
Thr flatworld issue is a good point, we could make the rule accept a number as the max y value too, but maybe thats overcomplicating things

src/main/java/carpet/CarpetSettings.java
```diff
@@ -769,7 +769,7 @@ public static void applyLightBatchSizes()
     public static boolean renewableBlackstone = false;
 
     @Rule(
-            desc = "Lava and water generate deepslate and cobbled deepslate instead below y0",
+            desc = "Lava and water generate deepslate and cobbled deepslate instead below y16",
```
gnembon, line 772:
```suggestion
            desc = "Lava and water generate deepslate and cobbled deepslate instead below Y16",
            extra = "This rule may change Y value to 0 with 1.18",
```

PR #918, opened 2021-06-12
Compact fabric initialization code

altrisi:
Completely subjective, feel free to close it.

I just thought it was using too much space, and decided to convert it into a handle stored in a field instead. Also makes `CarpetServer` not implement those interfaces directly.

### DIFF
--- a/src/main/java/carpet/CarpetServer.java
+++ b/src/main/java/carpet/CarpetServer.java
@@ -1,6 +1,5 @@
 package carpet;
 
-
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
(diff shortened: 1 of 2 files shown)


manyrandomthings:
Why have both ClientModInitializer and DedicatedServerModInitializer both doing the same thing? Couldn't you just make it use ModInitializer which does both?

altrisi:
No, because ModInitializer run order is arbitrary, and extensions must be registered before Carpet. Therefore ModInitializer is left for extensions to hook into Carpet before it starts.

gnembon:
I like this approach, not just of the compactness of the result, but also because CarpetServer class it not a 'mod' itself, but facilitates mod operation. It is cleaner to me for sure.

Issue #919, opened 2021-06-12
Function `store_app_data()` broken in 1.17


julienvincent:
Hi there,

I updated my MC server to 1.17 yesterday and have not been able to run our [cam.sc](https://github.com/gnembon/scarpet/blob/master/programs/survival/cam.sc) (called `c.sc` on our server) script since.

When attempting to run, it fails with

`Unhandled io_exception exception: Unable to write tag to ./world/./scripts/c.data.nbt in c at line 120, pos 4]`

Which is the line `store_app_data(apptag);`. I have also tried to play with it a bit and tested with the `write_file` function and had the same issue. The path in the error message looks malformed and the fact that I get the same issue for both functions makes me think there is some underlying issue in carpet.

Any assistance would be greatly appreciated :)

altrisi:
The `unable to write to...` is just a generic error message Carpet throws when some error occurs in the process. Path looks good to me (`.` means "current directory", so should be practically ignored being in the middle, the first one is basically relative to server jar).

Something may have changed in the vanilla code that is used for saving nbt, will check it in a bit once I fix my dev environment, it isn't currently launching for 1.17.

altrisi:
Hm... So I've tested it, and it works for me in dev... Not sure what's going wrong for you. Details about your system would probably be useful, since when dealing with filesystems, many things change between each other.

julienvincent:
> Path looks good to me ...

Right, yea you are correct the path should be interpreted correctly.

> Details about your system would probably be useful

This is running in a docker container (https://github.com/itzg/docker-minecraft-server specifically) which is based off of ubuntu 20 and running Java 16:

```
root@139b434119f9:/data# cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.2 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.2 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

Perhaps you will have better luck reproducing if you can run the above container. I originally suspected file permissions but AFAIKS the server has full rw access

[...]

PR #920, opened 2021-06-12
A bit more nio and log IO errors

altrisi:
This PR completely gets rid of `File` usages in Scarpet file managing to use nio's Path instead (to hopefully help in more complicated file systems), and also adds logging to IOExceptions that occur in those cases, so it's easier to see what's going on (will help with issues like #919, already resolved that one).


gnembon:
Yeah makes sense to log these.

Suggest using CarpetScriptServer.LOG, not CarpetSettings.LOG for that

altrisi:
Changed.

gnembon:
Thanks :)

PR #956, opened 2021-06-26
Change how resource `type` works, allow proper app and library installation

### LINKED ISSUE #954: [Scarpet] App resource type is unable to load apps
I'm currently trying to make a kind-of installer for the world-edit app to place it on the app store while still keeping it in its own repo.

However, I'm facing a problem: Downloaded resources, even of `app` type, cannot be loaded as actual apps since they are downloaded to either app data or shared app data. That renders the app downloading feature almost useless since it can't ever be used to download neither apps or libraries from wherever, app store or custom url.

Also, in case this functionality gets added to the `app` type only: Is there a way to specify an `app` resource type from a remote url, such as another repository?

Alternative: I could use some ugly `run('script download...')` by setting the app store url carpet rule and similar things, but I'd prefer not to.

altrisi:
Fixes #954.

This PR changes how the `type` argument is used for and allows properly loading apps and libraries when requested resource is of `app` type.

The `type` property is changed to instead of reflecting whether the path is relative, absolute or a full url with (IMO) misleading names, reflect whether you are requesting a regular file or another app/library. In order to determine the type of path, I added a new method (`getFullContentUrl`) that checks whether the path starts with `http(s)://` (then URL), starts with `/` (then absolute path) or none of the previous ones (relative path). In order to keep possible older apps working, `type -> store` is kept but deprecated to create absolute paths from otherwise relative ones. `type` is no longer mandatory.

The loading of apps and libraries checks two extra things that regular files don't: Urls must end with `.sc(l)` and their target must not contain any slash (trying to sandbox them a bit so they don't get elsewhere, tell me if you know a better way). Then a new `StoreNode` is calculated to use for that app in order to search for respective dependencies. It will either be none (app was from full url), or one calculated according to its location. The app will then be downloaded using the same method that gets the initial app.

Once/if this is merged, please merge the updated `resource_app.sc` in gnembon/scarpet#228 too.

~TODO~:

- [x] Check new library system works
- [x] Trash
- [x] Update docs

### DIFF
--- a/docs/scarpet/api/Overview.md
+++ b/docs/scarpet/api/Overview.md
@@ -104,7 +104,7 @@ predicate that is volatile and might change, the command might falsely do or do 
 however player will always be able to type it in and either succeed, or fail, based on their current permissions.
 Custom permission applies to legacy commands with `'legacy_command_type_support'` as well
 as for the custom commands defined with `'commands'`, see below.
-*  `'resources'` - list of all downloadable resources or dependencies when installing the app from an app store. List of resources needs to be 
+*  `'resources'` - list of all downloadable resources when installing the app from an app store. List of resources needs to be 
 in a list and contain of map-like resources descriptors, looking like
    ```
    'resources' -> [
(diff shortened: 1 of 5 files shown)


gnembon:
I like the removal of 'type' since indeed all path cases are clear here. I would just get rid of it completely.
as mentioned in the ticket - don't feel like mixing resources and libraries is a good idea - I would rather add them as a separate key in __config. but before that allow imports from scripts/<app>.libs/ location  and install them there, not to mix app resources.

gnembon:
the 'carpets.app' example was... unfortunate - it didn't mean to install that app alongside, just to download it, showing how you can download other stuff.

gnembon (approved):

PR #958, opened 2021-06-26
Filter script download suggestions according to player input

### LINKED ISSUE #957: [Script download] Suggestions don't get filtered according to characters typed
Unless you type the entire name of the folder (in which case it will show the files inside it, which would have the same problem), the suggestions dropdown will not get filtered to only show the suggestions that start with what the user typed.

altrisi:
Fixes #957.

This PR makes the suggestion builder of the `/script download` command filter the suggestions according to the player's input. The reason it's a bit more complex than just stream & filter is to "open" and suggest directory contents when only a single folder is left after the filtering.


gnembon:
Thanks for applying filtering of results by matching - that was needed. The other part - where it expands when only one matches, don't thing its that necessary, but there was no hint to put a / either, to this is better.

PR #959, opened 2021-06-26
Add missing `remap = false` to mixins to fix build warnings

altrisi:
Removes build log warnings about unresolved references, since those don't actually have to be remapped for various reasons.

Still a draft because it has not yet been tested in production, there's at least one mixin I think may need an explicit `remap = true` inside the `@At`.

About the `UnresolvedMixinReference` SuppressWarnings... No idea. I haven't set up Mixin/whatever gives those in my IDE, so mine just shows unsupported SuppressWarnings for them at any moment.

### DIFF
--- a/src/main/java/carpet/mixins/BiomeArray_cleanLogsMixin.java
+++ b/src/main/java/carpet/mixins/BiomeArray_cleanLogsMixin.java
@@ -12,7 +12,8 @@ public class BiomeArray_cleanLogsMixin
 {
     @Redirect(method = "<init>(Lnet/minecraft/util/collection/IndexedIterable;[I)V", at = @At(
             value = "INVOKE",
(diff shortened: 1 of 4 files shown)


gnembon:
Now, the UnresolvedMixinReference is a red warning issued by Minecraft Development plugin. Adding that makes the whole code appear errorless. 

Adding remap=false has a positive effect that the warning is not raised during compilation, but - these are untracked mixins that may fail silently, but I guess having them remap=false doesn't make it harder to track updates either.

I remember that there were issues with having remap-false, but I can't remember why - I guess we can try.

altrisi:
Then when testing I'll try them as required just to be sure they apply, not sure when will that be though.

altrisi:
This got filled with conflicts after mojmap. Will remake it and open a new PR when I have the time (and will) to redo and fully test them.

PR #986, opened 2021-07-06
Provide script download suggestions async

### LINKED ISSUE #985: [Script download] Loading suggestions async
Currently, the app store and its subfolders are fetched synchronously, halting the game thread. This can be a problem if internet connection is slow. Given the suggestions are not crucial and shouldn't cause problems if being loaded asynchronously, it probably would be a good idea to make them async.

I'll try this one SoonTM.

altrisi:
Resolves #985.

Does basically that.

Haven't tested it with a vanilla client, but given it's a brigadier feature I guess it should work?


altrisi:
Wait a sec, I may have felt into a trap there.

gnembon (approved):

PR #990, opened 2021-07-09
Add web requests to scarpet

Ghoulboy78:
This will add HTTP API scarpet, so allow for access to the broader internet. As happens with most of my prs, it seems, I will end up making a very restricted initial template, which gnembon, altrisi or both will improve ten-thousand-fold (see the state of my in-built draw and distance _beta apps, /script download, scarpet command system... ).

Putting my personal internal resentments aside, here is a comprehensive list of things I wanna add with this pr:

- [ ] Ability to query URLs, and get at the very least a string response. Here, I want to make a handler event of some sort, to account for slower connections, so when you make the html request, you can set a function to trigger when you get the response.
- [ ] One or more throwable errors (the current one is temporary until I figure out exactly what needs to be thrown and how, I just needed a commit so I could make the pr)
- [ ] Make the `http_request` function (not committed yet) return a task value, as opposed to running a function when the response arrives.
- [ ] Allow to send additional information along with the request, i.e some form of metadata
- [ ] All the suggestions and improvements that altrisi and gnembon will undoubtedly make.

### DIFF
--- a/src/main/java/carpet/script/exception/Throwables.java
+++ b/src/main/java/carpet/script/exception/Throwables.java
@@ -28,6 +28,7 @@ public class Throwables {
     public static final Throwables NBT_ERROR             = register("nbt_error", IO_EXCEPTION);
     public static final Throwables JSON_ERROR            = register("json_error", IO_EXCEPTION);
     public static final Throwables B64_ERROR             = register("b64_error", IO_EXCEPTION);
+    public static final Throwables HTML_ERROR            = register("html_error", IO_EXCEPTION);//todo figure out exactly what kind of error it is
     public static final Throwables USER_DEFINED          = register("user_exception", THROWN_EXCEPTION_TYPE);
 
     /**


gnembon:
first of all, why limiting to html. I would suggest to base this on wget / python requests

Second of all - no idea why you opened a PR rather than opened an issue

Ghoulboy78:
Well, to answer the first point: ok, sure. To answer the second, I opened a PR cos I wanted to implement this myself. Unless of course you wanted to implement this, in which case I will of course change to an issue.

Ghoulboy78:
whoops...

altrisi:
Since you asked for comments, here is my full point of view of web requests in Scarpet and this PR.

First of all, you most likely confused HTML with HTTP. HTML is a file type that is the most common one you see in the web since that's what the browser renders to you. HTTP is the protocol, basically the sending of requests and responses. (you may want to change that).

Now my concerns and views on this subject.

This addition could be debatable because it can easily break sandboxing of apps. You are basically handing them all the internet. If implemented poorly (by Carpet or by an app), a player could from making the server call a "bad" url (may leak things like IPs under a proxy) to tricking the app into downloading potentially unwanted files that could get flagged as provided by Carpet. Is this really an issue? Well, probably not. It should be the programmer's responsibility to use it with caution.

Now, complexity. You'd have to think what you want to support. The most common request type is `GET`, which is, in simple terms, just having an url and making a request to it. However, you also have `POST` (like GET but with more params), `PUT`, `DELETE`, `PATCH`. Would those be supported? Including sending whole files to an URL?

Not only that, given that you are handing this powers to an app that is most likely going to use this to access APIs, so you'll most likely need to provide methods to attach headers into requests in order to be able to attach things like authentication/API tokens, special request parameters, etc.

Let's get into performance and blocking the game thread... What to do with that? I can think of multiple ways to deal with that, since it'll be necessary since you don't really want a long web request (for example, 30s) to stall your game's thread (potentially triggering watchdog if you are not careful). One possibility would be limiting requests to not run on the server thread (throw IEE if that's the case) but be on a task (value), however this makes it more difficult to implement from Scarpet code since tasks is not really a heavily used feature. Another way would be the request actually returning a task, making it more flexible to work with, since if it's a critical request you can immediately join it, but we still hit a bit of complexity given we are still using tasks (plus maybe you could even cancel it). Last but not least, you could add a callback parameter to the request function that would be called when the request finishes (probably docking it to the main thread/caller's thread), however that would remove the flexibility of it given you can't really control what happens with it, plus there's no way at all to force a critical request to block the execution of the game.

About the responses... A string response is probably good enough. There's no need for anything else, there are built-in methods to parse things like JSON, and the programmer will have to implement their methods for other encodings anyway.

Ghoulboy78:
Thank you very much for your comment. I was an idiot, and mixed up HTTP with HTML, which is dumb of me, sorry.

With regards to downloading potentially dangerous files, I had not actually planned on allowing to download files actually, only allowing string responses, and at most copying what has been recently added with `resources` parameter in `__config`. In that sense, I meant to add a function like `http_request(url,function)`, where you would give it a url, and then when the result came back it would run the function in question, with the first argument being the string response.

As to the time taken for returning results, I mentioned that I wanted to send the request, and then when it came back run the function, cos I hadn't thought of task values (having never used them before), so this should be a good opportunity for me to learn how they work and how to use them.

I also planned on potentially restricting the number of requests to like a few per tick or something, to avoid spam.

altrisi:
Tbh, I'd not limit it, just task/callback them, and I'd personally like being able to at least use `POST` other than `GET` (it's commonly used for submitting forms or data, could allow sending stats to your website or similar) and setting headers (for example Github API is extremely limited without a token, and you have to put the token in a header, same for a lot of other APIs), that's why this is a complicated subject.

Actually hadn't fully thought about downloading files, but it wouldn't be that bad to allow downloading files as long as they only get into the app's sandboxed space I guess (there's a function in one of MC's libraries, `FileUtils.copyURLToFile(URL, File, [int, int])` (for `GET`) that downloads a file in one line, plus there are some quite performant ways using Java's nio). About uploading, not really sure how that could work, if it would work, or if would really be useful, given the limited amount of files Scarpet can produce.

gnembon:
Technically each app from the appstore is capable of fetching any arbitrary web resource, but its all limited to downloading. 

Consideration about 'preventing downloading dangerous files' vs 'just fetching string responses' is that they are equivalent (pretty much). The only way to combat it is the fact that apps are scripts, so open source, and at least submitting them to the default appstore is warranted with some community supervision. As an admin, what you do with your scripts is up to you.

In terms of considerations for blocking aspects of requests - seeing no problem with that. The simplest the API the better, otherwise one would unnecessarily complicate this stuff. If you need to lazy fetch those making a task is the proper way to do it. I guess if you know how to handle parallel tasks you would understand scarpet tasks enough to do that.

In terms of a good model, pythons 'requests' library is a good example of not overcomplicating things. You would have as input (uri, method, timeout?, params?, headers?, body?) or (uri, method, options?) where options is a map with all optional requests keys, like params, headers, body, timeout etc) and with a response {`content` -> ..., `status` -> code/timout,etc , `headers` -> ...} and that's it? I woudn't bother with encoding / streaming at this point. @Ghoulboy78 before you comment further on that topic I recommend you take a look at that classic library.

Another thing that would be complex to provide is cookies and sessions, Don't think that's needed

src/main/java/carpet/script/exception/Throwables.java
```diff
@@ -28,6 +28,7 @@
     public static final Throwables NBT_ERROR             = register("nbt_error", IO_EXCEPTION);
     public static final Throwables JSON_ERROR            = register("json_error", IO_EXCEPTION);
     public static final Throwables B64_ERROR             = register("b64_error", IO_EXCEPTION);
+    public static final Throwables HTML_ERROR            = register("html_error", IO_EXCEPTION);//todo figure out exactly what kind of error it is
```
gnembon, line 31: That's not even needed - failed response could just be returned in the status.

gnembon:
Waist, why a PR?

That should be just a submitted issue

Ghoulboy78:
ok

PR #998, opened 2021-07-10
Add condition argument to Rule annotation

### LINKED ISSUE #997: Add optional conditions for Rules, when to apply
The idea is a new optional argument for `@Rule`, to specify a condition, which has to be true in order to apply the rule.

That would make things like MultiCoreNetwork/carpet-redcraft-addons#78 / MultiCoreNetwork/carpet-redcraft-addons#80 possible, without needing to mixin into a Carpet class.

RubixDev:
Closes #997

### DIFF
--- a/src/main/java/carpet/settings/Condition.java
+++ b/src/main/java/carpet/settings/Condition.java
@@ -1,0 +1,5 @@
+package carpet.settings;
+
+public interface Condition {
+    boolean isTrue();
+}
--- a/src/main/java/carpet/settings/Rule.java
+++ b/src/main/java/carpet/settings/Rule.java
@@ -64,4 +64,9 @@ public @interface Rule
      * The class of the validator checked when the rule is changed.
      */
     Class<? extends Validator>[] validate() default {};
+
+    /**
+     * The class of the condition checked when the rule is parsed.
+     */
+    Class<? extends Condition>[] condition() default {};
 }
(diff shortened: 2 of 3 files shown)


gnembon:
isn't that what validator is for?

consider: 
```
private static class ConditionValidator extends Validator<Boolean>
    {
        @Override public Boolean validate(ServerCommandSource source, ParsedRule<Boolean> currentRule, Boolean newValue, String val) {
            return 2 + 2 < 5 ? null : newValue;
        }

        @Override
        public void notifyFailure(ServerCommandSource source, ParsedRule<Boolean> currentRule, String providedValue) {
            Messenger.m(source, "gi Function disabled");
        }
    }
    @Rule(desc = "Placing carpets may issue carpet commands for non-op players", category = SURVIVAL, validate = ConditionValidator.class)
    public static boolean carpets = false;
```
makes rule not-switchable

gnembon:
fyi - I can see why different behaviours for rule not able to set would make sense here so I allowed to customize it in validators.

gnembon:
https://github.com/gnembon/fabric-carpet/commit/7421e566705aca76ecdbbb235813ec7f9defb075

RubixDev:
The difference between the validator and the condition I implemented is when that check is executed. At least to my knowledge with validators, that always return `null`, the rule still "exists", just isn't able to be changed. In the case I mentioned in the issue, where 2 extensions use the same rule name, the last loaded rule would still overwrite the first one. With my condition checking while *parsing* the rules, it doesn't even "exist" when the condition is false.

I hope that makes sense and I don't just overlook something

gnembon:
Ok, I get that. Makes sense.

PR #1006, opened 2021-07-16
Update Math.md -- Fixed Typo

Mythbusters123:
Fix Absolut Value to Absolute Value.

### DIFF
--- a/docs/scarpet/language/Math.md
+++ b/docs/scarpet/language/Math.md
@@ -14,7 +14,7 @@ Square root. For other fancy roots, use `^`, math and yo noggin. Imagine square 
 
 ### `abs(n)`
 
-Absolut value.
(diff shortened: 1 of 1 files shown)


gnembon:
Is this a typo? https://en.wikipedia.org/wiki/Absolut_Vodka

Mythbusters123:
Sorry, just me being a stupid american, although it does bother me. Idk you chose!

Ghoulboy78:
~~Well at least they admit that Americans are stupid...~~

Also the brand is a Svedish brand, and `absolut` is a Svedish word meaning `absolute` or `absolutely` in English. Thus, you have actually snuck a Svedish vord into de scarpet docs, and I actually agree vid de American on this.

PR #1009, opened 2021-07-17
Use newer Java features in places where it makes sense and more

altrisi:
So I did some upgrading off camera...

This PR (ready):

- Changes all tuples (pair and triple) I knew how to call their parameters to use records instead (the rest are still there)
- Changes all references to guava ImmutableList and ImmutableMap to use Java's new List.of() and Map.of() that also return immutable collections (also Set)
- Upgrades some classes that were candidate to be a record to, well, a record
- Removes deprecated `NumericValue(boolean)` constructor given it was already deprecated and the only extension that used it hasn't used since early 1.17 snapshots
- Also adds `@Override` to non-mixin classes, makes it easier to find that a method implements something. Not understanding why a `divide` method in ListValue and discovering it's an override later can be frustrating.

Not in this PR (Coming Soon™ in different PRs):

- #1019
- Finding and upgrading more of the double `{{` initializations to Map.of(), etc
- Maybe making use of abstract classes in mixins to be able to use the more compatible (relatively-recently upgraded in Mixin) Accessor and Invoker with the current fakes and that can live together with other Mixin code to make them simpler
- Maybe adding `@Override` to mixins too
- Maybe porting some Scarpet functions to annotations
- Maybe some `var`s, though those are currently remapped to their type on migrateMappings until Mercury is updated and gets bumped in Loom

Take your time to review it.

NOTE: At the time of sending this PR (and now), it has got almost no testing.

NOTE 2: In those nested records that previously accessed public final fields, I haven't changed the calls to their functional versions, since it's still possible from the same class (though I _think_ it produces bridge methods), tell me if you want me to change that.

### DIFF
--- a/src/main/java/carpet/helpers/HopperCounter.java
+++ b/src/main/java/carpet/helpers/HopperCounter.java
@@ -5,8 +5,6 @@ import carpet.fakes.IngredientInterface;
 import carpet.fakes.RecipeManagerInterface;
 import carpet.utils.WoolTool;
 import carpet.utils.Messenger;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Maps;
 import it.unimi.dsi.fastutil.objects.Object2LongLinkedOpenHashMap;
 import it.unimi.dsi.fastutil.objects.Object2LongMap;
 import net.minecraft.block.AbstractBannerBlock;
@@ -42,6 +40,8 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import static java.util.Map.entry;
+
 /**
  * The actual object residing in each hopper counter which makes them count the items and saves them. There is one for each
  * colour in MC.
@@ -67,7 +67,7 @@ public class HopperCounter
         {
             counterMap.put(color, new HopperCounter(color));
         }
-        COUNTERS = Maps.immutableEnumMap(counterMap);
+        COUNTERS = Collections.unmodifiableMap(counterMap);
     }
 
     /**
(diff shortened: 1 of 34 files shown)


gnembon:
Got a Butt for you. It all makes sense, but core mc code doesn't have any of these. Having said that - I think its a good idea. Also - the the 'future' branch being only a few commits ahead, that's the best / only time for refactors of any kind. Will merge later.

altrisi:
Fixed conflicts with latest commits.

src/main/java/carpet/helpers/HopperCounter.java
```diff
@@ -67,7 +67,7 @@
         {
             counterMap.put(color, new HopperCounter(color));
         }
-        COUNTERS = Maps.immutableEnumMap(counterMap);
+        COUNTERS = Collections.unmodifiableMap(counterMap);
```
gnembon, line 70: Isn't enum map faster or the collections method is smart enough to figure it out?

altrisi: `Collections.unmodifiableMap` is the same as `Maps.immutableEnumMap`: Just a wrapper over the other map. Since it's already an EnumMap, it should have the same performance as the guava version, because they both do the same.

Only difference is that guava's method creates a copy of the map instead of saving the same one, and has a custom class for it. Since the old map is not passed anywhere, there's no problem with that.

src/main/java/carpet/script/utils/SystemInfo.java
```diff
@@ -133,7 +133,7 @@
         put("java_system_cpu_load", c -> {
             OperatingSystemMXBean osBean = ManagementFactory.getPlatformMXBean(
                     OperatingSystemMXBean.class);
-            return new NumericValue(osBean.getSystemCpuLoad());
+            return new NumericValue(osBean.getCpuLoad());
```
gnembon, line 136: What changed here?

altrisi: The old one is deprecated because of "being a historically named method". The old one calls the new one so they do the same.

gnembon (approved):

PR #1015, opened 2021-07-18
Fix numeric suggestions in command arguments

### LINKED ISSUE #961: Integer suggestions fail to follow argument limits
When an arguments for a command is of type `int` and only has `min` and `max` provided, the automatically generated suggestions that appear when using the command do not follow these limits.
To replicate:
1) make a command with argument of type `int`, `min` of 1,`max` of 500
2) load into Minecraft and start typing the command, leaving this argument blank

Expected behavior:
3) default suggestions of -123,0,123 are replaced with just 123, the only valid integer of the three

Actual behavior:
3) all default suggestions of -123,0,123 appear, even though two of them are not valid

Ghoulboy78:
Fixes #961 
It turns out I didn't need to implement what happens when max is defined and not min, as apparently that's not even allowed, so less work for me

### DIFF
--- a/src/main/java/carpet/script/command/CommandArgument.java
+++ b/src/main/java/carpet/script/command/CommandArgument.java
@@ -4,7 +4,6 @@ import carpet.CarpetServer;
 import carpet.fakes.BlockStateArgumentInterface;
 import carpet.script.CarpetScriptHost;
 import carpet.script.argument.FunctionArgument;
-import carpet.script.exception.InternalExpressionException;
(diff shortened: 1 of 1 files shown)


Ghoulboy78:
Ok, I have added it, also making sure that I respect the 'suggest' parameter (which in the first draft I didn't), as well as making sure that with regards to int arguments, there be dealing with `long` type arguments, so that when converting to strings for the suggestions they lose the `.0` suffix which got added accidentally.

PR #1019, opened 2021-07-21
Use pattern matching for instanceof

altrisi:
Another newer Java feature.

I've checked every non-mixin class with `instanceof` and changed those that immediately assigned a variable with a cast, other than those (one) that I thought hurt readability. Didn't touch those that only cast the existing variable later to pass it/invoke something on it, since that would add an extra field allocation etc.

Lightly tested, but shouldn't break.


gnembon:
oh, that's neat, syntactic sugah

gnembon (approved):

PR #1020, opened 2021-07-21
Add `commandBlockPermissionLevel` rule

BeetMacol:
[Showcase video](https://streamable.com/1etv8b)

### New rule: `commandBlockPermissionLevel`
The rule lets you change the permission level command blocks will use when executing commands. In vanilla it's a constant set to `2`.

This lets you allow command blocks to kick players, manage whitelists or even save or stop the server.

### New rule property: `allowCommandBlocks`
This also adds a new property to the `@Rule` class - `allowCommandBlocks`. It stops command blocks from changing the value if set to false. `commandBlockPermissionLevel` uses it to stop command blocks from granting themselves bigger permissions.


altrisi:
A few small things:
- You sorted the mixins file!
- Instead of adding a new property to rules that is quite likely only going to be used for this, I'd recommend checking the permission level of the source changing the rule to be equals or greater to the one requested, just like `commandScriptACE` does (therefore in a `Validator`). This way you also prevent an op with level 2 from allowing command blocks level 4 and then doing `/stop` in a command block (and save a mixin).
- Also consider using the default `_COMMAND_LEVEL_VALIDATOR` that allows strings like `ops`, or a custom implementation to not allow booleans since they seem pointless in here

BeetMacol:
Yeah, I realized `allowCommandBlocks` wasn't a good idea, and that's why I marked the pr a draft. The way I did it would allow op players of level 2 to change the command block permission level to anything and that wouldn't be good.

gnembon:
Agree with alrisi, and yourself - `allowCommandBlocks` is not really needed. `commandBlockPermissionLevel` might be useful on the other hand, and setting to use them as altrisi suggested is probably the best idea.

altrisi:
Any updates on this?

BeetMacol:
Nah, I don't think I will be finishing that, it's outdated anyways and isn't really that useful. I will not find time to finish this PR. Sorry for leaving it open for so long.

PR #1022, opened 2021-07-22
added more customizability to renewableCoral

HED-GIT:
allows you to switch between disabled, coral_only, fan_only and enabled-mode to allow you to also disable/enable growing corals from coral-fans to make them easier to farm.

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -768,7 +768,7 @@ public class CarpetSettings
     }
     @Rule(
             desc = "Coral structures will grow with bonemeal from coral plants",
-            extra = "Expanded also allows growing from coral fans for easier farming",
+            extra = "Expanded also allows growing from coral fans for sustainable farming outside of warm oceans",
             category = FEATURE
     )
     public static RenewableCoralMode renewableCoral = RenewableCoralMode.FALSE;
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -18,9 +18,7 @@ import net.minecraft.server.dedicated.DedicatedServer;
 import net.minecraft.server.world.ServerWorld;
 import net.minecraft.util.Identifier;
 import net.minecraft.util.math.BlockPos;
-import net.minecraft.util.math.Box;
 import net.minecraft.util.math.ChunkPos;
-import net.minecraft.util.math.Vec3d;
 import net.minecraft.util.registry.Registry;
 import net.minecraft.world.World;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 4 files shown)


altrisi:
Please don't make String comparisons for rule checks in-game, they aren't cheap (in relation to equality of booleans/etc). Instead I'd  suggest you either make changing the rule change other boolean/int fields and compare from that or use an Enum rule type where you can check equality directly in-game without much cost while keeping readability.

Also suggesting to use `true`/`false` as enabled/disabled to both keep existing enabled rules working and be more consistent with the rest of rules.

HED-GIT:
i changed it to use enums and use true/false

i hope it is preferred that the enum definition is directly part of the CarpetSettings class, else i can of course move it to whatever location you think will fit best

gnembon:
Practical questions, or UX questions if you will - what is the benefit of allowing growing corals from fans comparing just from small coral plants, and second question - what's the benefit of adding extra mental complexity to the rule to provide that fine grained control, rather than allowing on both coral plant types, i.e. what is the practical use of enabling one and disabling another.

HED-GIT:
I primarily changed do to me wanting to be able to farm coral-blocks even outside of warm-oceans. With growing them only from coral-plants you run out of them pretty fast, since they don't generate on every coral, which forces you back to the warm-oceans to continue farming.

If you can farm from both coral types you can farm into infinity just by growing corals and collecting the plants and fans.

The question is basically how difficult the user wants the farming to be, i wanted it to be easier to farm so i changed it for me and made the pull request in case someone else also wants it that way, 

But yes fan-only mode is pretty much useless.

Better solution might be:
- FALSE -> disabled
- TRUE -> like in the current version, only from coral-plants
- EXPANDED (or some other name) -> also allowing them to grow from fans

If you prefer it that way then i can remove the fan-only mode.

gnembon:
Ok, that's a good answer - if you only rely on growing corals to get 'saplings', with small corals that's not enough to keep the loop going. That makes sense. I think 'true', 'false' and 'extended' with the proper description (and the reasoning you provided) should be good. If you don't say why there is 'extended' option it would be difficult to understand the 'why'.

HED-GIT:
Okay i removed the fan_only and renamed the other options

the explanation was added as an "extra" for the rule

src/main/java/carpet/CarpetSettings.java
```diff
@@ -762,9 +760,18 @@ public static void applyLightBatchSizes()
...
-    public static boolean renewableCoral = false;
+
+    public enum RenewableCoralMode {
+        FALSE,
+        EXPANDED,
+        TRUE;
+    }
+    @Rule(
+            desc = "Coral structures will grow with bonemeal from coral plants",
+            extra = "Expanded also allows growing from coral fans for easier farming",
```
gnembon, line 771:
```suggestion
            extra = "Expanded also allows growing from coral fans for sustainable farming outside of warm oceans",
```

PR #1030, opened 2021-07-26
Remove some no longer needed Mixins

altrisi:
Haven't checked in-depth, but found a few that were no longer needed because of changes in the visibility in the original classes, or just because their rules were modified, and decided to change them.


gnembon:
thanks :)

gnembon (approved):

PR #1038, opened 2021-08-01
Max collisions optimization

### LINKED ISSUE #1034: `maxEntityCollisions` is labeled as a "optimization", but does not actually improve TPS significantly
### The problem:
In the course of my analysis of tick cramming lag for Lithium (https://github.com/CaffeineMC/lithium-fabric/pull/260), I found this Carpet setting which at first glance seemed to limit the `n^2` nature of overlapping collision checks to a mere `k*n` where `k` is the current setting and `n` is the number of overlapping entities. Unfortunately, this is actually not the case, as I will explain.

### The background:
The mixin that implements this feature ([here](https://github.com/gnembon/fabric-carpet/blame/master/src/main/java/carpet/mixins/LivingEntity_maxCollisionsMixin.java)) does prevent collision code from running after `k` entities have been processed. However, the code it prevents from running isn't actually the code that causes the most lag. This] is actually the code that causes the lag:
https://github.com/gnembon/fabric-carpet/blob/2a391962ff7ebed382289f278959355c6f43785d/src/main/java/carpet/mixins/LivingEntity_maxCollisionsMixin.java#L33-L33

In my PR to Lithium, I identified three sources of lag: 1) the bounding box checks for every entity in the chunk section, 2) every collision check getting the entity's health from the `DataTracker`, and 3) every collision check getting the climb-ability of the block the entity is in. The crucial part about this is that all these expensive checks are done at the very beginning of the method when it queries for a list of all matching colliding entities. This means that the mixin doesn't affect the number of checks, and lag isn't reduced.

### The solution:
The mixin should instead limit the initial search of matching colliding entities to the current setting. This is not possible with the current `getOtherEntities` method, so a new helper method should be created that bails out of the search after reaching the specified limit.
@gnembon, I'm able to get started on a PR for this right away. Just let me know if you have any comments, concerns, or if I should just get started.

magneticflux-:
Closes #1034

### DIFF
--- a/src/main/java/carpet/fakes/WorldInterface.java
+++ b/src/main/java/carpet/fakes/WorldInterface.java
@@ -5,12 +5,18 @@ import net.minecraft.block.entity.BlockEntity;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.EntityType;
 import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.Box;
+import org.jetbrains.annotations.Nullable;
 
+import java.util.List;
 import java.util.Map;
+import java.util.function.Predicate;
 
 public interface WorldInterface
 {
     Map<EntityType, Entity> getPrecookedMobs();
     
     boolean setBlockStateWithBlockEntity(BlockPos blockPos, BlockState blockState, BlockEntity newBlockEntity, int int1);
+
+    List<Entity> getOtherEntitiesLimited(@Nullable Entity except, Box box, Predicate<? super Entity> predicate, int limit);
 }
--- a/src/main/java/carpet/fakes/WorldInterface.java
+++ b/src/main/java/carpet/fakes/WorldInterface.java
@@ -14,7 +14,7 @@ import java.util.function.Predicate;
 
 public interface WorldInterface
 {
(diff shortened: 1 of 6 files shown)


magneticflux-:
If you don't want the small changes in the other commits, I can rebase and skip everything but d85294a.

James103:
Have you ran some tests with this optimization? If so, can you please post the results of the testing?

magneticflux-:
`/carpet maxEntityCollisions 0`
image


`/carpet maxEntityCollisions 20`
image

gnembon:
Thanks, this is exactly what it needed to be. One question - would replacing throwing an exception with just skip be more compute effective? Exceptions do cost a lot of CPU money, and typically these queries do not run through a lot of containers.

magneticflux-:
I believe throwing and catching an exception is the only way to break out of the `forEachIntersects` call without rewriting 3 or 4 more methods to accept `limit` parameters. A simple `return` inside `forEachIntersects` would skip the entity adding code, but wouldn't skip the more expensive bounding box checks that make up most of the execution time.

gnembon:
true

gnembon (approved):

PR #1063, opened 2021-08-16
Stupid fast build script and smaller jars

MaxBr5:


altrisi:
Hmmmm no?

You are literally removing all the release configuration, and not having debug info is like the worst idea I've ever heard (that's what -g none does). Like why would you even care about the build speed, even less if you loose any debug info?

gnembon:
detailed logs in builds are actually quite useful - I wouldn't touch this part.

PR #1083, opened 2021-08-31
Make ErrorSnooper report values for locals in the stacktrace

### LINKED ISSUE #1051: [Scarpet Debugging] Display local values when hovering over references
Shouldn't be that difficult?

Given those are definitely available at exception time, because `try()` can access them in `_trace`, may be a matter of making some String indexOf in the displayed snippet with the available locals and then separating the line so the hover text can be added.

I think it could be quite useful. May try it if I have time.

Edit: Will require passing the `Context` to the snooper. Shouldn't be a problem. Maybe even just passing the exception to it and let it handle everything?

altrisi:
Resolves #1051.

This PR adds a hover over tooltip for references to locals in the exception's stacktrace. It does so by trying to find the variables in the Strings directly, so there can be false positives (though _many_ of those that cause actual variables to not be available should be handled by the logic).

Things I'm not sure about: Should ErrorSnooper instead ask for the entire `ExpressionException`, since it uses most of it anyway? Should the color of the tooltip be changed to gray (or some other color)? Should the `evalValue` for locals be inside a try-catch in case there's a possibility for another exception to be thrown in there?

### DIFF
--- a/src/main/java/carpet/script/CarpetScriptHost.java
+++ b/src/main/java/carpet/script/CarpetScriptHost.java
@@ -43,6 +43,7 @@ import net.minecraft.nbt.NbtElement;
 import net.minecraft.server.command.ServerCommandSource;
 import net.minecraft.server.network.ServerPlayerEntity;
 import net.minecraft.server.world.ServerWorld;
+import net.minecraft.text.BaseText;
 import net.minecraft.text.LiteralText;
 import net.minecraft.text.Text;
 import net.minecraft.util.math.BlockPos;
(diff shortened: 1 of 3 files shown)


gnembon (approved):

gnembon:
Looks mighty fine. 

Regarding your questions. I guess you could pass ExpressionException. Color is up to you, use best judgement. Regrading try/catch definitely bacause there might be some infinite recursive references inside application locals, so they cannot be fully trusted with.

altrisi:
Wrapped it into a try-catch for StackOverflows (since a more faulty lv shouldn't have been able to get into a variable since those shouldn't do nothing on eval other than returning a value, right?), left the rest unchanged since I think it's ok and I don't know how to choose colors, and gray may be more difficult to read than white.

Should be ready then.

gnembon (approved):

PR #1087, opened 2021-09-02
Use `player.interactionManager.tryBreakBlock` instead of a copy to be more compatible

### LINKED ISSUE #1086: Scarpet `harvest` bypassing Fabric API Player Block Break Event
The scarpet harvest function uses a copy of [`tryBlockBreak`](https://github.com/gnembon/fabric-carpet/blob/4dd3cc7fd62865d401a8900aaee69ba524cf644e/src/main/java/carpet/script/api/WorldAccess.java#L222) from `PlayerInteractionManager` instead of invoking it directly which leads to Fabric API Player Block break event being bypassed along with mixins with a similar goal.

altrisi:
Resolves #1086.

The interaction manager and the method is (at least now) public, and I compared them and they were the same, so I assumed it would be safe to just use it directly, so also if in an update the method is changed, there is no discrepancy with vanilla.

Should I keep the comment and move it right above the line where the method is called?

### DIFF
--- a/src/main/java/carpet/script/api/WorldAccess.java
+++ b/src/main/java/carpet/script/api/WorldAccess.java
@@ -219,46 +219,6 @@ public class WorldAccess {
         return bs;
     }
 
-    private static boolean tryBreakBlock_copy_from_ServerPlayerInteractionManager(ServerPlayerEntity player, BlockPos blockPos_1)
-    {
-        //this could be done little better, by hooking up event handling not in try_break_block but wherever its called
-        // so we can safely call it here
-        // but that would do for now.
-        BlockState blockState_1 = player.world.getBlockState(blockPos_1);
-        if (!player.getMainHandStack().getItem().canMine(blockState_1, player.world, blockPos_1, player)) {
-            return false;
-        } else {
-            BlockEntity blockEntity_1 = player.world.getBlockEntity(blockPos_1);
-            Block block_1 = blockState_1.getBlock();
-            if ((block_1 instanceof CommandBlock || block_1 instanceof StructureBlock || block_1 instanceof JigsawBlock) && !player.isCreativeLevelTwoOp()) {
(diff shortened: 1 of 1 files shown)


gnembon:
problem is that it would trigger all events, including scarpet's breaking block.

See the vein miner example.

gnembon:
from the private discussion with the author of the ticket:
```
simple reason is that this was also causing all scarpet events to trigger as well, which wasn't intended. Think about vein miner application where its triggered by a player breaking a block and causing other blocks to break without triggering that event recursively. Other option was to use some sort of threadlocal global, but that would complicate stuff down the road. Feel free to submit that as an issue on the bugtracker and we can give it a thought for sure. Argument about skipping  FAPI events is valid in this case.
```

altrisi:
Oh true, I completely forgot mixins may have been applied to the method, including Carpet's.

Then I see two possibilities: Either injecting an optional dependency with the FAPI module with that and call the event from there, or a boolean somewhere to skip the Scarpet event temporarily, not sure if it there's risk for multiple threads accessing that (haven't checked the code).

gnembon:
threadlocal boolean is a better idea not only because of that, but also for further use in cases like handling chat messages or potentially other scarpet APIs that could cause some other events to trigger.

This mechanism could potentially be more generalized - when handling an event - other events are disabled until it is handled, but with the veinminer example - this could happen later via 'schedules' so it makes sense its disabled for 'harvest'

gnembon:
I could do that, but if you want to do it that would be great. For reference: https://github.com/gnembon/fabric-carpet/blob/19c1924a269278eee5274a2036a1c0380c20d62d/src/main/java/carpet/script/api/WorldAccess.java#L704

gnembon:
https://github.com/gnembon/fabric-carpet/commit/229d0b037bd00a5c7dc1be50dadbdce6b95ab7d8
this should actually enable your PR as is.

gnembon (approved):

PR #1090, opened 2021-09-03
Initial work on a Mixin error handler

altrisi:
This is thought to handle mixin conflicts of _specific_ (not important) features and instead of crashing the game, printing a warning and disabling the rule from being changed (not sure how to handle `carpet.conf` having it already set yet).

Why? Because Carpet (and extensions, official or community) have many, many rules, and most people will not use them all. This allows them to play with some Carpet features and mods that conflict with the features that they don't use.

Current progress is the actual errorhandler, an annotation to specify the rule to disable and the entrypoint to set it up.

What's missing is actually disabling the rule, absolutely any testing, and any other TODOs I've left in the code. Maybe also a MixinConfig to stop applying extra mixins for a disabled feature, but not sure if worth it (feels unnecessary as long as all mixin'd features have proper checks against rules). Also now that I think about it this should be disabled in a dev environment.

Also moves FabricAPIHooks into a new `compat` package with this.

What do you think about the concept? Is it a good idea or should I leave it? (ignore the implementation for now, most of it could probably be better, mostly the access of some Mixin internals could be replaced with its API and some asm, but I don't really know ASM atm, so it's for convenience).

### DIFF
--- a/src/main/java/carpet/CarpetServer.java
+++ b/src/main/java/carpet/CarpetServer.java
@@ -17,6 +17,8 @@ import carpet.commands.ScriptCommand;
 import carpet.commands.SpawnCommand;
 import carpet.commands.TestCommand;
 import carpet.commands.TickCommand;
+import carpet.compat.CarpetMixinErrorHandler;
+import carpet.compat.FabricAPIHooks;
 import carpet.network.ServerNetworkHandler;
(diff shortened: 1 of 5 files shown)


altrisi:
Closing this for now, may come back to it if it's actually needed at some point.

PR #1094, opened 2021-09-05
Add bitwise operations to scarpet

Ghoulboy78:
This PR adds the following functions:
- `bitwise_and` function
- `bitwise_xor` function
- `bitwise_or` function
I had previously added operators, but got rid of them cos xor didnt have one

- `bitwise_shift_left` function
- `bitwise_shift_right` function

- `bitwise_roll_left` function
- `bitwise_roll_right` function
These last two do not exist natively in Java, so I took about half an hour to make a fast implementation in python, which I then moved over to java:
```java
long rollLeft(long number, long amount) {
    amount = amount % 64 //cos otherwise weird stuff could happen
    long amountToRoll = 64 - amount;
    long rolledBits = ((-1L) >> amountToRoll) << amountToRoll;
    long rolledAmount = (num & rolledBits) >> amountToRoll;

    return num << amount | rolledAmount;
}
```
Just switch the `<<` for `>>` and viceversa to make `rollRight` function.
I don't need to throw an error for negative rolling, cos that will be caught when you try to bit-shift by that much. (todo test that tho)

Still left to do:
- [x] Docs
- [x] Add `bitwise_` prefix to all functions
- [x] Ditch operators
- [x] add `bitwise_not`
- [x] add`bitwise_popcount`
- [x] Make shifts and rolls only accept 2 args
- [x] Proper testing (i.e testing if it all actually works). 
I kinda need help for this, cos for some reason my `Minecraft Development` plugin in IDEA seems broken, so it won't recognize `net.minecraft` and `com.mojang` packages. It used to give me red lines, then I downgraded IDEA and now it won't recognize these as classes, which is why one of the previous commits failed.


gnembon:
good idea.
few feedback elements
 - ditch all operators - not all are available, plus these are unique operations, functors are enough for that.
 - a few important ops are missing: bitwise_not, bitwise_popcount
 - for consistency could rename all to have a `bitwise_` prefix, so `bitwise_shift_left` for example
 - not sure shift and roll ops need to accept more than 2 args

Ghoulboy78:
After consulting with gnembon, we have decided to get rid of the operators as symbols, and only keep the functions.

Ghoulboy78:
Ah, finally no errors!
(also, I rly need ot fix my gradle...)

Ghoulboy78:
I have now added docs and `double_to_long_bits` and `long_to_double_bits` functions.

Ghoulboy78:
Well it's all done now, so ig this is ready for review now!

src/main/java/carpet/script/language/Operators.java
```diff
@@ -29,12 +29,12 @@
...
-        put("or||", 4);
-        put("assign=<>", 3);
-        put("def->", 2);
-        put("nextop;", 1);
+        put("equal==!=", 8);
+        put("and&&", 4);
+        put("or||", 3);
+        put("assign=<>", 2);
+        put("def->", 1);
+        put("nextop;", 0);
```
altrisi, line 37: If the operators aren't included in the end, shouldn't this be reverted?

Ghoulboy78: oh yeah...

src/main/java/carpet/script/language/Operators.java
```diff
@@ -202,6 +226,46 @@ public static void apply(Expression expression)
         });
         expression.addFunctionalEquivalence("<=", "nondecreasing");
 
+        expression.addMathematicalBinaryIntFunction("bitwise_shift_left", (l1, l2) -> {
```
gnembon, line 229: just use names (num, amount)

src/main/java/carpet/script/language/Operators.java
```diff
@@ -84,6 +84,30 @@ public static void apply(Expression expression)
...
+            long accumulator = NumericValue.asNumber(lv.get(0)).getLong();
+            for (Value v: lv.subList(1, size)) accumulator = accumulator & NumericValue.asNumber(v).getLong();
+            return new NumericValue(accumulator);
+        });
+
+        expression.addFunction("bitwise_xor", lv -> {
+            int size = lv.size();
+            if (size == 0) return Value.NULL;
+            long accumulator = NumericValue.asNumber(lv.get(0)).getLong();
+            for (Value v: lv.subList(1, size)) accumulator = accumulator | NumericValue.asNumber(v).getLong();
```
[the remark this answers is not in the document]
Ghoulboy78: oop

src/main/java/carpet/script/language/Operators.java
Ghoulboy78: k, but cant do that for rolling, cos I need to do mod 64

gnembon:
forgot to add the function names to N++ and Idea's keyword list

Ghoulboy78:
oh ye

gnembon:
You didn't add all methods to the keywords lists, like `long_to_double_bits`

Ghoulboy78:
ok, should be done now

gnembon (approved):

PR #1096, opened 2021-09-07
Add on_move Entity Event for Scarpet

lntricate1:
This is just the necessary code to add the on_move entity event, which is triggered at the end of the setPos method in the Entity class, and it gives you the velocity as well as the starting and ending positions of the movement.

This is necessary for some cases in which it's impossible to access the movement of an entity from scarpet, such as with entities that spawn in with velocity, like ender pearls, and never tick at the first location, and with entities that die while moving, like TNT, that never tick at the end location.

It is used as `entity_event(e, 'on_move', _(e, m, p1, p2) -> ())`, where e is the entity, m is the motion of the entity while it moves, p1 is the position where it starts the movement, and p2 is the position where it finishes.

### DIFF
--- a/docs/scarpet/api/Entities.md
+++ b/docs/scarpet/api/Entities.md
@@ -801,6 +801,7 @@ defining the callback with `entity_event`.
 The following events can be handled by entities:
 
 *   `'on_tick'`: executes every tick right before the entity is ticked in the game. Required arguments: `entity`
+*   `'on_move'`: executes every time an entity changes position, at the end of the `setPos()` method. Required arguments: `entity, velocity, pos1, pos2`
 *   `'on_death'`: executes once when a living entity dies. Required arguments: `entity, reason`
 *   `'on_removed'`: execute once when an entity is removed. Required arguments: `entity`
 *   `'on_damaged'`: executed every time a living entity is about to receive damage.
--- a/docs/scarpet/api/Entities.md
+++ b/docs/scarpet/api/Entities.md
@@ -801,7 +801,7 @@ defining the callback with `entity_event`.
 The following events can be handled by entities:
 
 *   `'on_tick'`: executes every tick right before the entity is ticked in the game. Required arguments: `entity`
-*   `'on_move'`: executes every time an entity changes position, at the end of the `setPos()` method. Required arguments: `entity, velocity, pos1, pos2`
+*   `'on_move'`: executes every time an entity changes position, invoked just after it has been moved to the new position. Required arguments: `entity, velocity, pos1, pos2`
 *   `'on_death'`: executes once when a living entity dies. Required arguments: `entity, reason`
 *   `'on_removed'`: execute once when an entity is removed. Required arguments: `entity`
 *   `'on_damaged'`: executed every time a living entity is about to receive damage.
--- a/src/main/java/carpet/mixins/Entity_scarpetEventsMixin.java
+++ b/src/main/java/carpet/mixins/Entity_scarpetEventsMixin.java
@@ -94,9 +94,6 @@ public abstract class Entity_scarpetEventsMixin implements EntityInterface
     @Inject(method = "setPos", at = @At("TAIL"))
     private void secondPos(CallbackInfo ci)
     {
-        events.onEvent(EntityEventsGroup.Event.ON_MOVE,
-        motion.x, motion.y, motion.z,
-        pos1.x, pos1.y, pos1.z,
-        this.pos.x, this.pos.y, this.pos.z);
+        events.onEvent(EntityEventsGroup.Event.ON_MOVE, motion, pos1, this.pos);
     }
 }
--- a/src/main/java/carpet/mixins/Entity_scarpetEventsMixin.java
+++ b/src/main/java/carpet/mixins/Entity_scarpetEventsMixin.java
@@ -3,6 +3,7 @@ package carpet.mixins;
 import carpet.fakes.EntityInterface;
 import carpet.script.EntityEventsGroup;
 import net.minecraft.entity.Entity;
+import net.minecraft.util.math.Vec3d;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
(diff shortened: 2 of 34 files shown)


Ghoulboy78:
now that could be rather cool. Does it trigger for players too?

src/main/java/carpet/script/EntityEventsGroup.java
```diff
@@ -108,6 +109,25 @@ public void addEvent(Event type, ScriptHost host, FunctionValue fun, List<Value>
...
+                        new EntityValue(entity),
+                        ListValue.fromTriple((Double) providedArgs[0],
+                        (Double) providedArgs[1],
+                        (Double) providedArgs[2]),
+                        ListValue.fromTriple((Double) providedArgs[3],
+                        (Double) providedArgs[4],
+                        (Double) providedArgs[5]),
+                        ListValue.fromTriple((Double) providedArgs[6],
+                        (Double) providedArgs[7],
+                        (Double) providedArgs[8])
```
Ghoulboy78, line 127: Maybe (and this is just my opinion) you could make it more readable by indenting the arguments of the `ListValue.fromTriple` so we can see which arguments are in that and which are in the `Arrays.asList`

lntricate1: I like the idea. I'm kinda new to github so not sure if I should commit that or attach a suggestion here, so I'll do the latter and see if it works lol. If it doesn't just reply to this.
```suggestion
                        ListValue.fromTriple((Double) providedArgs[0],
                            (Double) providedArgs[1],
                            (Double) providedArgs[2]),
                        ListValue.fromTriple((Double) providedArgs[3],
                            (Double) providedArgs[4],
                            (Double) providedArgs[5]),
                        ListValue.fromTriple((Double) providedArgs[6],
                            (Double) providedArgs[7],
                            (Double) providedArgs[8])
```

Ghoulboy78:
Wait, you forgot to add the documentation. You just need to add it to Entities.md.

Ghoulboy78:
Also for some reason I can't seem to get it to work when I test. Are you able to?

lntricate1:
Yeah, for me it works correctly. `/script run entity_event(player(), 'on_move', _(e,m,p1,p2) -> draw_shape('line', 100, 'from', p1, 'to', p2, 'player', player()));`

src/main/java/carpet/script/EntityEventsGroup.java
```diff
@@ -108,6 +109,25 @@ public void addEvent(Event type, ScriptHost host, FunctionValue fun, List<Value>
                 );
             }
         };
+        public static final Event ON_MOVE = new Event("on_move", 3)
```
Ghoulboy78, line 112: Oh, you made a typo here, it has 4 arguments not 3.

lntricate1: Oh that is correct, the argument count does not include the first argument which is the entity for every event.

src/main/java/carpet/mixins/Entity_scarpetEventsMixin.java
```diff
@@ -79,4 +84,19 @@ private void onRemove(CallbackInfo ci)
...
+        motion = this.velocity;
+    }
+
+    @Inject(method = "setPos", at = @At("TAIL"))
+    private void secondPos(CallbackInfo ci)
+    {
+        events.onEvent(EntityEventsGroup.Event.ON_MOVE,
+        motion.x, motion.y, motion.z,
+        pos1.x, pos1.y, pos1.z,
+        this.pos.x, this.pos.y, this.pos.z);
```
altrisi, line 100: You could also pass the `Vec3d`s to `ValueConversions.of(Vec3d)` instead of passing every single coordinate separately.

lntricate1: Ah that's really convenient, I didn't know about `ValueConversions.of`!

docs/scarpet/api/Entities.md
```diff
@@ -801,6 +801,7 @@ defining the callback with `entity_event`.
 The following events can be handled by entities:
 
 *   `'on_tick'`: executes every tick right before the entity is ticked in the game. Required arguments: `entity`
+*   `'on_move'`: executes every time an entity changes position, at the end of the `setPos()` method. Required arguments: `entity, velocity, pos1, pos2`
```
gnembon, line 804: I suggest not include implementation details in the documentation, but add information that might be important to the implementer
```suggestion
*   `'on_move'`: executes every time an entity changes position, invoked just after it has been moved to the new position. Required arguments: `entity, velocity, pos1, pos2`
```

Ghoulboy78:
I would suggest, after this pr gets accepted, to put your [entitytracker.sc](https://github.com/lntricate1/EntityTrackerScript/blob/main/entitytracker.sc) script in a pr to the 'utilities' folder of the scarpet app repository.

lntricate1:
> I would suggest, after this pr gets accepted, to put your [entitytracker.sc](https://github.com/lntricate1/EntityTrackerScript/blob/main/entitytracker.sc) script in a pr to the 'utilities' folder of the scarpet app repository.

If I do that, and then update the script, will that also update in the scarpet app repository or would I have to make a new PR every time?

Ghoulboy78:
> If I do that, and then update the script, will that also update in the scarpet app repository or would I have to make a new PR every time?

Yes, but only if you choose to update the version in your repository. It is more practical to move it to the scarpet app repository, as more people will be able to see it and use it, and you can make use of any libraries you may wish to add as well.

lntricate1:
> I would suggest, after this pr gets accepted, to put your [entitytracker.sc](https://github.com/lntricate1/EntityTrackerScript/blob/main/entitytracker.sc) script in a pr to the 'utilities' folder of the scarpet app repository.

Oh I need to ask a question. Since this PR is for the master branch, if it gets accepted, would the changes also be added to the 1.16.5 branch, or only to future releases?

James103:
> would the changes also be added to the 1.16.5 branch, or only to future releases?

The changes will only be ported to future releases, as 1.16.5 and under is unsupported since July 13, 2021 (Carpet mod 1.4.44).

Ghoulboy78:
> Oh I need to ask a question. Since this PR is for the master branch, if it gets accepted, would the changes also be added to the 1.16.5 branch, or only to future releases?

You can always backport it manually if you want, but that feels like it might be a bit of a pain.

docs/scarpet/api/Entities.md
```diff
@@ -801,6 +801,7 @@ defining the callback with `entity_event`.
 The following events can be handled by entities:
 
 *   `'on_tick'`: executes every tick right before the entity is ticked in the game. Required arguments: `entity`
+*   `'on_move'`: executes every time an entity changes position, invoked just after it has been moved to the new position. Required arguments: `entity, velocity, pos1, pos2`
```
Firigion, line 804: Suggest renaming `pos1`and `pos2` to `prev_pos` and `new_pos` to make it more explicit?

gnembon (approved):

PR #1100, opened 2021-09-10
Fixed number function not returning list length

tioalexis:
According to [this document](https://github.com/gnembon/fabric-carpet/blob/master/docs/scarpet/language/SystemFunctions.md#numberexpr), passing a list to `number()` should return the length of the list.


altrisi:
Wondering if it would make sense to make it create a NumericValue from the result of `Value#readDoubleNumber()` (except for NumericValue). Not sure if there's any situation where that wouldn't make sense for the function. Would probably need checking for Double.NaN to return null instead to keep behaviour.

Issue #1091, opened 2021-09-03
[Internal] Convert ParsedRule into an implementation detail


altrisi:
Since almost when I started contributing to Carpet, the mess around ParsedRule has always upset me.

I'd like to (if that's not a problem for you) to convert ParsedRule into an implementation detail, by promoting a new base interface for rules instead (obviously with the references to it used in extensions still existing but deprecated and recommending versions that return the interface), and after that cleaning up ParsedRule and maybe SettingsManager a bit, to make their generics a lot more checked and easier to touch (I haven't been able to touch anything around ParsedRule's generics because any time I did it would either make the game crash at random moments or fail to compile in some random line).

This would also allow different rule implementations for extensions, specific rule types in Carpet's manager, and would make it easier to just change things around the settings without breaking everything in existence.

Is it bad if I give it a try?

PR #1106, opened 2021-09-14
Added ability to add comments to carpet.conf

### LINKED ISSUE #1102: Allowing comments in carpet.conf
I'm ready to add it, but what should be the "keyword"? `//` or `#`?

Mostly in the header, since other places it would be more complicated.

Ghoulboy78:
Fixes #1102

### DIFF
--- a/src/main/java/carpet/settings/SettingsManager.java
+++ b/src/main/java/carpet/settings/SettingsManager.java
@@ -433,9 +433,11 @@ public class SettingsManager
                 {
                     confLocked = true;
                 }
-                String[] fields = line.split("\\s+",2);
+                String[] fields = line.split("\\s+", 3);
                 if (fields.length > 1)
                 {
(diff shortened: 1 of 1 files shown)


altrisi:
This won't work (properly), mainly because you are not saving those comments anywhere, and therefore when saving back anything they will be removed (what's the point of adding comments if they'll be removed after you do anything).

That's why you need a more complex version (and why I mentioned about only allowing comments at the start of the file, as a faster way of coding it, else it would need a bigger rework of config file handling).

Also why isn't this allowed?
```properties
# This should be a comment
```

The way you have it means that you won't get those, since you are only checking after an initial space (if you want to allow space then comment, you could instead check if either first or second term startsWith, and if second does make sure first one is empty).

I already started a bit on this in my branch for #1091, but if you want to make it go ahead and I'll merge later if necessary.

Ghoulboy78:
Well if ur doing this, then Ill close this.

PR #1110, opened 2021-09-18
allow specifying default game modes of /player spawned players

Trumeet:
Our survival server is using rcon to spawn bot players. However, currently Carpet sets their game modes to creative if the spawn command is not from a player. This behaviour causes troubles in a survival server. The best solution is to allow administrators to specify the default game mode of bots.

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -20,6 +20,7 @@ import net.minecraft.util.Identifier;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.ChunkPos;
 import net.minecraft.util.registry.Registry;
+import net.minecraft.world.GameMode;
 import net.minecraft.world.World;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -957,6 +958,12 @@ public class CarpetSettings
             validate = updateSuppressionBlockModes.class
     )
     public static String updateSuppressionBlock = "false";
+
+    @Rule(
+            desc = "Define the default game mode of /player spawn players when Carpet fails to obtain the current game mode of the calling player.",
+            category = COMMAND
+    )
+    public static GameMode playerCommandDefaultGameMode = GameMode.CREATIVE;
 
     public static int getInteger(String s) {
         try {
--- a/src/main/java/carpet/commands/PlayerCommand.java
+++ b/src/main/java/carpet/commands/PlayerCommand.java
@@ -264,7 +264,7 @@ public class PlayerCommand
                 () -> DimensionArgumentType.getDimensionArgument(context, "dimension").getRegistryKey(),
                 () -> source.getWorld().getRegistryKey() // dimension.getType()
         );
-        GameMode mode = GameMode.CREATIVE;
+        GameMode mode = CarpetSettings.playerCommandDefaultGameMode;
         boolean flying = false;
         try
         {


gnembon:
I would rather not proliferate the carpet options, rather add that to the options for /player. You could plug that in both places:
/player spawn Steve in survival at ... (all other optional defaults)
and
/player spawn Steve at 0 0 0 facing 0 0 in the_nether in survival
(please add both)

just make sure that the invoker/sender OP permissions are checked when any gamemode option is specified.

Trumeet:
> I would rather not proliferate the carpet options, rather add that to the options for /player. You could plug that in both places:
> /player spawn Steve in survival at ... (all other optional defaults)
> and
> /player spawn Steve at 0 0 0 facing 0 0 in the_nether in survival
> (please add both)
> 
> just make sure that the invoker/sender OP permissions are checked when any gamemode option is specified.

What do you mean by "when any gamemode option is specified"? Which gamemode option specifically would you like to check? Thanks.

Trumeet:
> I would rather not proliferate the carpet options, rather add that to the options for /player. You could plug that in both places:
> /player spawn Steve in survival at ... (all other optional defaults)
> and
> /player spawn Steve at 0 0 0 facing 0 0 in the_nether in survival
> (please add both)
> 
> just make sure that the invoker/sender OP permissions are checked when any gamemode option is specified.

Under the current implementation, in order to support both `spawn xxx in survival` and `spawn xxx at 0 0 0 facing 0 0 in the_nether in survival`, I have to write the whole spawn arguments twice:
```java
.then(literal("spawn").executes(PlayerCommand::spawn)
                                .then(literal("in") /* gamemode command logic */.then(literal("at")) /* at, facing, in again */)
                                .then(literal("at").then(argument("position", Vec3ArgumentType.vec3()).executes(PlayerCommand::spawn)
                                        .then(literal("facing").then(argument("direction", RotationArgumentType.rotation()).executes(PlayerCommand::spawn)
                                                .then(literal("in").then(argument("dimension", DimensionArgumentType.dimension()).executes(PlayerCommand::spawn)
                                                        .then(literal("in") /* gamemode command logic */)
                                        )))
                                ))
                        )
```

Are you sure this is what you want? Personally I would prefer just appending the `in gamemode` argument after `in dimension`.

Trumeet:
> I would rather not proliferate the carpet options

In our situation, we have a special relay to allow server members to execute `/player` commands using rcon themselves. We do not want our relay to handle syntax parsing ourselves, but letting anyone in the server to specify game mode for bots is obviously not a safe option. This is why I would like to use carpet options to limit the game mode rather than allowing rcon clients to specify them on the fly.

Ghoulboy78:
> Are you sure this is what you want? Personally I would prefer just appending the `in gamemode` argument after `in dimension`.
Well you can do it in a lot more compact manner, so its easy

altrisi:
Maybe something like the `execute` command does could be done (order-independent arguments), though then gamemode would need its own keyword. Not sure how difficult would that be, haven't really checked it.

gnembon:
> > I would rather not proliferate the carpet options
> 
> In our situation, we have a special relay to allow server members to execute `/player` commands using rcon themselves. We do not want our relay to handle syntax parsing ourselves, but letting anyone in the server to specify game mode for bots is obviously not a safe option. This is why I would like to use carpet options to limit the game mode rather than allowing rcon clients to specify them on the fly.

Ok, I get your situation. In this case this is a perfect usecase for custom carpet commands where you can add a scarpet app that the only thing it is doing is to expose certain usecases from other commands that are restricted to the regular users. Please, have a look at such a usecase that is covered in scarpet docs that 'replicates' effect command:
https://github.com/gnembon/fabric-carpet/blob/master/docs/scarpet/Full.md#custom-commands

Trumeet:
> > > I would rather not proliferate the carpet options
> > 
> > 
> > In our situation, we have a special relay to allow server members to execute `/player` commands using rcon themselves. We do not want our relay to handle syntax parsing ourselves, but letting anyone in the server to specify game mode for bots is obviously not a safe option. This is why I would like to use carpet options to limit the game mode rather than allowing rcon clients to specify them on the fly.
> 
> Ok, I get your situation. In this case this is a perfect usecase for custom carpet commands where you can add a scarpet app that the only thing it is doing is to expose certain usecases from other commands that are restricted to the regular users. Please, have a look at such a usecase that is covered in scarpet docs that 'replicates' effect command:
> https://github.com/gnembon/fabric-carpet/blob/master/docs/scarpet/Full.md#custom-commands

OK, thank you for your reply. I understand what is the correct solution to my situation. I will send an updated patch to support `in <gamemode>` arguments later. Just closing this right now.

PR #1119, opened 2021-10-01
Better search through rules

fxmorin:
Changed carpets search from startsWith to Contains.
It's been annoying me for a long time now, the fact that you never remember the command name, but you remember something that was in it.


CommandLeo:
There is `/carpet list TEXT` to search for carpet rules with a certain text in their names. Although I agree that having autocomplete directly in `/carpet` is better.

gnembon:
yeah, exactly. `/carpet list` is the google - it should even search in descriptions and stuff, not just rule names. What's wrong with that?

having said that, some vanilla tab completed options complete for after '_', so since carpet rules are camel cased, I suggest you check for name.startswith(q) || name.contains(title(q)) this way will not be slicing through word boundaries

fxmorin:
I will see if I have time this afternoon to change it

PR #1120, opened 2021-10-01
Added Non Movable Block List check

JPisaBrony:
This patch was designed to fix #1096 but it technically could be easily extended to prevent any sort of block from being pushed by a piston. Im not really sure if that would be a very useful feature so I just went with the original patch of just adding a way to configure which tile entities could be moved. I was also not completely sure about the implementaiton I should go with for saving a list in the configuration file since I couldn't find any anyone else using a list. I just went with a really basic comma seperated list since it was pretty simple and I thought it would be be easily understood in the config.


Firigion:
I using a custom block tag thrpugh a datapack a bad idea? Could carpet package a datapack that adds a blocktag adn have it customizable that way? Would not be customizable from in-game, but still.

PR #1121, opened 2021-10-02
Settings rework and encapsulation

### LINKED ISSUE #1091: [Internal] Convert ParsedRule into an implementation detail

### LINKED ISSUE #1102: Allowing comments in carpet.conf

altrisi:
Resolves #1091, Resolves #1102.

More changes related to settings:
- Adds support for more field types in `ParsedRule` (they are now easier to add)
- Uses a new `TypedField` class for managing `ParsedRule` fields
- Fixes all possible exceptions (I found) when setting rules caused by Carpet and handles them with nicer messages (such as entering letters to numeric fields, etc)
- Moves many `ParsedRule` methods to `RuleHelper` so those aren't needed in every implementation, since they can be made without accessing internal state
- Makes rule observers use a new `RuleObserver` `@FunctionalInterface` instead of a generic `TriFunction`
- Keeps fallback `@Deprecated(forRemoval)` methods and fields for those that were previously public (and were likely used, the ones that definitely weren't have been encapsulated) to prevent crashes with extensions using them

At the same time:

- Changes the translation system to not use `null` maps but empty ones
- Removes the old `CarpetServer#registerCarpetCommands(CommandDispatcher)`
- Changes `chainStone` setting to use an enum
- Makes `lightEngineMaxBatchSize` validator not set the rule field (it's not guaranteed to exist in the new API), but pass the value being set to the method that changes those

The intention is to after some versions make ParsedRule package-private, or at least make all its fields private (for more specifics see comments around the `@Deprecated` annotations). The motivation for this started after some binary incompatibilities caused by #1009, the difficulty I've had to be properly extend it, and some older incompatibilities I can't remember where did they come from.

For another PR I'd like to change a bit the Scarpet rule implementation (as stated in the TODOs) to be in a subclass of ParsedRule (or something like that), once that is properly encapsulated.

### DIFF
--- a/src/main/java/carpet/CarpetExtension.java
+++ b/src/main/java/carpet/CarpetExtension.java
@@ -7,6 +7,7 @@ import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.command.ServerCommandSource;
 import net.minecraft.server.network.ServerPlayerEntity;
 
+import java.util.Collections;
 import java.util.Map;
 
 public interface CarpetExtension
@@ -115,10 +116,10 @@ public interface CarpetExtension
      * 
      * @param lang A {@link String} being the language id selected by the user
      * @return A {@link Map<String, String>} containing the string key with it's 
-     *         respective translation {@link String} or {@link null} if not available
+     *         respective translation {@link String} or an empty map if not available
      * 
      */
-    default Map<String, String> canHasTranslations(String lang) { return null;}
+    default Map<String, String> canHasTranslations(String lang) { return Collections.emptyMap();}
 
     /**
      * Handles each call that creates / parses the scarpet expression.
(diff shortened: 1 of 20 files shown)


altrisi:
Turning into a draft until api movement discussions are finished.

altrisi:
Superseded by #1330.

PR #1124, opened 2021-10-10
fixed logger option bug

Sylk0s:
Fixed a small bug I noticed when working with the explosion logger. Putting in an invalid option is not caught by carpet mod. Basically you could run the command /log explosions invalid (where invalid can be any random string) and it would allow that. When it logs an explosions, the logger would just output the tick but no actual info about the explosion. All I did was add a check to the LogCommand.java class which checks if the given option is a valid option or not.

### DIFF
--- a/src/main/java/carpet/commands/LogCommand.java
+++ b/src/main/java/carpet/commands/LogCommand.java
@@ -13,11 +13,7 @@ import net.minecraft.server.command.ServerCommandSource;
 import net.minecraft.server.command.CommandManager;
 import net.minecraft.entity.player.PlayerEntity;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 import static com.mojang.brigadier.arguments.StringArgumentType.getString;
 import static net.minecraft.command.CommandSource.suggestMatching;
@@ -201,15 +197,19 @@ public class LogCommand
             Messenger.m(source, "r Unknown logger: ","rb "+logname);
             return 0;
         }
+        if (!Arrays.asList(LoggerRegistry.getLogger(logname).getOptions()).contains(option)) {
+            Messenger.m(source, "r Unknown option: ","rb "+option);
+            return 0;
+        }
         LoggerRegistry.subscribePlayer(player_name, logname, option);
-        if (option!=null)
-        {
-            Messenger.m(source, "gi Subscribed to " + logname + "(" + option + ")");
-        }
-        else
-        {
-            Messenger.m(source, "gi Subscribed to " + logname);
-        }
-            return 1;
+            if (option != null)
+            {
+                Messenger.m(source, "gi Subscribed to " + logname + "(" + option + ")");
+            }
+            else
+            {
+                Messenger.m(source, "gi Subscribed to " + logname);
+            }
+        return 1;
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/carpet/commands/LogCommand.java
```diff
@@ -13,11 +13,7 @@
 import net.minecraft.server.command.CommandManager;
 import net.minecraft.entity.player.PlayerEntity;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
gnembon, line 16: Please don't collapse imports

gnembon:
what about loggers that have more complex setup? like `/log counter red,green,blue` ?

src/main/java/carpet/commands/LogCommand.java
```diff
@@ -201,15 +197,19 @@ private static int subscribePlayer(ServerCommandSource source, String player_nam
...
         }
-        LoggerRegistry.subscribePlayer(player_name, logname, option);
-        if (option!=null)
-        {
-            Messenger.m(source, "gi Subscribed to " + logname + "(" + option + ")");
-        }
-        else
-        {
-            Messenger.m(source, "gi Subscribed to " + logname);
+        if (!Arrays.asList(LoggerRegistry.getLogger(logname).getOptions()).contains(option)) {
```
Fallen-Breath, line 200: This would remove the possiblility for loggers to have more flexible custom options, since the `option` field might only be a suggestion argument set for players

e.g. In carpet tis addition, ticket logger or item/xp orb loggers allow player to use `,` separated arguments as the logger argument, so it can display what exactly player wants to know

Fallen-Breath:
Imo if you do want to add a validication check for logger, you can create a validicating method in the logger, so other people can easily overwrite the default validicating method if they want to keep the flexibility, though this still requires some work

If you just want to fix #1124 then why not just fix the explosion logger itself

src/main/java/carpet/commands/LogCommand.java
Fallen-Breath: Ah gnembon has also mentioned this issue above xd

Sylk0s: Yeah that's what I realized. My thought was that I could make it so that the player could pass some specific value into the loggers option (for example an empty String Array was what I thought of but you could probably do this better) which would signify that it can have any input. I think I'm gonna try to rework it so that it doesn't break any of the current loggers but also allows for them to optionally be restrictive.

gnembon:
Or easier - support 'strict' option similar to carpet rules.

src/main/java/carpet/logging/HUDLogger.java
```diff
@@ -8,21 +8,26 @@
...
+            return new HUDLogger(LoggerRegistry.class.getField("__"+logName), logName, def, options, strictOptions);
         }
         catch (NoSuchFieldException e)
         {
             throw new RuntimeException("Failed to create logger "+logName);
         }
     }
 
-    public HUDLogger(Field field, String logName, String def, String[] options)
+    public HUDLogger(Field field, String logName, String def, String[] options, boolean strictOptions)
```
gnembon, line 28: I would include the constructor that doesn't have the strictOptions param and default it to false for potential API compatibility, and lets mark it as @Deprecated

PR #1149, opened 2021-11-01
XpNoCooldown and combineXpOrbs refactor

### LINKED ISSUE #1041: Removing or repurposing `combineXPOrbs`
Since it basically got added to Vanilla.

Would save half a mixin and a helper class.

### LINKED ISSUE #1145: XPnocooldown does not work as intended
Basically the title, XPnocooldown does not work as its supposed to, it does not instantly transfer to you
1.16.5:
<a href="https://ibb.co/gjHQBMS">6v-HISVJ7-Wz</a>
1.17.1:
bruh is too laggy

Crec0:
- Resolves #1145 
- Maybe resolves #1041 ?
- ~~Removes code duplication for merge behavior~~
- ~~Use vanilla methods and mixins to implement merge behavior~~
- ~~Add option to disable xp orb merging entirely~~
- ~~Reconfigure the rule to better support new features~~
- ~~Rename some files for consistency~~
- ~~Sort mixins in carpet.mixins.json alphabetically by rules~~


altrisi:
Problem with sorting the mixin json is that any open PRs that touch it will get conflicts.

Crec0:
Hmm true. I can rollback and add the mixins I made somewhere OR I can share the code which I used to sort them so people can do it too ?

gnembon:
Yeah, don't sort mixins - they are kinda' organized.

Regarding refactor - no need to go ham on that I think to apply immediate absorbtion, right? Aim for minimal diff to create a behaviour you want. Also - don't think maintaining no-merge behaviour is something useful at all - I would either remove previous merging behaviour completely or keep it under rule solely for the easter egg side effect (which you removed)

Crec0:
yea mixin sort was reverted. 

For the instantaneous pickup for xp, I tried to keep it same as it was in 1.16 and below. Mojang separated it from rest of the entity collisions. Guess I can nerf it a bit. do age % 10 or something.
For the no merge behaviour, I agree it serves no use. I guess I can revert the rule back to true/false where enabled it will work alongside vanilla merging, bringing back easter egg. Otherwise it does nothing.

src/main/resources/carpet.mixins.json
```diff
@@ -131,7 +132,6 @@
 
     "ServerPlayNetworkHandler_antiCheatDisabledMixin",
     "PlayerEntity_antiCheatDisabled",
-
```
gnembon, line 134: empty slots

Crec0: oboi, on it

Crec0:
Review changes done

gnembon (approved):

src/main/java/carpet/mixins/ExperienceOrbEntityMixin.java
```diff
@@ -1,87 +1,26 @@
...
 
     @Redirect(method = "onPlayerCollision", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/player/PlayerEntity;addExperience(I)V"))
-    void addXP(PlayerEntity playerEntity, int experience)
-    {
-        playerEntity.addExperience(experience);
-        if (CarpetSettings.xpNoCooldown) while(pickingCount > 1)
-        {
-            pickingCount --;
-            playerEntity.addExperience(experience);
+    void addXP(PlayerEntity player, int experience) {
```
gnembon, line 17: I think this is incorrect. The provided value represent remaining xp on a single orb that was already reduced by mending, meaning the rest of the xp stack may be lost. I.E. `i!=this.amount`. 

You might want to loop through all orbs in the stack letting them mend with a random piece each time (mimicking what vanilla would do next tick) until picking count goes to zero.

Crec0: Okay I fixed it

gnembon:
The mixin doesn't seem to apply after local testing. The chance has no effect.

My advise - don't use complex mixins with opcodes, fields and stuff. There is a convenient call to mixin into, send_pickup or in general for this one, I would just inject at head and cancel everything

gnembon:
for instance the following mixin is much simpler conceptually, and works
```
@Inject( method = "onPlayerCollision", cancellable = true, at = @At(
            value = "INVOKE",
            shift = At.Shift.AFTER,
            target = "Lnet/minecraft/entity/player/PlayerEntity;sendPickup(Lnet/minecraft/entity/Entity;I)V"
    ))
    private void addXP(PlayerEntity player, CallbackInfo ci) {
        if (CarpetSettings.xpNoCooldown) {
            int remainder;
            while (this.pickingCount > 0) {
                remainder = this.repairPlayerGears(player, this.amount);
                if (remainder > 0) {
                    player.addExperience(remainder);
                }
                player.experiencePickUpDelay = 0;
                this.pickingCount--;
            }
            discard();
            ci.cancel();
        }
    }
    ```

gnembon (approved):

Issue #1052, opened 2021-08-07
Server crashes when running command /spawn tracking start 


Killakilla2:
![image](https://user-images.githubusercontent.com/88597528/128613495-945b703e-f8d4-4615-8cb4-20f6001bb58a.png)

This creates a ticking world problem and per Apex hosting staff "This seems like an internal issue with the carpet mod, so this should be reported to the developers"

Server is fabric 1.17.1 and using fabric-carpet-1.17.1-1.4.44+v210714.jar

Killakilla2:
edit.  when using 1.4.43 the server does not crash using this command

gnembon:
The serve shouldn't crash, but that's an an odd situation to happen. I didn't fix the problem, but fix the crash for now. More info (in terms of what you were measuring and how) would be helpful here.

Killakilla2:
server ran the commnad fin yesterday but then today when I ran the command again it crashed on the 1.4.43.  In the first crash scenario i was AFK at a single witch hut farm with 3 spawn platforms.  The second time I was AFK at a raid farm but had not started the raid I was lookign to see the data of othe mobs spawning to see how well i spawn proofed the areas.

9:52:10] [Server thread/INFO]: [KillaKilla: Spawning tracking started.]
[09:52:13] [Server thread/ERROR]: Encountered an unexpected exception
net.minecraft.class_148: Exception ticking world
	at net.minecraft.server.MinecraftServer.method_3813(MinecraftServer.java:895) ~[intermediary-server.jar:?]
	at net.minecraft.class_3176.method_3813(class_3176.java:335) ~[intermediary-server.jar:?]
	at net.minecraft.server.MinecraftServer.method_3748(MinecraftServer.java:831) ~[intermediary-server.jar:?]
	at net.minecraft.server.MinecraftServer.handler$zke000$modifiedRunLoop(MinecraftServer.java:4390) ~[intermediary-server.jar:?]
	at net.minecraft.server.MinecraftServer.method_29741(MinecraftServer.java:676) ~[intermediary-server.jar:?]
	at net.minecraft.server.MinecraftServer.method_29739(MinecraftServer.java:270) ~[intermediary-server.jar:?]
	at java.lang.Thread.run(Thread.java:831) [?:?]
Caused by: java.lang.NullPointerException: Cannot invoke "java.lang.Long.longValue()" because the return value of "java.util.HashMap.get(Object)" is null
[...]

PR #1151, opened 2021-11-01
Missing semicolon in draw_beta.sc

Ghoulboy78:


Ghoulboy78:
I'm also gonna take this opportunity to add some explanatory docs to each method in `shapes.scl`, to explain what it does and how it works.

altrisi:
Another thing: `player` should be `player()` in line 55 (or not specifying player in case you want to also give feedback when console does it).

Also I still think global scope should be kept, so the console can use the command (console can't use player scoped commands). At least until something that requires player scope is added.

altrisi:
BTW I found another bug, not sure if you know what causes it: `ball` is not making the same shape as `sphere` (not counting interior ofc):

Sphere:

imagen 
imagen

Ball:

imagen 
imagen

Ghoulboy78:
Actually `draw_sphere `was entirely unaffected by #1052, as along with `draw_diamond`, it doesn't use any calls to `fill_flat` to generate points. However, I am making a pr which fixes this issue, which I noticed when testing #1153.

Ghoulboy78:
I'm not going to change `shapes.scl` in this pr, if I'm already making another pr to fix all the other issues.

altrisi:
Please do the player thing (it throws else) though and consider the scope thing.

Ghoulboy78:
ah ok, I didn't read comment properly, I'll change right now

gnembon (approved):

PR #1153, opened 2021-11-01
Giving scarpet powers of perimeterinfo

### LINKED ISSUE #153: remove /perimeterinfo by adding scarpet all its powers

Ghoulboy78:
Fixes #153

Basically, this adds the `can_spawn()` function, which takes an entity or string, and a position. The reason it takes an entity or a string is cos passing an entity s a lot faster, cos then the function doesn't have to initiate its own entity and then kill it. I'll add this to the docs ofc.

As a side note, when coding this I came across the issue of drawing spheres, hence all the issues I have created in the past few days, cos I wanted to use those shapes. I'll leave this pr a draft, therefore, until another pr fixing shapes.scl is merged (and I'll link that here).

Still left to do:
- [ ] Add docs
- [ ] Use `shapes.scl` sphere instead of current slow version, once I fix that

### DIFF
--- a/src/main/java/carpet/script/CarpetScriptServer.java
+++ b/src/main/java/carpet/script/CarpetScriptServer.java
@@ -105,6 +105,7 @@ public class CarpetScriptServer
         registerBuiltInScript(BundledModule.carpetNative("draw_beta", false));
         registerBuiltInScript(BundledModule.carpetNative("shapes", true));
         registerBuiltInScript(BundledModule.carpetNative("distance_beta", false));
+        registerBuiltInScript(BundledModule.carpetNative("perimeter_info_beta", false));
     }
 
     public CarpetScriptServer(MinecraftServer server)
--- a/src/main/java/carpet/script/api/Entities.java
+++ b/src/main/java/carpet/script/api/Entities.java
@@ -13,6 +13,7 @@ import carpet.script.value.NBTSerializableValue;
 import carpet.script.value.BooleanValue;
 import carpet.script.value.NumericValue;
 import carpet.script.value.Value;
+import carpet.utils.Messenger;
 import carpet.utils.PerimeterDiagnostics;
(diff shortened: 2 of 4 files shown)


Ghoulboy78:
No need for shapes.scl, I just implemented `perimeter_info` as a function on its own, cos the other way was way too slow.

docs/scarpet/api/Entities.md
```diff
@@ -91,6 +91,19 @@ Spawns and places an entity in world, like `/summon` vanilla command. Requires a
...
 you get the entity back as a return value, which is swell.
 
+### `can_spawn(name, pos), can_spawn(entity, pos)`
+
+Returns a boolean value of whether or not an entity (either as a string or as an entity passed as an argument) cand spawn
+in that spot. The reason you might want to use a pre-existing entity is that if you just pass in a string, it will spawn
+in that entity and then kill it instantaneously, and if you are doing multiple of these calls on the same entity type, it
+could get rather slow after a while.
+
+### `perimeter_info(name, pos)`
```
gnembon, line 101: No point of adding that - the script author can do their own iteration.

gnembon:
I would go over the perimeter info code and identify API calls that would be needed to support it. i.e. `mob_spawns()` returning mobs with their spawning parameters at a given position, `can_spawn()` for static checks and `will_spawn()` for live checks including calling multiple times to estimate probability, Rest of the logic - block iteration, conditions, formatting, should be part of a script

docs/scarpet/api/Entities.md
Ghoulboy78: Ye but I tried that and it was too slow

Ghoulboy78:
Welp, even with modifications the `can_spawn` is too slow alone for me to do `perimeter_info` app without `perimeter_info` call.
I'll hold off adding docs until we decide on whether or not to keep it as it is rn.

PR #1154, opened 2021-11-01
Fixing shapes.scl

### LINKED ISSUE #1152: Shapes.scl broken by new method of adding lists and sets via +=
Now `list_variable += set` will make the set an element of the list, as opposed to previously when it would add all the elements. While this is good in most cases, it screws up. I have a solution for that though, and hear it out.

I propose to add streams, or sequentially generated objects. The way they will work is that you call `next(stream)` and it will evaluate the next item, and you declare them via `stream = stream(_(prevstate)->{'ret'->prevstate:'acc'+1, 'acc' -> prevstate:'acc'+1})` 
So basically, the `prevstate` is a map of the previous state, and you can set whatever state you want, holding whatever variables yo want as keys, and then there is a `'ret'` element of the map (can be renamed `return`, whatevs) which means whenever you do like:
`x = next(stream)`, x will be assigned to that `ret` parameter. This would fix it as then I could sequentially generate the lists and simultaneously add them .

Now after further thought, the above problem can be fixed using `x = x + set` isntead of `+=`, but I'm still gonna leave this stream object stuff here, cos I think it's still interesting. I will submit a pr to fix the shapes.scl problem, then make another issue to discuss the stream stuff.

Ghoulboy78:
This pr will fix a lot of problems in shapes.scl:

- [ ] Lack of documentation comments explaining code
- [x] Fixes #1152
- [ ] Spheres are not spherical
- [ ] Balls are laterally swollen
- [ ] Other stuff that will indubitable pop up

### DIFF
--- a/src/main/resources/assets/carpet/scripts/shapes.scl
+++ b/src/main/resources/assets/carpet/scripts/shapes.scl
@@ -54,19 +54,19 @@ draw_filled_diamond(args)->(
 );
 
 draw_pyramid(args)->(
-    [pos, radius, height, pointing, orientation, fill_type, is_square] = args;//todo change to hollow boolean arg
+    [pos, radius, height, pointing, orientation, fill_type, is_square] = args;
     positions = {};
     hollow = fill_type=='hollow';
     pointup = pointing=='up';
     for(range(height),
         r = if(pointup, radius * ( 1- _ / height) -1, radius * _ / height);
-        positions += fill_flat(pos, _, r, is_square, orientation, !((pointup&&_==0)||(!pointup && _==height-1)) && hollow)//Always close bottom off
+        positions = positions + fill_flat(pos, _, r, is_square, orientation, !((pointup&&_==0)||(!pointup && _==height-1)) && hollow)//Always close bottom off
     );
     positions
 );
 
 draw_prism(args)->(
-    [pos, rad, height, orientation, fill_type, is_square]=args;//todo change to hollow boolean arg
+    [pos, rad, height, orientation, fill_type, is_square]=args;
     positions = {};
     hollow = fill_type =='hollow';
     radius = rad+0.5;
@@ -86,7 +86,7 @@ draw_prism(args)->(
                 positions+= _ + offset_vector * offset
             )
         ),
-        
+
         for(range(height),
             offset = _;
             for(filled_circle_points,
@@ -111,7 +111,10 @@ fill_flat_circle(pos, offset, dr, orientation, hollow) ->( //Bresenham circle al
 
     positions = {};
 
-    positions += fill_flat_circle_points(pos, orientation, offset, hollow, a, b);
+    //special case for radius 0 or it messes up weirdly
+
+
+    positions = positions + fill_flat_circle_points(pos, orientation, offset, hollow, a, b);
 
     while(a<=b, 10^10,
         if(d<=0,
@@ -120,7 +123,7 @@ fill_flat_circle(pos, offset, dr, orientation, hollow) ->( //Bresenham circle al
             b=b-1
         );
         a+=1;
-        positions += fill_flat_circle_points(pos, orientation, offset, hollow, a, b)
+        positions = positions + fill_flat_circle_points(pos, orientation, offset, hollow, a, b)
     );
     positions
 );


Ghoulboy78:
Ok, now just need to update draw_beta.sc incase something drastic changed

Ghoulboy78:
I removed the fix for bs suggestions in draw_beta cos it will automatically be fixed when #1015 is merged.

Ghoulboy78:
@gnembon @altrisi this is ready for review

Firigion (changes requested):
Sideways hollow pyramides and cones ahve strange shapes. If point is `up` there's a protruding bit in the uppar and lower sides if direction is `x` and on the right and left side if direction is `z`. All shapes should be isotropic in at least  two axial directions (in this case, y and x or z).
image

Pyramid algorithm seems a tiny bit off, old pyramids look better. Here's old on the left, new on the right, diff in the back.
image

Finally sphere look too square and (again), unisotropic. Old sphere looksed much better. If you didn't end up improving the algo, I'd suggest just implementing the old one here.
image

The rest looks awsome and is ready to merge by me.

src/main/resources/assets/carpet/scripts/draw_beta.sc
```diff
@@ -21,14 +21,13 @@ __config() -> {
     },
     'arguments'->{
         'center'->{'type'->'pos', 'loaded'->'true'},
-        'radius'->{'type'->'int', 'suggest'->[], 'min'->0},//to avoid default suggestions
+        'radius'->{'type'->'int', 'min'->0},
```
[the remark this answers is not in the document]
Ghoulboy78: ik, but that's because it rests on the merging of #1015, which would fix it.

src/main/resources/assets/carpet/scripts/draw_beta.sc
```diff
@@ -1,4 +1,4 @@
-import('shapes','draw_sphere', 'draw_diamond', 'draw_filled_diamond', 'draw_pyramid', 'draw_prism'); //importing all the shape funcs from the other app
+import('shapes','draw_sphere', 'draw_diamond', 'draw_filled_diamond', 'draw_pyramid', 'draw_prism'); //importing all the shape funcs from shapes.scl
 
 __config() -> {
     'commands'->{
```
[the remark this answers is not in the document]
Ghoulboy78: Thanks, I hadn't noticed

src/main/resources/assets/carpet/scripts/draw_beta.sc
```diff
@@ -21,14 +21,13 @@ __config() -> {
...
     'arguments'->{
         'center'->{'type'->'pos', 'loaded'->'true'},
-        'radius'->{'type'->'int', 'suggest'->[], 'min'->0},//to avoid default suggestions
+        'radius'->{'type'->'int', 'min'->0},
         'replacement'->{'type'->'blockpredicate'},
-        'height'->{'type'->'int', 'suggest'->[],'min'->0},
+        'height'->{'type'->'int', 'min'->0},
         'orientation'->{'type'->'term', 'suggest'->['x','y','z']},
         'pointing'->{'type'->'term','suggest'->['up','down']},
         'hollow'->{'type'->'term','suggest'->['hollow','solid']},
```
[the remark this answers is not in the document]
Ghoulboy78: hmm, yes I see your point. Ill do that

Ghoulboy78:
Ok, so I've finished the review points (finally). I also came across #1374 when looking a the pyramid code, and had to implement a workaround for it.

Ghoulboy78:
Note that hollow pyramids are still kinda broken if the height is less than the radius, but I have literally 0 clue how to fix that, and also my Java implementation doesn't even support hollow pyramids (maybe for that reason, though I don't remember now). I'm contemplating yeeting hollow pyramids entirely, but idk.

Firigion:
> Note that hollow pyramids are still kinda broken if the height is less than the radius, but I have literally 0 clue how to fix that, and also my Java implementation doesn't even support hollow pyramids (maybe for that reason, though I don't remember now). I'm contemplating yeeting hollow pyramids entirely, but idk.

Take a look at my implementation for hollow pyramids in Scarpet edit. It's not trivial, but it's doable with decent efficiency.

Ghoulboy78:
OK I looked at Firi's scarpet-edit pyramids, but the implementation was essentially the same as mine as far as I can see. The problem persists, and tbh I doubt there's much that can be done about that.

Firigion:
No, mine works just fine (you scared me for a bit there)
image
cone works too, btw (filled in with white so you can se ethe cutout clearer)
image

Ghoulboy78:
Wait what? I swear I copied the code...

Ghoulboy78:
Ok, thanks to firi I've actually gotten them working:
image

Firigion:
lol, nice. AS discussed on discord: I noticed you don't allow for even sided pyramids, cones, spheres etc, and the reason you quote is that you'd have to bias in some direction to get say a 6x6 pyramid. I'd argue that having the option is better than not having it and biasing either always to the positive direction is a good idea. Alternatively, if `~ ~ ~` was given as coordinates, you could check what position the player was on and see if it's biased towards some direction (for example, in the negative half of the block) and use that as bias direction.

src/main/resources/assets/carpet/scripts/shapes.scl
```diff
@@ -1,15 +1,20 @@
 //This is a library app, mostly for the draw command, which allows to import these shape-drawing functions into your own app.
-//Arguments for each are to be given as a singleton list, for ease of use. This is with the exception of the fill_flat functions, with
-//the assumption that they're unlikely to be used outside of internal calls, but it is still available to be called if necessary.
-//Make sure you read the arguments and input them in the correct order, or it will mess up.
-//these functions return a set of all the positions in the shape, given the parameters, without repeating a position.
+//The _fill_flat functions are mostly used in internal calls, but can be used outside as well
```
altrisi, line 2: Maybe don't suggest that (using them). That way it's implied it's not something supported and it can be changed or removed without breaking anything.

src/main/resources/assets/carpet/scripts/shapes.scl
```diff
@@ -1,15 +1,20 @@
...
-
-draw_sphere(args)->(
-    [centre, radius, hollow] = args;
+//This algorithm turns half a circle (in radians), then at each angle
+//it turns another half circle in radians and then fills in the other half using - signs
+//For a filled circle we just fill in the end points of the circle
+//Algorithmic efficiency:
+//  O(n^2) for hollow spheres cos 2d surface
+//  O(n^3) for filled spheres cos 3d volume
+draw_sphere(centre, radius, hollow)->(
```
altrisi, line 14: But you don't say what it does! Explaining how it works is nice, but it's missing what it does.

src/main/resources/assets/carpet/scripts/shapes.scl
```diff
@@ -20,60 +25,90 @@ draw_sphere(args)->(
...
 );
 
-draw_pyramid(args)->(
-    [pos, radius, height, pointing, orientation, fill_type, is_square] = args;//todo change to hollow boolean arg
+//Draws a bunch of circles or squares of decreasing or increasing radii based on where it's pointing.
+//Circles are drawn using _fill_flat_circle
+//Algorithmic efficiency:
+//  O(n^2) for hollow cones / pyramids
+//  O(n^3) for filled cones / pyramids
+draw_pyramid(pos, radius, height, pointing, orientation, fill_type, is_square)->(
```
altrisi, line 73: Again no comment on what it does! Ik, ik, the name kinda gives it away, but it's always nice to confirm it more than the name.

altrisi: Same for some others.

Or was this going to have a docs section elsewhere? Can't remember.

src/main/resources/assets/carpet/scripts/shapes.scl
```diff
@@ -97,134 +132,178 @@ draw_prism(args)->(
     positions
 );
 
-fill_flat(pos, offset, dr, rectangle, orientation, hollow)->
+//Allows to fill a rectangle or circle
+//'pyramid' parameter is cos this function is called by draw_prism and draw_pyramid
+//Now in pyramids, I want circles to be ever so slightly thicker,
+//whereas in prisms I don't, hence the parameter.
+//The reason I do abs() of the radius for _fill_flat_circle is because sometimes,
+//the calculations result in -0 being inputted as a radius.
```
altrisi, line 140: Another (probably cheaper?) way to get rid of `-0` should just be adding 0 to it. Wondering, how does it cause problems? It should behave like 0 in most cases.

src/main/resources/assets/carpet/scripts/shapes.scl
```diff
@@ -97,134 +132,178 @@ draw_prism(args)->(
...
-            c_for(c = -b, c<=b, c+=1,
-                positions += [x + offset, a+y,c + z];
-                positions += [x + offset,-a+y,c + z]
+            for(range(-a, a+1),
+                positions += [x + offset,y+b,_ + z];
+                positions += [x + offset,y-b,_ + z]
             )
         ),
         orientation=='y',
         if(hollow,
```
altrisi, line 224: IMO this is very difficult to realize that it's changing branch in an `if`, been confused for a while in why it was repeated.

altrisi (commented):
First quick review! Very small things for now, going to actually test it and follow the code further rn.

altrisi:
Seems to be making bigger spheres than `/draw`, from some sides at least. Diff:



Repro:
- /draw ...
- /draw_beta ... (same args)
- /draw ... air

altrisi:
Similar thing for cone, but the diff is points.

altrisi:
May be in the general circle function.

altrisi:
Cuboid seems to work fine, cylinder too. Diamond seems to be one block bigger?
Pyramid seems fine as well.

And after another test, there's more sphere diff with bigger radius.

Ghoulboy78:
Well the sphere algorithm is gnembon's code, not mine. The cone code is likely due to the circle algorithm as u said, but tbh im not too bothered. And about the diamond, do you mean filled diamonds or hollow diamonds? Cos ik that hollow diamonds have the exact same code (copy pastad), tho im not 100% sure about filled diamonds.

src/main/resources/assets/carpet/scripts/shapes.scl
Ghoulboy78: k

src/main/resources/assets/carpet/scripts/shapes.scl
```diff
@@ -1,15 +1,20 @@
 //This is a library app, mostly for the draw command, which allows to import these shape-drawing functions into your own app.
-//Arguments for each are to be given as a singleton list, for ease of use. This is with the exception of the fill_flat functions, with
-//the assumption that they're unlikely to be used outside of internal calls, but it is still available to be called if necessary.
-//Make sure you read the arguments and input them in the correct order, or it will mess up.
-//these functions return a set of all the positions in the shape, given the parameters, without repeating a position.
+//The _fill_flat functions are mostly used in internal calls, but can be used outside as well
+//These functions return a set of all the positions in the shape, given the parameters, without repeating a position.
 //This way, you can efficiently iterate over the positions once having generated them. See draw_beta.sc for an example. todo change to draw once renamed
+//You can also read the documentation code above each function to see what it does, how it does it and how well it does it
+//One way you can speed things up is by threading the operations (see draw_beta.sc)
```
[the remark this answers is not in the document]
Ghoulboy78: defo

src/main/resources/assets/carpet/scripts/shapes.scl
Ghoulboy78: err... got caught up in the explanation, but will do

src/main/resources/assets/carpet/scripts/shapes.scl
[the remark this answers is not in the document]
Ghoulboy78: Wait, no. I will still suggest to thread if the programmer wants, I'll just remove the "(see draw_beta.sc" bit.

src/main/resources/assets/carpet/scripts/shapes.scl
```diff
@@ -1,15 +1,20 @@
...
+//This algorithm turns half a circle (in radians), then at each angle
+//it turns another half circle in radians and then fills in the other half using - signs
+//For a filled circle we just fill in the end points of the circle
+//Algorithmic efficiency:
+//  O(n^2) for hollow spheres cos 2d surface
+//  O(n^3) for filled spheres cos 3d volume
+draw_sphere(centre, radius, hollow)->(
     positions = {};
     [cx,cy,cz]=centre;
+    radius = floor(radius);
```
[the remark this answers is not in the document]
Ghoulboy78: actually not my code, but ig the idea is that it doesn't make sense to send a non-integer radius. But if we send say, 5.9, then you'll end up drawing most of the blocks of a sphere radius 6, but with one or two slightly pushed in. But I'll remove it and put it in the docs.

src/main/resources/assets/carpet/scripts/shapes.scl
Ghoulboy78: Well the reason I opened #1374 was because it was breaking this. It was actually quite a hassle to figure out the culprit, but my workaround adds basically 0 compute.

src/main/resources/assets/carpet/scripts/shapes.scl
Ghoulboy78: Wdym? If ur asking about the orientations thing (where I have a different branch for x orientation, y orientation and z orientation), that was an old decision by gnembon, and I fully agree with it.

altrisi:
I tested hollow for diamonds IIRC (or default if hollow isn't default, but I think they were hollow?).

Just did this with the different shapes, with quite a big size (around 50 or100 IIRC):
> Repro:
>    /draw ...
>    /draw_beta ... (same args)
>    /draw ... air

Expected to be nothing remaining, always got blocks.

Ghoulboy78:
all right, the reason for this is because the /draw command itself is kinda flawed (i.e. the shapes in it are).
Firstly, it uses the old algorithm for circles, so obv there r gonna be differences there, and the same dealio for spheres. Square pyramids woulda been the exact same, but now I use relatively different code to generate them, and that accounts for minor differences.
Square prisms shouldnt have changed, and if they have that's an issue.
And for diamonds i am perplexed, cos there should be 0 changes there, Ill see what's going on.

Ghoulboy78:
Ok, that seems weird. I tested diamond, and it works perfectly, as expected. I think it may be because /draw can only draw hollow diamonds, and the default option for /draw_beta diamond is filled, hence the blocks left behind.

PR #1156, opened 2021-11-03
Optimize explosion BlockState caching

lntricate1:
Instead of caching 2 BlockState HashMaps, this uses 1 HashMap for BlockState and 1 HashMap for blast resistance, meaning that more calculations are cached, using only the necessarry ones. Also doesn't try to blow up air.

### DIFF
--- a/src/main/java/carpet/helpers/OptimizedExplosion.java
+++ b/src/main/java/carpet/helpers/OptimizedExplosion.java
@@ -465,10 +465,10 @@ public class OptimizedExplosion
             {
                 posImmutable = posMutable.toImmutable();
                 BlockState state = world.getBlockState(posImmutable);
-                FluidState fluid = world.getFluidState(posImmutable);
                 Float resistance = null;
-                if (state.getMaterial() != Material.AIR)
+                if (!state.isAir())
                 {
+                    FluidState fluid = world.getFluidState(posImmutable);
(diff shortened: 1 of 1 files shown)


src/main/java/carpet/helpers/OptimizedExplosion.java
```diff
@@ -457,38 +458,44 @@ private static boolean checkAffectedPosition(Explosion e, double xRel, double yR
...
-
-            if (state.getMaterial() != Material.AIR)
-            {
-                float resistance = Math.max(state.getBlock().getBlastResistance(), fluid.getBlastResistance());
-
-                if (eAccess.getEntity() != null)
+                BlockState state = world.getBlockState(posImmutable);
+                FluidState fluid = world.getFluidState(posImmutable);
+                Float resistance = null;
+                if (state.getMaterial() != Material.AIR)
```
gnembon, line 470: I believe there is world.isAir() function - not sure which would be faster here

gnembon:
Good idea. It would be great if someone else could test a few cases if the behaviour is the same, but seems good.

src/main/java/carpet/helpers/OptimizedExplosion.java
lntricate1: Looking at that particular code again, I think I can save some more calculations by moving getFluidState into that if statement. I will make that change and then test world.isAir() using Spark to see which is faster.

gnembon (approved):

PR #1157, opened 2021-11-04
Explosions mixin rework

Crec0:
Moved `doExplosionB` in OptimizedExplosions to be entirely mixin based


gnembon:
I reckon its not 'that' important because it still leaves the entire thing in similar state. Plus, most ugly / nasty compatibility issues have been already solved. Plus - seems like if either mod wants to modify explosions it makes sense to take all or nothing. Wander what could happen if two mods would successfully mixin to explosion optimizations and didn't cause any issue - would that make a monster?

gnembon:
I am voting to leave this one out. Maybe this just should be in vanilla really.

Crec0:
Reason for this change was the mods or api’s which rely on triggering explosion based events shouldn’t have the need to mixin into carpet mod’s implementation. they can just mixin into vanilla without worrying that it will cause issues. Since it resolves this unnecessary conflicts and still keeps explosions optimized. I do agree it should be vanilla behaviour but until it is, we have to keep this. Also accounts backward compatibility for people who don’t wanna play on latest versions.

gnembon:
Looks good. Tested locally - it blows blocks.

There is a high risk of incompatibility with other mods that this would introduce, plus would need to verify it works the same, so if others could test it that would be great, and will merge it after 1.18 is out so in case there are some obvious incompatibilities, they could be resolved later.

gnembon:
Please rebase on current master.

Crec0:
Rebased and squashed. Had to force push because of conflict with origin not being updated

Issue #890, opened 2021-05-28
Add: Scarpet functions to query and modify player's `abilities`


James103:
Currently (as of Carpet mod 1.4.35 for Minecraft 1.16.5), there are abilities listed in the player data for each player which are used and modified by the game and some other mods, but for which (S)Carpet has no access to except via `/data get entity` commands.

It would be great if there was a new section in `query` and `modify` called `ability`, allowing Scarpet apps to view and perhaps modify the abilities of a player. These functions could be used to change a player's abilities without changing his/her game mode.

For example,
`modify(p, 'ability', 'invulnerable', false)` allows you to take non-void damage when in Creative mode.
`modify(p, 'ability', 'mayfly', true)` allows you to fly even in Survival mode.
`modify(p, 'ability', 'instabuild', true)` gives you instant break powers as if you were in Creative mode.
`modify(p, 'ability', 'walkSpeed', 1.0)` makes you walk 10x faster (stacks with attributes).
`modify(p, 'ability', 'mayBuild', false)` disables placing and breaking blocks for you, even in Survival or Creative mode.
`modify(p, 'ability', 'flying', false)` forces you to stop flying.
`modify(p, 'ability', 'flySpeed', 0.4)` allows you to fly 8x faster than normal while in Spectator mode (2x faster than the `0.2` max speed achievable by scrolling the wheel)

`query(p, 'abilities')` returns a list of all the player's abilities and their current states. (Optional to implement)
`query(p, 'ability', 'flying')` checks whether you are flying or not, which could be used to drain extra (food) energy while in creative flight.
(and so on and so forth, see `modify(p, 'ability', ...)` above for full list)

---

Practical example: Suppose that you want to allow creative flight when you have an elytra and 100 XP levels at the cost of massively increased food energy usage and XP usage.

In event handler for `__on_tick()`:
```
if(player() == null, return());
player = player();

if(player~'gamemode' == 'creative' || player~'gamemode' == 'spectator',
	modify(player, 'ability', 'mayfly', true);
	return();
);

if(inventory_get(player, -3):0 == 'elytra' && player~'xp_level' >= 100,
	modify(player, 'ability', 'mayfly', true);
	, // else
	modify(player, 'ability', 'mayfly', false);
);

if(query(player, 'ability', 'flying'),
	modify(player, 'add_exhaustion', 0.2);
	modify(player, 'add_xp', -1);
);
```

gnembon:
yeah

James103:
[...]

PR #1160, opened 2021-11-06
Added player abilities actions to the entity API

Gnottero:
This PR adds some new actions for the `modify()` function.
The new actions are the following:

- `may_fly` -> Allows to modify the player's `mayfly` ability (whether the player can fly or not).
- `flying` -> Allows to modify the player's `flying` ability (whether the player is flying or not).
- `invulnerable` -> Allows to modify the player's `invulnerable` ability (whether the player can take damage or not). Now, it also works for players in creative mode.
- `may_build` -> Allows to modify the player's `mayBuild` ability (whether the player can place blocks or not).
- `insta_build` -> Allows to modify the player's `instaBuild` ability (whether blocks placed are reduced from the ItemStack count or not. Also controls if the player needs arrows to shot or not).
- `fly_speed` -> Allows to modify the player's `flySpeed` ability (the speed at which the player moves while flying).
- `walk_speed` -> Allows to modify the player's `walkSpeed` ability (the speed at which the player moves while walking).
- `selected_slot` -> Allows to modify the selected slot of the player.

I also added all the missing queries, related to these `modify()` actions


Ghoulboy78:
Maybe change `mayfly` to `may_fly`, cos it's two words

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -1,13 +1,6 @@
...
 
-import carpet.fakes.BrainInterface;
-import carpet.fakes.EntityInterface;
-import carpet.fakes.ItemEntityInterface;
-import carpet.fakes.LivingEntityInterface;
-import carpet.fakes.MemoryInterface;
-import carpet.fakes.MobEntityInterface;
-import carpet.fakes.ServerPlayerEntityInterface;
-import carpet.fakes.ServerPlayerInteractionManagerInterface;
+import carpet.fakes.*;
```
Ghoulboy78, line 3: Try to avoid collapsing all imports like that. I know the IDE does it and it's annoying, but please expand this

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -20,39 +13,32 @@
 import com.google.common.collect.Sets;
 import com.mojang.brigadier.StringReader;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
-import net.minecraft.entity.SpawnGroup;
+import net.minecraft.command.EntitySelector;
+import net.minecraft.command.EntitySelectorReader;
+import net.minecraft.entity.*;
```
Ghoulboy78, line 18: same here pls

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -74,16 +60,7 @@
...
-import java.util.Collection;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
Ghoulboy78, line 63: And here

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -963,6 +940,38 @@ private static void updateVelocity(Entity e, double scale)
             }
             if (e instanceof LivingEntity) ((LivingEntity) e).setHealth(health);
         });
+
+        put("mayfly", (e, v) -> {
+            if (v == null) {
+                throw new InternalExpressionException("'mayfly' requires parameters");
+            }
```
Ghoulboy78, line 947: This is handled automatically, no need for check

Ghoulboy78:
Looks pretty good, I would recommend adding the `query()` equivalents (so `query(player, 'may_fly')` as well as `query(player, 'fly_speed')`.
Also don't forget to add it to the docs. You don't need to modify Full.md, just add it to Entities.md in the appropriate place, and it will all get merged together into Full.md automatically later on.

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -963,6 +940,38 @@ private static void updateVelocity(Entity e, double scale)
...
+                }
+                player.sendAbilitiesUpdate();
+            }
+        });
+
+        put("fly_speed", (e, v) -> {
+            Float flySpeed;
+
+            if (v == null) {
+                flySpeed = 0.05F;
```
altrisi, line 964: This doesn't seem right...

The problem I see is that it seems counter-intuitive that modify with no values would reset it to a default value. I'd instead mention the default value in the docs and remove the null branch.

Else the call to reset to default would look something like `modify(entity, 'fly_speed')`, that doesn't really look right to me.

Ghoulboy78: Well actually I'm farily certain that caling modify with no value causes an error, so that branch of code would never trigger.

altrisi: Nope, it's totally fine, for things such as `modify(entity, 'kill')`, modify(entity, 'remove'), and/or some like that.

Ghoulboy78: ah oop my bad

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: nvm abt this

altrisi: It actually is, there's a try-catch for NPE outside in the method that calls this.

Ghoulboy78: Well do mind then, never mind me telling you not to mind about this.

Ghoulboy78: (also I thought I remembered seeing a try/catch there, hence why I sent the message, but then seeing this comment https://github.com/gnembon/fabric-carpet/pull/1160#discussion_r744160669 confused me a bit cos I didn't understand the context)

gnembon:
I know that flying seems the most appealing, but there are more abilities, that could be added for completion. Suggest adding all relevant ones

James103:
See #890 for a full list of abilities.

Basically, at least the following abilities should have `query` and `modify` functions:
* [ ] mayfly
* [ ] flying
* [ ] invulnerable
* [ ] mayBuild
* [ ] instaBuild

There are two abilities that aren't that important due to attributes, but could be included here for completeness:
* [ ] flySpeed
* [ ] walkSpeed

gnembon:
Makes sense what James suggested, also use snake case, since its how the API is constructed. Let us know @Gnottero if you want to continue on that.

Gnottero:
> Makes sense what James suggested, also use snake case, since its how the API is constructed. Let us know @Gnottero if you want to continue on that.

Yes, I'm still working on it, but I've been a little busy lately. At the moment I'm finishing implementing everything and fixing some bugs with some of the abilities

docs/scarpet/api/Entities.md
```diff
@@ -693,6 +717,34 @@ Applies status effect to the living entity. Takes several optional parameters, w
...
+
+Sets the value of the player's `instabuild` (The ability of the player to place blocks without reducing their count in the inventory and shoot with a bow without arrows) ability to the given boolean value.
+
+### `modify(e, 'fly_speed', float)`
+
+Sets the value of the player's `flySpeed` (The speed at which the player moves while flying) ability to the given float value.
+
+### `modify(e, 'walk_speed', float)`
+
+Sets the value of the player's `walkSpeed` (The speed at which the player moves while walking) ability to the given float value.
```
altrisi, line 742: This is just aesthetic and very personal opinion, but I think it would be nice to have those (and the queries) inside an `ability` or `abilities` "section" (so it would end up `modify(e, 'ability', 'walk_speed')`).

To do those I _think_ you just have to expect a `ListValue` inside the lambda of the `ability`/`abilities` map key instead of a single value, should just add a switch on string in there.

What do you think?

altrisi (commented):
Few small things, looks quite great to me

altrisi:
BTW you can "mark as resolved" the already resolved reviews so they don't clutter the PR.

docs/scarpet/api/Entities.md
Gnottero: I'll fix all these thing tomorrow. Thank you for the review!

gradle.properties
```diff
@@ -10,7 +10,7 @@ org.gradle.jvmargs=-Xmx1G
 	fabric_version=0.40.1+1.17
 
 # Mod Properties
-	mod_version = 1.4.52
+	mod_version = 1.4.53
```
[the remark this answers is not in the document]
gnembon: yeah, this would merge automatically. You can rebase/merge into your branch to update to the newer version

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -20,39 +20,41 @@
...
-import net.minecraft.entity.attribute.AttributeContainer;
-import net.minecraft.entity.attribute.EntityAttribute;
-import net.minecraft.entity.projectile.ProjectileEntity;
-import net.minecraft.entity.vehicle.AbstractMinecartEntity;
-import net.minecraft.nbt.NbtString;
-import net.minecraft.nbt.NbtElement;
-import net.minecraft.network.packet.s2c.play.EntityPassengersSetS2CPacket;
-import net.minecraft.network.packet.s2c.play.EntityPositionS2CPacket;
-import net.minecraft.network.packet.s2c.play.ExperienceBarUpdateS2CPacket;
-import net.minecraft.network.packet.s2c.play.PlayerPositionLookS2CPacket;
```
gnembon, line 38: I know its just imports, but refrain from resorting them mid double branch

docs/scarpet/api/Entities.md
gnembon: abilities is an internal thing. From the outsider, its yet another attributes of players in particular. Hence I don't think there is a need of grouping them under 'abilities'

Gnottero: So, is it ok as it is now, with a modify action for each of the abilities?

src/main/java/carpet/script/value/EntityValue.java
Gnottero: Uhm, what do you mean? I'm not using this imports in my PR

gradle.properties
```diff
@@ -10,7 +10,7 @@ org.gradle.jvmargs=-Xmx1G
 	fabric_version=0.40.1+1.17
 
 # Mod Properties
-	mod_version = 1.4.54
+	mod_version = 1.4.53
```
altrisi, line 13: This happened again, now backwards!

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -963,6 +1007,67 @@ private static void updateVelocity(Entity e, double scale)
...
+                player.getAbilities().creativeMode = instaBuild;
+                player.sendAbilitiesUpdate();
+            }
+        });
+
+        put("fly_speed", (e, v) -> {
+            Float flySpeed = NumericValue.asNumber(v).getFloat();
+            if (e instanceof ServerPlayerEntity player) {
+                player.getAbilities().setFlySpeed(flySpeed);
+                player.sendAbilitiesUpdate();
```
altrisi, line 1050: Any reason why you are using the boxed (uppercase F) float here and in the ones below? You should probably use primitive `float`.

docs/scarpet/api/Entities.md
```diff
@@ -400,6 +400,30 @@ query(p,'effect','resistance')  => null
 
 Number indicating remaining entity health, or `null` if not applicable.
 
+### `query(e, 'may_fly')`
+
+Boolean. Returns the value of player's `mayFly` ability.
```
altrisi, line 405: For consistency, around the docs there's usually not a type -> what, but a slightly more descriptive without saying "this is a float" or "this is an int" description. Mostly since in scarpet there's not really a differentiation between integer, float, double or long (at a language level, internally of course there's some).

For booleans to keep a bit of consistency with the rest of the docs you could keep it like "Returns a boolean indicating the value of the player's `flying` ability" for example but for numbers (floats in this case) don't specify the type, so maybe "Returns a number being the value of the player's `flySpeed` ability", though you don't need to specify it's a number if it's obvious, and (while in this case I don't really know how it could be replaced cleanly, so I'd keep it as is) it's better to not specify implementation details like "where does this exactly come from", mostly cause the api is expected to work even if the exact names/impl details change.

docs/scarpet/api/Entities.md
```diff
@@ -693,6 +717,34 @@ Applies status effect to the living entity. Takes several optional parameters, w
...
+
+Sets the value of the player's `mayBuild` (The ability of the player to place blocks) ability to the given boolean value.
+
+### `modify(e, 'insta_build', boolean)`
+
+Sets the value of the player's `instabuild` (The ability of the player to place blocks without reducing their count in the inventory and shoot with a bow without arrows) ability to the given boolean value.
+
+### `modify(e, 'fly_speed', float)`
+
+Sets the value of the player's `flySpeed` (The speed at which the player moves while flying) ability to the given float value.
```
altrisi, line 738: Same thing about specific number types around here, if you really want say to the given number.

altrisi (commented):
I think it's basically done, very minor things only, mostly about docs, and IMO with that is ready.

src/main/java/carpet/script/value/EntityValue.java
altrisi: This was from a previous commit.

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -947,7 +991,7 @@ private static void updateVelocity(Entity e, double scale)
 
     private static final Map<String, BiConsumer<Entity, Value>> featureModifiers = new HashMap<String, BiConsumer<Entity, Value>>() {{
         put("remove", (entity, value) -> entity.discard()); // using discard here - will see other options if valid
-        put("age", (e, v) -> e.age = Math.abs((int)NumericValue.asNumber(v).getLong()) );
+        put("age", (e, v) -> e.age = Math.abs((int)NumericValue.asNumber(v).getLong()));
```
gnembon, line 994:
```suggestion
        put("age", (e, v) -> e.age = Math.abs((int)NumericValue.asNumber(v).getLong()) );
```
```suggestion
        put("age", (e, v) -> e.age = Math.abs((int)NumericValue.asNumber(v).getLong()));
```

gradle.properties
gnembon, line 13:
```suggestion
	mod_version = 1.4.54
```

gnembon (approved):

ch-yx:
`modify(e, 'selected_slot', int)` cannot work if e is a fake player. 

but i can use `run('player '+fakeplayer~'command_name'+' hotbar '+slot);` to replace it.

PR #1168, opened 2021-11-14
Adding min_y, max_y and world_border_info to system_info

Ghoulboy78:
Adds `world_min_y` and `world_max_y` to `system_info`. Rn only works in caller dimension, though I bet there's a way to get around that.

### DIFF
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -667,6 +667,8 @@ Available options in the scarpet app space:
   * `world_gamerules` - returns all gamerules in a map form (`rule`->`value`). Like carpet rules, values are returned as strings, so you can use appropriate value conversions using `bool()` or `number()` to convert them to other values. Gamerules are read-only to discourage app programmers to mess up with the settings intentionally applied by server admins. Isn't that just super annoying when a datapack messes up with your gamerule settings? It is still possible to change them though using `run('gamerule ...`.
   * `world_spawn_point` - world spawn point
   * `world_time` - Returns dimension-specific tick counter.
+  * `world_max_y` - Returns maximum y-value in caller dimension. For other dimensions run `in_dimension('dimension', system_info('world_max_y')))`.
+  * `world_min_y` - Same as above, returning minimum y-value.
 
  Relevant gameplay related properties
   * `game_difficulty` - current difficulty of the game: `'peaceful'`, `'easy'`, `'normal'`, or `'hard'`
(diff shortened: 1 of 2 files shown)


Ghoulboy78:
After a bit of digging, I found the right method for the task, and now `system_info('world_max_y')` returns a map of each dimension to its maximum y value (and same for `world_min_y`.

I also went ahead and added the ability to query world borders, cos it went with the theme. I didn't add the ability to modify world borders, as that's a lot of power, and you can do it with `run(...)`. Similar idea to gamerule modification.

altrisi:
-1 on no worldborder modification, it could be quite useful if you are developing a game, and I don't think it would be abused like gamerules could.

-0 on returning a map of dimensions mapped to heights instead of just the one for the world, I think it should just return the one from the current dimension, you can always use `in_dimension` if you need another one. That'd make the call less expensive and more suitable for inline use (the dimension thing probably documented though).

Also is it really `getLocalHeight()`? May be just that the mapping aren't perfect/aren't updated, but it feels weird to me that it is called that way.

Ghoulboy78:
On wolrdborder modification, sure I can add that, if gnembon approves.

On map of dimensions, this is calculated ahead of time, and stored, so it just returns it when u ask for it. I know, cos I had the same concern for a little while.

And ye I was also confused when IDE suggested `getLogicalHeight` but apparently it calls a method called `getMaximumHieght` or something under the hood, so I went with it.

This is still due for a full-scale test ofc,

altrisi:
System info isn't cached, the code inside the lambda is executed every time. You can check because the code does get on the map and apply every time. Else things like server_last_tick_times, app_list, java_used_memory and others like those wouldn't work.

Ghoulboy78:
well, im dum then arent I. Ill revert to old code then

src/main/java/carpet/script/utils/SystemInfo.java
```diff
@@ -60,6 +64,38 @@
             WorldProperties prop = c.s.getServer().getOverworld().getLevelProperties();
             return ListValue.of(NumericValue.of(prop.getSpawnX()), NumericValue.of(prop.getSpawnY()), NumericValue.of(prop.getSpawnZ()));
         });
+
+        put("world_min_height", c-> {
```
gnembon, line 68: tbh, more practical and simpler would be to return that info for current dimension. Since prefix `world` means current save, regardless of the dimension, maybe you could use `dimension_min_height`, `dimension_max_height`, `dimension_center`, `dimension_radius` (I would separate them)

docs/scarpet/api/Auxiliary.md
```diff
@@ -667,6 +667,9 @@ Available options in the scarpet app space:
   * `world_gamerules` - returns all gamerules in a map form (`rule`->`value`). Like carpet rules, values are returned as strings, so you can use appropriate value conversions using `bool()` or `number()` to convert them to other values. Gamerules are read-only to discourage app programmers to mess up with the settings intentionally applied by server admins. Isn't that just super annoying when a datapack messes up with your gamerule settings? It is still possible to change them though using `run('gamerule ...`.
   * `world_spawn_point` - world spawn point
```
gnembon, line 668:
```suggestion
  * `world_spawn_point` - world spawn point in the overworld dimension
```
```suggestion
  * `world_spawn_point` - world spawn point
```

docs/scarpet/api/Auxiliary.md
```diff
@@ -667,6 +667,9 @@ Available options in the scarpet app space:
...
   * `world_time` - Returns dimension-specific tick counter.
```
gnembon, line 669: this one is fine - I think all dimensions share the same counter anyways, so its not dimension specific

gnembon (approved):

PR #1176, opened 2021-11-20
Multi noise support for scarpet

Crec0:
With the release of 1.18 and multi noise, scarpet should have the ability to access the multi noise values and use them for specific things like biome look up.

Thank you @Starmute for coming up with the idea.

- [x] Add function `query_noise(pos, type?)` to scarpet to get noise values
- [x] Add multinoise biome look up support as `biome(noiseMap)`
- [x] Add documentation for new functions


src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -93,14 +90,17 @@
...
+import net.minecraft.world.biome.source.util.MultiNoiseUtil;
 import net.minecraft.world.chunk.Chunk;
 import net.minecraft.world.chunk.ChunkStatus;
 import net.minecraft.world.chunk.WorldChunk;
 import net.minecraft.world.explosion.Explosion;
-import net.minecraft.world.gen.random.ChunkRandom;
 import net.minecraft.world.gen.chunk.ChunkGenerator;
 import net.minecraft.world.gen.feature.ConfiguredStructureFeature;
 import net.minecraft.world.gen.feature.StructureFeature;
+import net.minecraft.world.gen.random.ChunkRandom;
```
gnembon, line 103: Please, stop your IDE to sort imports

gnembon: and reformat files for no particular reason

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -1510,6 +1458,24 @@ else if (ticket == ChunkTicketType.POST_TELEPORT) // post teleport
                 cc.s.getWorld().getChunkManager().addTicket(ChunkTicketType.UNKNOWN, target, radius, target);
             return new NumericValue(ticket.getExpiryTicks());
         });
+    }
+
+    @ScarpetFunction(maxParams = -1)
+    public Value query_noise(Context c, int x, int y, int z, String... noiseQueries) {
```
gnembon, line 1464: Not sure this gives you ability to run query_noise(block), or you always need to provide x, y and z

altrisi: Using `@Locator.Block BlockPos block` instead of the three ints should allow that.

src/main/java/carpet/script/annotation/AnnotationParser.java
```diff
@@ -13,19 +25,6 @@
 import java.util.Optional;
 import java.util.stream.Collectors;
 
-import org.apache.commons.lang3.ArrayUtils;
-import org.apache.commons.lang3.ClassUtils;
-
-import carpet.CarpetExtension;
-import carpet.script.Context;
```
gnembon, line 20: This is unnecessary, right?

src/main/java/carpet/mixins/NoiseColumnSamplerMixin_scarpetMixin.java
```diff
@@ -0,0 +1,235 @@
...
+    public abstract MultiNoiseUtil.NoiseValuePoint sample(int i, int j, int k);
+
+    @Shadow
+    public abstract TerrainNoisePoint createTerrainNoisePoint(int x, int z, float continentalness, float weirdness, float erosion, Blender blender);
+
+    @Override
+    public double getNoiseSample(String name, int x, int y, int z) {
+        MultiNoiseUtil.NoiseValuePoint noiseValuePoint = this.sample(x, y, z);
+        switch (name) {
+            case "temperature" -> {
```
gnembon, line 118: not sure what's better, a giant switch, or a string keyed map with lambda calls.

Crec0:
Did all the review changes. Sorry about the reformatting and yes I now disabled auto import formatting.

src/main/java/carpet/script/api/WorldAccess.java
Crec0: I replaced it with altisi's suggestion, doing testing rn

src/main/java/carpet/mixins/NoiseColumnSamplerMixin_scarpetMixin.java
Crec0: I just find switch much more readable. Is there a performance difference between using map over this giant switch ? If so I can switch it over.

Ghoulboy78: ig using a map would allow you to move it about where you're gonna use it more, whereas with a switch it's only here in a mixin, where people may not instinctively go to find it. Plus the mixin folder is huge, so finding the right one could be a pain.

Crec0: true. I will give it a go

Crec0:
Documentation has been added

gnembon (approved):

PR #1186, opened 2021-11-26
hacky make carpet command permisison level configurable

adryd325:
quick and dirty way of making carpet command permission level configurable.

### DIFF
--- a/src/main/java/carpet/settings/SettingsManager.java
+++ b/src/main/java/carpet/settings/SettingsManager.java
@@ -72,6 +72,7 @@ public class SettingsManager
      * correctly react to changes.
      */
     public boolean locked;
+    private String carpetCommandPermissionLevel = "ops";
     private final String version;
     private final String identifier;
     private final String fancyName;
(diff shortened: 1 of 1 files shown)


gnembon:
why that would ever be useful?

adryd325:
I have a creative server where "admins" are op level 4 and normal players are op level 2 (to have access to /gamemode, /tp, etc). I don't want normal players messing with the carpet config but i'd like op level 4 to still be able to.

gnembon:
Ok, I get that. Well, settings.conf is writable by the /carpet command and your line will disappear so your method is very dirty.

You have to take care to write it back to the conf in case someone changes it. locked configuration cannot be changed, so that's why its not saved back.

altrisi:
Why not just make it a regular carpet rule though?

gnembon:
true, it could be just a read only rule so one that cannot be changed in game. That would be safest

adryd325:
How would I make a read only rule, I don't see anything that would specify a rule is read only. Would it be given a Validator that always returns null?

gnembon:
something like that - making sure it can be set on startup

PR #1202, opened 2021-12-06
[Scarpet] Task value refactor for better extendability

replaceitem:
This refactors the constructor of `ThreadValue`, in order to allow extending it in extensions more easily. As an example, in Discarpet I am hoping to use scarpet tasks as the return value of functions that run API calls, and as such, they take a while to finish and return a value. (https://github.com/replaceitem/carpet-discarpet/issues/18#issuecomment-986299250) However, the current constructor only allows for the creation of tasks from `FunctionValue`s.

### DIFF
--- a/src/main/java/carpet/script/value/ThreadValue.java
+++ b/src/main/java/carpet/script/value/ThreadValue.java
@@ -19,15 +19,15 @@ public class ThreadValue extends Value
     private final long id;
     private static long sequence = 0L;
 
-    protected ThreadValue(CompletableFuture<Value> taskFuture, long id)
+    protected ThreadValue(CompletableFuture<Value> taskFuture)
     {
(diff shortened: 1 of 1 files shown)


replaceitem:
Update: I found a better solution to my problem, but i will leave this here since I think it still makes sense for extensions.

gnembon:
Question - why would you want to provide own id?

replaceitem:
> Question - why would you want to provide own id?

Hmm good question, i just made both fields assignable in the contructor, but i agree, `id` doesn't make sense in this case.

altrisi:
Why only allow subclasses to use its own CompletableFuture though? I don't think it would be a problem to have a public constructor that allows an arbitrary completablefuture without needing to subclass.

gnembon (approved):

Issue #1200, opened 2021-12-05
Suggestion: Add a command for changing simulation distance


mangotango69202:
Like the command for changing view distance without restarting. Add it for simulation distance

gnembon:
makes sense

PR #1211, opened 2021-12-11
Feature: simulationDistance rule

ishland:
Closes gnembon/fabric-carpet#1200

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -682,6 +682,47 @@ public class CarpetSettings
     )
     public static int viewDistance = 0;
 
+    private static class SimulationDistanceValidator extends Validator<Integer>
+    {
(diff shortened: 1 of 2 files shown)


src/main/java/carpet/CarpetSettings.java
```diff
@@ -682,6 +682,47 @@ public String description() {
...
+            if (newValue < 0 || newValue > 32)
+            {
+                Messenger.m(source, "r simulation distance has to be between 0 and 32");
+                return null;
+            }
+            MinecraftServer server = source.getServer();
+
+            if (server.isDedicated())
+            {
+                int vd = (newValue >= 2)?newValue:((DedicatedServer) server).getProperties().simulationDistance;
```
altrisi, line 702: Is the minimum simulation distance 2? This minimum could also probably be moved above so setting it to values in the middle doesn't silently ignore the user's choice.

ishland: I just copied and pasted the implementation for viewDistance and that should also happen for viewDistance as well.

PR #1215, opened 2021-12-16
Fire the scarpet event when changing hotbar slots via the action pack

altrisi:
Simply that.

### DIFF
--- a/src/main/java/carpet/helpers/EntityPlayerActionPack.java
+++ b/src/main/java/carpet/helpers/EntityPlayerActionPack.java
@@ -1,6 +1,7 @@
 package carpet.helpers;
 
 import carpet.fakes.ServerPlayerEntityInterface;
+import carpet.script.CarpetEventServer;
 import net.minecraft.block.BlockState;
 import net.minecraft.command.argument.EntityAnchorArgumentType;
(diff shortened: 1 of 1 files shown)


Ghoulboy78:
Are other events triggered? Like ik that this one was the one mentioned in discord, but it's work assuring that other events are also triggered.

altrisi:
The rest of events (most at least, didn't test them) are not triggered, but that's another discussion, mostly because they are a lot more frequent and (imo) are less likely to be actually needed for fakes in most cases at least.

Feel free to continue the discussion about the rest in the issue or maybe there could be a different way for your use case, like being able to query the actions from a fake player.

Up to gnembon to decide if this should get in alone or not.

altrisi:
Hmmm thinking about it again, maybe yeah this shouldn't work alone...

gnembon:
Yeah, I always thought that events on players would work on the vanilla stuff, while fake players is just a 'modded gimmick' and doesn't require to be supported at all - its like it came from some random other mod and the mod author would complain that their actions are not registered.

But in general I agree - it would be 'better' is all actions would respond even for 'modded' players.

altrisi:
Closing in favour of #1490.

PR #1216, opened 2021-12-16
Fix caveEntrance sampling being incorrect (please check)

Starmute:
hi - caveEntrance noise was previously being checked incorrectly (it was scaled up by 4x when it should actually just use world pos.) This should fix it, but I haven't tested it (to verify, stand in a cave entrance and do /script run sample_noise(x,y,z,'caveEntrance'). If it returns less than 0, the value is accurate.


Crec0 (approved):
Looks good

gnembon (approved):

PR #1218, opened 2021-12-17
Reinstating old behaviour of += with lists

### LINKED ISSUE #1152: Shapes.scl broken by new method of adding lists and sets via +=

Ghoulboy78:
Properly fixes #1152, as opposed to the hack-fix in #1154 which just works around it.
In ye olden days, when the += operator simply called `Value.add()`, this is the behaviour that would occur, i.e.: doing `l=[]; l+=['3','4']`, would result in `l==['3','4']`, whereas now it results in `l==[['3','4']]` . This broke `shapes.scl` accidentally. I have also fixed this behaviour for maps, cos it's also broken there.

### DIFF
--- a/src/main/java/carpet/script/value/ListValue.java
+++ b/src/main/java/carpet/script/value/ListValue.java
@@ -147,7 +147,10 @@ public class ListValue extends AbstractListValue implements ContainerValueInterf
     @Override
     public void append(Value v)
     {
-        items.add(v);
+        if(v instanceof AbstractListValue)
+            items.addAll(((AbstractListValue) v).unpack())
+        else
+            items.add(v);
     }
 
     @Override
--- a/src/main/java/carpet/script/value/ListValue.java
+++ b/src/main/java/carpet/script/value/ListValue.java
@@ -148,7 +148,7 @@ public class ListValue extends AbstractListValue implements ContainerValueInterf
     public void append(Value v)
     {
         if(v instanceof AbstractListValue)
-            items.addAll(((AbstractListValue) v).unpack())
+            items.addAll(((AbstractListValue) v).unpack());
         else
             items.add(v);
     }
--- a/src/main/java/carpet/script/value/MapValue.java
+++ b/src/main/java/carpet/script/value/MapValue.java
@@ -162,7 +162,13 @@ public class MapValue extends AbstractListValue implements ContainerValueInterfa
     @Override
     public void append(Value v)
     {
-        map.put(v, Value.NULL);
+        if (v instanceof MapValue)
+            map.putAll(((MapValue) v).getMap());
+        else if (v instanceof AbstractListValue)
+            for (Value value : ((AbstractListValue) v))
+                map.put(value, Value.NULL);
+        else
+            map.put(v, Value.NULL);
     }
 
     @Override


ch-yx:
Oh no. 🤦‍
Do you think the semantics of += are not ambiguous enough?

Firigion:
I think that if you want to use `+=` on lists to mean `lhs = lhs + rhs`, you need to also add an easy way to append into a list. `+=` was added as append because  `put(my_list, null, element)` is way to verbose for sucha a common action. I think `+=` functioning as append is a fantastic addition and old code that used the fact that adding lists of the same length wolud add them element-wise just need to switch their `l += [1,2]` to `l = l+[1,2]`, which is not that bad.

Also, `l=[]; l+=['3','4']` returning `l==['3','4']` also doesn't make much sense. If addition between lists is meant to add element-wise, then this should error out. If it's meant to _extend_ a list (append all the contents of the right hand side list into the end of the left hand side one), this is gonna silently fail when the two lists are the same length. I'd strongly advice against removing element-wise operations of lists, since we dont' ahve a better data type for that.

Basically, this bolis down to what is more commonly used: extend or append. One should keep the `+=` action and the other one stick to `put(list, null, value, mode)`. My vote is for append to keep it.

Ghoulboy78:
Ok, well what I did was make it so if you do `l += a_list_or_map`, the elements of `a_list_or_map` (for a map it's the key-value pairs in a list, like you used to declare them in the olden days) get added one by one to the end of list `l`. For anything else though, doing `l+=v` extends the list and sticks `v` to the end of `l`. But you can still do element-wise addition with `l = l+a_list` (not with a map I don't think). Basically I didn't touch `+` operator, just `+=` for lists, which is called via the `append()` method in `AbstractListValue` class.

Ghoulboy78:
But you see Firi, I'm reinstating old behaviour which was accidentally removed earlier on, and no one noticed that it broke stuff.

Firigion:
I know, but I don't like it, and I assume gnemon changed it on purpose :P

Still, if you are making a PR reisntating the old behaviour, you can use the opportunity to clarify the thing to new user with a good documentation, because the behaviour is anything but trivial.

Ghoulboy78:
I made a discord thread to discuss it

gnembon:
Yeah - current behaviour is less confusing then proposed. 'put' has clear usecases for 'insert' and 'extend'. Plus that would break a few examples that are currently in the docs and work because += assume the value to add is always a scalar.

PR #1233, opened 2021-12-23
Remove querying full `system_info()` and refactor `Monitoring` to use the annotation API

### LINKED ISSUE #1085: [Scarpet RFC] Should `system_info()` without params return only a list of possible params instead of getting all info?
`system_info()` has grown quite a bit since it's introduction, and functions inside have been getting more expensive.

At the moment running the entire `system_info()` can be relatively expensive depending on the environment, since it will generate a map with _all_ of the `system_info()` available parameters mapped to the result of their operations.

Most of the time, you don't need to get the entirety of `system_info` in one go, so the obvious choice is to call `system_info(requested)`, with a parameter. However, people may end up using `system_info():requested` for whatever reason, creating expensive calls. I'd assume mostly because of not knowing about the better alternative/the downsides, but whatever.

Therefore I'm suggesting replacing the return value of `system_info()` without parameters to a list of the possible parameters, without executing their responses.

Another option is removing the parameter-less response completely, since (IMO) it's unlikely you'll be checking that programmatically, and if you are running it to see the list yourself, you are better off just checking the docs, that will have more information.

altrisi:
Resolves #1085 by removing the no-arg version. I couldn't find any use of `system_info()` on the app store or in any of the other scarpet github repos that I know or found.

Also refactors the `Monitoring` class to use the annotation API, because I was feeling like migrating it and else there was just a single function there using the old system (since I moved system_info). Basically moves `get_mob_counts` to use it, no functional or API changes around it. And it's actually pretty much the same, just with some types already known.

And removes a no-longer needed accessor around it.

### DIFF
--- a/src/main/java/carpet/script/api/Monitoring.java
+++ b/src/main/java/carpet/script/api/Monitoring.java
@@ -37,7 +37,7 @@ public class Monitoring {
         if (category.isPresent())
         {
             SpawnGroup group = SpawnGroup.byName(category.get().toLowerCase(Locale.ROOT));
-            if (group == null) throw new InternalExpressionException("Unreconized mob category: " + category.get());
+            if (group == null) throw new InternalExpressionException("Unrecognized mob category: " + category.get());
             return ListValue.of(
                     new NumericValue(mobcounts.getInt(group)),
                     new NumericValue((int)(group.getCapacity() * chunks / SpawnReporter.currentMagicNumber()))
--- a/src/main/java/carpet/script/api/Monitoring.java
+++ b/src/main/java/carpet/script/api/Monitoring.java
@@ -1,8 +1,8 @@
 package carpet.script.api;
 
-import carpet.fakes.SpawnHelperInnerInterface;
 import carpet.script.CarpetContext;
-import carpet.script.Expression;
+import carpet.script.Context;
+import carpet.script.annotation.ScarpetFunction;
 import carpet.script.exception.InternalExpressionException;
 import carpet.script.utils.SystemInfo;
 import carpet.script.value.ListValue;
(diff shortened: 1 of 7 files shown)


src/main/java/carpet/script/api/Monitoring.java
```diff
@@ -13,62 +13,50 @@
...
-                Value res = SystemInfo.get(what, (CarpetContext) c);
-                if (res == null) throw new InternalExpressionException("Unknown option for 'system_info': " + what);
-                return res;
-            }
-            throw new InternalExpressionException("'system_info' requires one or no parameters");
-        });
-        // game processed snooper functions
-        expression.addContextFunction("get_mob_counts", -1, (c, t, lv) ->
+            SpawnGroup group = SpawnGroup.byName(category.get().toLowerCase(Locale.ROOT));
+            if (group == null) throw new InternalExpressionException("Unreconized mob category: " + category.get());
```
James103, line 40:
```suggestion
            if (group == null) throw new InternalExpressionException("Unrecognized mob category: " + category.get());
```

altrisi: Fixed, weird it wasn't noticed yet, since it was in the original implementation.

gnembon:
Can you reverse running these functions through annotation parser - it adds an overhead that could only be acceptable in more complex functions / these that would take complex parameters to simplify its implementation. system_info is not one of them. running system_info('scarpet_version') for example runs twice slower with the annotation version according to my profiling.

altrisi:
Will do.

I'll also try to try and make the annotation system faster at some point, because I remember at the very beginning it was quite fast, but I may have changed something at some point that made it slower (maybe one of the last transformations of method handles, will check, or just try to find faster ways).

It'll be tough to compete against `system_info('scarpet_version')`, given that's pretty much only a map and variable lookup, so even almost nothing is likely going to make it slower, but I'll try my best. If the annotation system is only as slow as it I'm personally quite happy already tbh.

PR #1234, opened 2021-12-23
Give feedback when stepping ticks

### LINKED ISSUE #1232: [Easy feature request] Print /tick step executions
`/tick step` is a great feature for redstone debugging, but the command runs silently (no print output in chat). A very easy quality of life improvement would be to simply print out the number of ticks stepped in chat everytime you run the command.

Logging tick steps is helpful because it helps debuggers keep track of how many ticks they have stepped through since the beginning of the debug session.

For example, if running `/tick step 2`, could print out `[Timestamp] Stepped 2 game ticks`.

> Printing out the current timestamps would also be helpful in determining whether the log is associated with the current debug session.

altrisi:
Fixes #1232.

### DIFF
--- a/src/main/java/carpet/commands/TickCommand.java
+++ b/src/main/java/carpet/commands/TickCommand.java
@@ -51,10 +51,10 @@ public class TickCommand
                             then(literal("deep").executes( (c)-> setFreeze(c.getSource(), true, true)))).
                         then(literal("off").executes( (c) -> setFreeze(c.getSource(), false, false)))).
                 then(literal("step").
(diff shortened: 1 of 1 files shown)


aigagror:
Can we merge this now?

altrisi:
It'll be merged when it's reviewed, gnembon is busy atm. You can always download the PR artifacts in the [actions tab of this PR](https://github.com/gnembon/fabric-carpet/actions/runs/1616857984) so you can use it before release.

gnembon (approved):

PR #1235, opened 2021-12-24
Make fake players compatible with adventure platform lib and minor refactor around them

### LINKED ISSUE #1142: Carpet Players Incompatibility with Floodgate and Geyser
Spawning of carpet players does not work with Floodgate Fabric and Geyser Fabric.

### LINKED ISSUE #1205: Can't now use /player <name> spawn in version 1.18
IMG_20211207_142214

altrisi:
Fixes #1142, fixes #1205 (dupe), fixes GeyserMC/Floodgate-Fabric#37.

This PR adds an `EmbeddedChannel` to fake player's `ClientConnection`, in order to allow adventure platform (fabric) to not npe when using it, that was effectively disabling fake players (no crash because commands catch the exception and it's just `unexpected error`).

Adventure platform is used by many mods, even more lately, including but not limited to GeyserMC's floodgate (bedrock players on java, very used), blockbot or tabtps.

~~While the way of fixing it seems harmful (creating a `ChannelHandlerContext` that provides invalid returns for everything but `Channel`), I've checked and minecraft doesn't use anything else from it than the `Channel`, neither does any of those mods. Just in case I added a toggle that will automatically disable calling that if it fails once, but it should never happen (it was just a lot of methods and I felt like I had to, can be removed if not wanted).~~ Outdated

While I was at it, refactored the way fake players handle the duplicate login disconnection to use the already implemented `disconnect` method in their `ServerPlayNetworkHandler` instead of mixing in to cancel and redo an entire `for` loop. Also renamed `NetworkManagerFake` to `FakeClientConnection`, I guess the old one was based on different mappings, it was confusing.


EvilSquirrelGuy:
Would this be compatible with 1.18.1 with loader 0.12.12? can't seem to get it to compile
(also if it's any help, the `future` branch houses 1.18 development builds)

altrisi:
With loader 0.12.12 for sure since that's what I've been using in development for a few days. And IIRC I actually tested this on the future branch since I was unable to connect to my testing server with a BE client on 1.17. You may have merged it incorrectly or something.

I've uploaded the 1.18 jar I tested and worked here: https://github.com/altrisi/fabric-carpet/releases/tag/hahafunnytagthatisntevencorrect (note that it's not supported and I can't actually remember if it had anything else than geyser compat, shouldn't be harmful though).

altrisi:
Converted to a draft because I still don't feel right about the future-proofness of this implementation, will rethink it when I have a bit of time (likely tomorrow).

gnembon:
The removed mixin could be tested in a server situation where a player shadows itself, getting kicked properly and then reconnects properly kicking out the bot and regaining control. As far as I remember all of that was to properly handle shadowed players.

altrisi:
Changed to a cleaner approach.

@gnembon that mixin isn't necessary anymore, it was in the past because fake players didn't have their own ServerPlayNetworkHandler, if you look at it, what happens without the mixin is that it calls ServerPlayNetworkHandler#disconnect, and given that's overridden for fakes, the kill method is invoked from there.

Also tested it ofc and it works.

For anyone interested in testing the fix: https://github.com/altrisi/fabric-carpet/releases/tag/adventure-take-two

altrisi:
> That mixin isn't necessary anymore, it was in the past because fake players didn't have their own ServerPlayNetworkHandler, if you look at it, what happens without the mixin is that it calls ServerPlayNetworkHandler#disconnect, and given that's overridden for fakes, the kill method is invoked from there.
> 
> Also tested it ofc and it works.

Removed another no longer needed mixin, this time the one to make enderpearls work with fake players, given they now have a channel and therefore their ClientConnection isOpen returns true by itself.

altrisi:
This also fixes compatibility of relogging after shadow with some mod (likely cardboard) that seems to not use the old redirected path, that is now specific in the fake player implementation.

According to an user on discord (https://discord.com/channels/882822986795716608/882825359517712436/932724438795948104).

gnembon (approved):

PR #1242, opened 2021-12-25
Support more actions in Messenger (and therefore `format()`)

### LINKED ISSUE #1182: [Scarpet Suggestion] Support for clickable links in format
It would be useful if `format` could create a text fragment that when clicked would open an url.

Suggested prefix (feel free to suggest a better one): `@`

altrisi:
Resolves #1182.

This PR adds the following two actions to `Messenger` (and therefore to the scarpet's `format()` function):
- `'@<url>'` Opens the given url when clicking the message
- `'&<text>'` Copies the given text to clipboard when clicking the message

Also changed it to use a switch statement and made some small changes to conventions in the rest of Messenger (where they didn't affect api).

### DIFF
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -252,6 +252,8 @@ Decorators (listed as extra argument after the component they would affect):
  * `'^<format> <text>'` - hover over tooltip text, appearing when hovering with your mouse over the text below.
  * `'?<suggestion>` - command suggestion - a message that will be pasted to chat when text below it is clicked.
  * `'!<message>'` - a chat message that will be executed when the text below it is clicked.
+ * `'@<url>'` - an url will be opened when the text below it is clicked.
+ * `'&<text>'` - a text will be copied to clipboard when the text below it is clicked.
  
 Both suggestions and messages can contain a command, which will be executed as a player that clicks it.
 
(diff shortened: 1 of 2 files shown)


docs/scarpet/api/Auxiliary.md
```diff
@@ -252,6 +252,8 @@ Decorators (listed as extra argument after the component they would affect):
  * `'^<format> <text>'` - hover over tooltip text, appearing when hovering with your mouse over the text below.
  * `'?<suggestion>` - command suggestion - a message that will be pasted to chat when text below it is clicked.
  * `'!<message>'` - a chat message that will be executed when the text below it is clicked.
+ * `'@<url>'` - an url will be opened when the text below it is clicked.
+ * `'&<text>'` - a text will be copied to clipboard when the text below it is clicked.
```
James103, line 256:
```suggestion
 * `'@<url>'` - a URL will be opened when the text below it is clicked.
 * `'&<text>'` - some text will be copied to clipboard when the text below it is clicked.
```
Use "a" instead of "an" before "URL" ([Writing Explained](https://writingexplained.org/a-vs-an-difference)).
Capitalize "URL" because it is an acronym.
Use "some text" instead of "a text" as "text" is not necessarily atomic.

altrisi: I'll do the first two, not doing the third one because it's referring to the text given in the parameter, and that's the way it's mentioned in the rest of options (I'll add a `that`, because I forgot it in both).

gnembon (approved):

PR #1245, opened 2021-12-27
Fixing IDEA pictures 

Ghoulboy78:
Basically the pictures for setting up scarpet language in IDEA were outdated


Ghoulboy78:
What it used to look like:
image

What it looks like now:
image

Ghoulboy78:
Ik that the pictures are blurry, but when you open them in the page they're all blown up so they look less blurry.

Issue #1248, opened 2021-12-27
【suggestion】 query WorldGenSettings


ch-yx:
system_info('WorldGenSettings') return a nbt value object of WorldGenSettings from level.dat.
![图片](https://user-images.githubusercontent.com/45825511/147483308-791c710a-8969-4563-bc4d-774c69a4f5d3.png)

Ghoulboy78:
We could add these all to `system_info`, but add them separately, cos we have the border size already, as well as the seed, and like the dragon fight is useful but tbh could be put separately.

ch-yx:
but.... border size&dragon fight are not in that tag......
and...i just looks at the code. there is a method already to store them as a NBT tag in that file.
i think i could use it directly.
needlessly to remove 'seed' from that tag.
just leave it here

altrisi:
Take account that it's unlikely that the game stores them as an NBT tag, so having returning them together as an nbt value is likely more work than just giving individual values.

Plus please no nbt if not needed!

ch-yx:
```
        put("world_gen_settings", c-> {
            GeneratorOptions generatorOptions=c.s.getWorld().getServer().getSaveProperties().getGeneratorOptions();
            Optional<NbtElement> x = GeneratorOptions.CODEC.encodeStart(NbtOps.INSTANCE, generatorOptions).result();
            if(x.isPresent()){
                return new NBTSerializableValue(x.get());
            }else{
                return NullValue.NULL;
            }
        });
```
i added that to system_info code.
seems works.
here are 2 samples.(`/script run write_file('aqq','shared_nbt',system_info('world_gen_settings'))`)
[aqq.zip](https://github.com/gnembon/fabric-carpet/files/7784609/aqq.zip)

i think that that result will satisfiy any users needs.
should i make a pr?

ch-yx:
I fully understand that (as many have reminded me) generating nbt is expensive.
But it is really hard for me to imagine any other way to represent this data. After all, the data itself is a complex structure.

ch-yx:
Not only that, the parameters of various types of dimensional types, biome types, and terrain generation types vary tremendously. They are also able to be combined with each other freely. When dealing with such a complex data, to implement it by yourself  without errors is difficult.

ch-yx:
As altrisi said, data provide by scarpet should be better not simply put "returns the value of the tag X in the level file" in the docs.
But I can't think of a good way to carefully break this data down into a series of smaller ones.

[...]

PR #1253, opened 2021-12-30
Improve (about issues#1248)

ch-yx:
https://github.com/gnembon/fabric-carpet/issues/1248

this pr is to add some wor gen related systeminfo.

layers_of_super_flat_world should return a map value. which includes the layers of all dimensions if that dimension is a super flat world.     eg: [[1,bedrock],[3,stone],[52,sandstone]]
for dimensions that are not superflat, it may return a string value says why it is not layered XD.

...want to add more methods here......

world_gen_settings should be view as a 'ultimate solution', only consider using it if the information provided by other methods are not enough. may change in the future ver. of minecraft.

### DIFF
--- a/src/main/java/carpet/script/utils/SystemInfo.java
+++ b/src/main/java/carpet/script/utils/SystemInfo.java
@@ -5,8 +5,11 @@ import carpet.CarpetSettings;
 import carpet.script.CarpetContext;
 import carpet.script.CarpetScriptHost;
 import carpet.script.value.BooleanValue;
+import carpet.script.value.BlockValue;
 import carpet.script.value.ListValue;
 import carpet.script.value.MapValue;
+import carpet.script.value.NBTSerializableValue;
+import carpet.script.value.NullValue;
 import carpet.script.value.NumericValue;
 import carpet.script.value.StringValue;
 import carpet.script.value.Value;
@@ -17,6 +20,8 @@ import com.sun.management.OperatingSystemMXBean;
 import net.fabricmc.loader.api.FabricLoader;
 import net.fabricmc.loader.api.ModContainer;
 import net.minecraft.SharedConstants;
+import net.minecraft.nbt.NbtElement;
+import net.minecraft.nbt.NbtOps;
 import net.minecraft.server.world.ServerWorld;
 import net.minecraft.util.WorldSavePath;
 import net.minecraft.util.registry.RegistryKey;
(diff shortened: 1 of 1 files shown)


ch-yx:
cannot come up with any new ideas ...

Ghoulboy78:
You should add it to the docs for system_info. They are in Auxiliary.md

gnembon:
serialized wg settings is a useful thing for sure

layers_of_super_flat_world - should be named differnetly and more generically. Please find a more generic name for it. Besides - the info should only contain information about the current dimension of the invoker, to be manipulated in the script with `in_dimension(..., system_info(...))`

ch-yx:
> layers_of_super_flat_world - should be named differnetly and more generically.


Yes. i m to change it in to `super_flat_layers`



> the info should only contain information about the current dimension of the invoker

dont think so... the name `system_info` suggests that it should be identical in the whole server. and i think to query one item from the map like this is really handy in scarpet (it can reduce one layer of parentheses nesting!!).

ch-yx:
🆗. i changed the name.
but i m not willing to change the other thing.

gnembon:
first is that 'noise' or 'debug' has nothing to do with layers. Something like 'world_type_config' or something more generic would fit better, don't you think?

Regarding world-specific output - its already the case for some other elements of system info, where pulling info for all dimensions would be impractical, but I can get that here you would not query it that often.

Please resolve merge conflicts as well.

ch-yx:
> first is that 'noise' or 'debug' has nothing to do with layers. Something like 'world_type_config' or something more generic would fit better, don't you think?

yes i think that i should rethink it.
i designed it like that because I, as I said in the doc , did not guarantee the result string of the non-flat world..... i only guarantee that it must be a string and not a list. and it is just a friendly reminder about why it not flat. And users can use `type()` to know if it is superflat easily.
and i was thinking about provide more info for noise worlds by another function such as xxx_of_noise_world then.... and in that function, superflat worlds will only get a string said `superflat` 😅

maybe i **should** guarantee them as your said and "upgrade" that function and change it name into your said 'world_type_config' . but i cannot make up my mind to do so because this will block the roads to add more information about noise worlds in the future.

gnembon:
well, another way to approach it is to have two calls. one for `world_config_type` returning a string including 'superflat' and another `world_superflat_layers` returning null for all other types and returning layers for the dimension if its superflat. But in this case specifically you would need to target current dimension otherwise it all gets unnecessarily wrapped around dimension map to me.

ch-yx:
> well, another way to approach it is to have two calls. one for `world_config_type` returning a string including 'superflat' and another `world_superflat_layers` returning null for all other types and returning layers for the dimension if its superflat. But in this case specifically you would need to target current dimension otherwise it all gets unnecessarily wrapped around dimension map to me.

🆗. now it only shows current dims superflat info and i add a new function as you said.

and it seems that the `world_gen_settings` works non-stably. sometimes it simply returns `null`. dont know why. need help.

gnembon:
please update this PR to the new mappings and fix the imports

PR #1254, opened 2021-12-31
Shulker box stack size is now configurable

TomB-134:
This is to help balance this feature a bit. Also makes a bit more sense to have shulker boxes to stack to 16 instead of 64 since Minecraft has already established (see: bucket) that empty stuff can only stack to 16. Anyway this just allows it to be configurable.

The old stackableShulkerBoxes rule has been removed and replaced with simply using the value "1" as the stack size for shulker boxes. It's up to you if this is how you want to implement it, I think it makes the most sense to have it as the one option but people might prefer having two separate options, one for stack size and one for whether they stack at all. 

With my implementation, the configuration will simply default to 1, which is basically "off" like before and no calculations will be attempted until the value exceeds 1.

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -205,11 +205,22 @@ public class CarpetSettings
     @Rule( desc = "Players absorb XP instantly, without delay", category = CREATIVE )
     public static boolean xpNoCooldown = false;
 
+    public static class ShulkerBoxStackSizeValidator extends Validator<Integer>
+    {
+        @Override
+        public Integer validate(ServerCommandSource source, ParsedRule<Integer> currentRule, Integer newValue, String string) {
+            return (newValue>=1&&newValue<=64) ? newValue : null;
+        }
+        @Override
+        public String description() {return "Must choose value between 1 and 64";}
+     }
+
     @Rule(
             desc = "Controls the amount of empty shulker boxes that can be stacked together when dropped on the ground",
             extra = ".. or when manipulated inside the inventories",
             options = {"1", "8", "16", "64"},
             strict = false,
+            validate = ShulkerBoxStackSizeValidator.class,
             category = {SURVIVAL, FEATURE}
     )
     public static int shulkerBoxStackSize = 1;
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -236,7 +236,8 @@ public class CarpetSettings
     }
 
     @Rule(
-            desc = "Empty shulker boxes can stack to 64 or a custom value.",
+            desc = "Empty shulker boxes can stack to 64 or a custom value when thrown on the ground.",
+            extra = ".. or when manipulated inside the inventories",
             validate = StackableShulkerBoxValidator.class,
             options = {"false", "true", "16"},
             strict = false,
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -205,24 +205,44 @@ public class CarpetSettings
     @Rule( desc = "Players absorb XP instantly, without delay", category = CREATIVE )
     public static boolean xpNoCooldown = false;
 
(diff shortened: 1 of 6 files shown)


Fallen-Breath:
Changing the rule name without any migration logic may cause contraptions in the world to be broken if users don't enable the new rule early enough, that's very dangerous imo

James103:
In other words, some conversion code will need to be added in order to detect if `/carpet stackableSkulkerBoxes` is set to true in the world config file when the world is being loaded, and if so, set the value of `/carpet shulkerBoxStackSize` to 64.

TomB-134:
I agree with the both of you. As i see it there are 2 options, the lazy way which is just to leave the old option in or write conversion code (which is trival to do). I think the second approach is best. What do you guys think?

CommandLeo:
Second approach is better in my opinion

TomB-134:
Fairly simple implementation, old shulker rule gets recognized and and replaced with new rule with value of 64. Old rule will get replaced if any change is made to the defaults.

gnembon:
but why? Stacking of xboxes makes a big difference in the gameplay and that's what this rule is for. Having it configurable makes is more confusing IMO. You could argue that any other powerful item should have a stacking rule. Another argument agains is that sbox based contraption that do count items, stacks etc now become dependent on this value, so when you take someone elses stacker / unstacker it may not work, since on your server its limited to 16 for example.

I am not yet convinced that extra complication is needed.

TomB-134:
I understand your point about contraptions breaking due to a different value being set. Maybe this rule would work better as a carpet extra feature since a lot of rules in that will also break contraptions depending on the value, so I don't think we're changing the status quo with making it configurable. Turning on moveable tile entities for example would break quite a lot of existing contraptions. Thoughts?

It's also worth noting there are contraptions that break dependent on the server simulation/render distance, should that not be configurable because of the risk of breaking contraptions? I don't think so. I acknowledge that it's a rare situation but so is needing a shulkerbox stacker/unstacker. (Yes I understand those rules exist primarily for performance reasons but in an ideal world, shulkerbox stack size wouldn't need to be configurable and would have a hard limit in the game, but since we're using a mod we might as well allow the user to control exactly what that value is, because isn't carpet meant to allow you to take full control of the vanilla game?)

Finally, I believe that setting the shulker box stack size to 64 was a mistake right from the introduction of this rule. I understand it was done that way because people were used to the crafting glitch in 1.12, but ultimately I think it flew in the face of minecraft's already established rules about buckets stacking. Having the value be 16 makes much more sense and introduces (in my opinion) some much needed balance to this.

gnembon:
The migration option in SettingsManager is absolutely awful, I hope you can see this. 

While I don't agree about this change, what you could do is change the rule type to String allowing boolean values primarily (false = 1, true = 64), but also accepting other strings like 16. That's the only option I can see to adopt this.

TomB-134:
> While I don't agree about this change, what you could do is change the rule type to String allowing boolean values primarily (false = 1, true = 64), but also accepting other strings like 16. That's the only option I can see to adopt this.

How does what you suggest improve anything? As in what's the advantage of having it a string over an int like it is now?

TomB-134:
I understand what you mean now, will implement it once I have time.

src/main/java/carpet/settings/SettingsManager.java
```diff
@@ -485,6 +485,7 @@ private ConfigReadResult readSettingsFromConf(Path path)
         }
     }
 
+
```
gnembon, line 488:
```suggestion
```

src/main/java/carpet/CarpetSettings.java
```diff
@@ -205,13 +205,46 @@ public String description()
     @Rule( desc = "Players absorb XP instantly, without delay", category = CREATIVE )
     public static boolean xpNoCooldown = false;
 
+    public static class StackableShulkerBoxValidator extends Validator<String> //Doing the logic here avoids doing the logic several times over in the relevant mixin classes.
```
gnembon, line 208:
```suggestion
    public static class StackableShulkerBoxValidator extends Validator<String> 
```

gnembon:
yay

PR #1258, opened 2022-01-02
adding structure block support

ch-yx:


James103:
Will you please add some documentation for this PR as well?

ch-yx:
@James103 OK.

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -673,7 +683,104 @@ else if (v instanceof EntityValue)
             });
             return (cc, tt) -> result[0];
         });
+        // =======================================
+        expression.addContextFunction("structure_block_save", -1, (c, t, lv) -> {
+            CarpetContext cc = (CarpetContext) c;
+            ServerWorld lv2 = cc.s.getWorld();
```
Ghoulboy78, line 689: why do you name your variables like lv2, lv3, so on? It kinda makes it hard to know what is what when reading the code

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -673,7 +683,104 @@ else if (v instanceof EntityValue)
...
+                return Value.NULL;
+            }
+            lv6 = lv3.getStructureOrBlank(struident);
+
+            // name,start,dimensions, ignoreEntities, ignoredBlock,~~author~~,disk
+            lv6.saveFromWorld(lv2,
+                    new BlockPos(((NumericValue) lv.get(1)).getInt(), ((NumericValue) lv.get(2)).getInt(),
+                            ((NumericValue) lv.get(3)).getInt()),
+                    new Vec3i((int) lv.get(4).readInteger(), (int) lv.get(5).readInteger(),
+                            (int) lv.get(6).readInteger()),
```
Ghoulboy78, line 707: There is a specific way to get triplets of positions from command arguments, so that users can either input 3 numbers or a list of 3 numbers. See code for block() function and look for BlockLocator class

docs/scarpet/api/BlocksAndWorldAccess.md
```diff
@@ -857,3 +857,28 @@ with the game and assumed not changing.
...
+
+Save part of the world into a structure.
+`ignoredBlock` may be null or a BlockValue (eg: `block('air')`).if not null, that kind of block will not be saved.
+If `disk` is true, the structure will saved in disk. Or it will only exist temporarily.
+Return True if saved successfully. Return a falsy value if not.
+
+### `structure_block_load(name,start_x,start_y,start_z,ignoreEntities,integrity,awake,noupdate,fluid,gravity,rotation,mirror)`
+
+Place a saved structure into the world.
+If `integrity` is not null, it will be the probability of each block's not disappearing.
```
Ghoulboy78, line 873: Is this `integrity` argument between 0 and 1, cos if so you should specify it

docs/scarpet/api/BlocksAndWorldAccess.md
```diff
@@ -857,3 +857,28 @@ with the game and assumed not changing.
...
+If `noupdate` is true, it will not update blocks near them.
+If `gravity` is not null, the structure will fall (or fly) to the ground. And the value of `gravity` will be the Yoffset. (Newton will be happy about it.)
+Structure will rotate 90°*`rotation` before placed.
```
Ghoulboy78, line 876: So if I put 1 then the structure would be rotated 90° (again, could you put an example to explain here pls?)

ch-yx:
there are too many inputs of the functions. 
difficult to remember and use.
should i change it into a map?

docs/scarpet/api/BlocksAndWorldAccess.md
ch-yx: i added that in doc

ch-yx:
now, start of structure_block_save/load could be a block value.
but if it is a block value, its block type and the dimension it is in will be ignored.

src/main/java/carpet/script/api/WorldAccess.java
ch-yx: done.

gnembon:
Good job - this is definitely needed.

Few comments - the fact that structure manager can store structures temporarily is implementation detail, or rather implementation annoyance. Also the use of 'unload' is unclear - why you would ever need that. Making structure api non-transactional / atomic would make is much more useful especially in parallel usecase.

Suggest `save_structure(id, corner_pos, other_pos, save_entities=false?, ignored_block?)` and `place_structure(id, corner_pos, ...)`

the use of 'save' could be debatable - I could still leave it as option and return actual nbt value as a return (looking it the calling context if VOID then not doing that, but I bet you could dig out the NBT of it somewhere when saving, so it doesn't have to be done twice). if ID is null maybe the strucuture is not saved, just returned.
```
structure_data = save_structure(null, from, to);
....
place_structure(structure_data, other_place);
```

this way you could use save and load to store structures in app data, not in world structures place by storing and loading NBT value.

this means place_structure could accept NBT value instead of id, allowing copy blocks using structure API internally without touching the disk. - that would be useful. As you said, place - could use an options map instead of a long parameter list.

Again, still no clue why you would need 'unload'.

ch-yx:
> Again, still no clue why you would need 'unload'.

I added it because it can cast a forced reload if I changed structure\`s nbt file by some editors or some other functions about datapack. 

And I do agree that it should be able to return/accept a NBT object. @gnembon

ch-yx:
图片
图片
@gnembon now `﻿save﻿` supports returning a nbt obj and `﻿load﻿` supports receiving it.
if `save` s output file name is `null` it will return the nbtfile instead

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -673,7 +683,122 @@ else if (v instanceof EntityValue)
...
+
+            return Value.TRUE;
+        });
+
+        // =====================================
+        // =======================================
+        expression.addContextFunction("structure_block_load", -1, (c, t, lv) -> {
+            CarpetContext cc = (CarpetContext) c;
+            ServerWorld world = cc.s.getWorld();
+            boolean recievnbt=lv.get(0).getTypeString().equals("nbt");
```
gnembon, line 735: instanceof ?

gnembon:
👍 

Can you please explain what the issue is with unload - didn't follow the example. i.e. in which case that could be messed up? Still can't understand why/when you would need to use it since structure blocks pretty much have save and load and that's it.

ch-yx:
> Suggest `save_structure`

I don't think it is good to rename them save_structure/place_structure.

it is confusing. I saw someone in the discord server plan to add some functions about structures(wg,not structure block).

ch-yx:
> 👍
> 
> Can you please explain what the issue is with unload - didn't follow the example. i.e. in which case that could be messed up? Still can't understand why/when you would need to use it since structure blocks pretty much have save and load and that's it.

because in vanilla, unload exists.

corner structure blocks unload structure when powered.

and it is useful, I said before. 

>   because it can cast a forced reload if I changed structure`s nbt file by some editors or some other functions about datapack.

that is why i m gonna leave it there. if you dont need it, you can ignore it. but if you need it, it have to be there.

src/main/java/carpet/script/api/WorldAccess.java
ch-yx: yes. i should use instanceof here.

and i should have some tests: if it crashes when received a wrong nbt file.

gnembon:
well, there is a usecase for unload maybe, but here the documentation provides zero usecase for it. Like why I would bother if save and load does all the job I need about them? Note that internal API is internal for a reason - scarpet is an example of an external API and if there is no use for some function it shoudn't be exposed. 

Like the corner structure block - it maybe useful to define a corner of a structure becuase in the game you do everything step by step, but here you need to provide both corners at once anyways, right? Am I missing something?

gnembon:
regarding the naming things - structure blocks are blocks because they are actual blocks in the game. The thing they are loading and saving are just structures. Another naming convention you could use here to distinguish from WG structures would be to use the mojmap name for it so load_template save_template or load_structure_template and save_structure_template

gnembon:
yeah, looking at now how management of structures looks like - there should not be any unloading - save and load should not leave any residues in the StructureManager - just don't need to leave any state with them - the contract seems to be nice and clear here

ch-yx:
as the using mapping is mojmap now,

please allow me to start a new pr.
and i m closing this one.

PR #1264, opened 2022-01-05
Fake players can be spawned in survival mode while flying

### LINKED ISSUE #1263: [Bug][Bots] Players spawned in survival while spawning player is in spectator, spawn flying in survival
Basically the title.

## Steps to reproduce
- Run `/player Steve spawn at x y z facing u v in minecraft:overworld in survival`, y being any y level mid air, while in spectator, 

## Expectation
- Steve should fall on ground since its in survival mode

## Observation
- Steve will spawn flying midair in survival mode. If you do `/player Steve move forward` they walk forward mid air

## Version info
- Game: 1.17.1
- Carpet 1.4.56

Ghoulboy78:
Fixes #1263 
Basically if you're in spectator (or creative for that matter), spawning a fake player in survival or adventure mode would cause that player to have flying abilities.

### DIFF
--- a/src/main/java/carpet/commands/PlayerCommand.java
+++ b/src/main/java/carpet/commands/PlayerCommand.java
@@ -299,6 +299,9 @@ public class PlayerCommand
         {
             // Force override flying to true for spectator players, or they will fell out of the world.
             flying = true;
+        } else if(mode.isSurvivalLike()){
+            // Force override flying to true for survival-like players, or they will fly too
+            flying = false;
         }
         String playerName = StringArgumentType.getString(context, "player");
         if (playerName.length()>maxPlayerLength(source.getServer()))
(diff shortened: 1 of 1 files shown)


src/main/java/carpet/commands/PlayerCommand.java
```diff
@@ -299,6 +299,9 @@ private static int spawn(CommandContext<ServerCommandSource> context) throws Com
         {
             // Force override flying to true for spectator players, or they will fell out of the world.
             flying = true;
+        } else if(mode.isSurvivalLike()){
+            // Force override flying to true for survival-like players, or they will fly too
```
James103, line 303:
```suggestion
            // Force override flying to false for survival-like players, or they will fly too
```

gnembon (approved):

PR #1269, opened 2022-01-08
updated documentation for 1.18

random-j-farmer:
I encountered some out-of-date documentation while working on https://github.com/gnembon/scarpet/issues/280


gnembon:
no need to change Full.md - its automatically compiled from individual md files.

I suggest more general messaging there, like `be aware that depending on the MC version and dimension settings biome can be set either in a 1x1x256 column or 4x4x4 hyperblock, so for some versions Y will be ignored and for some precision of biome setting is less than 1x1x1 block)`. - feel free to word it properly.

random-j-farmer:
I changed it as requested.

So setting at y 0 was never required, it's just that it effected the whole column?

Otherwise I would leave something about y 0 in.

PR #1279, opened 2022-01-19
minor issue

ch-yx:


src/main/java/carpet/script/value/NumericValue.java
```diff
@@ -56,7 +56,7 @@ public String getString()
         }
         try
         {
-            if (Double.isInfinite(value)) return "INFINITY";
+            if (Double.isInfinite(value)) return (value>0)?"+INFINITY":"-INFINITY";
```
Crec0, line 59: `+INFINITY` is not a thing. Its `INFINITY` or `-INFINITY`

Can also do `(value > 0 ? "" : "-") + "INFINITY"`

gnembon, line 59:
```suggestion
            if (Double.isInfinite(value)) return (value>0)?"INFINITY":"-INFINITY";
```

PR #1283, opened 2022-01-22
Update Overlay app : modded structures parsing

aria1th:
Modded structures uses modname : structurename, so it can't be directly used to command. Instead, use stripped structure names to process it.


altrisi:
I wonder if an argument type for structures could be made so it gets them directly... Specially with the new vanilla plop, that must exist somewhere.

altrisi:
Do you have something to test this with compatible with 1.19 pres?

ch-yx:
please do not hate me, but it seems that your code made it more bad.....
im in 1.19.1-pre4
图片
in fact, i only replace one word in the original file and it can correctly draw bbox of my scructure in my datapack:
图片
change the word from 'term' to 'string'  ⬆️ :


i replaced the word to string in your version as well,......and it still cannot work......
here is the datapack i used : 
[jvl.zip](https://github.com/gnembon/fabric-carpet/files/9127398/jvl.zip)

ch-yx:
图片
seems that all you need to do to make it works is change the argument type to `identifier` . then it support `:` naturely.
no " needed

altrisi:
Confirmed just changing the type to `identifier` works, will do that instead since it handles conflicts as well.

PR #1284, opened 2022-01-23
Fixed firing of on_move event event after no change in position.

### LINKED ISSUE #1282: on_move unexpected behavior
The `on_move` entity event is documented as:
`executes every time an entity changes position, invoked just after it has been moved to the new position.`

Yet in my testing on players, it executes regardless of a change in position up to 4 times per tick, with pos1 and pos2 equating to be the same a majority of the time.
The 4 times per tick issue can be reproduced by falling or flying. Though I can see this being intentional if position truly changes more than once per tick. Which is not the case in my testing so far outside of an app/command setting positions itself.
Standing still sees the event execute once per tick, with pos1 equating to pos2.

Either this is incorrect behavior, or the documentation ought to be adjusted to:
`executes every time an entity position is set regardless of it changing, invoked just after it has been set to the new position.`

Ghoulboy78:
Fixes #1282 
It's basically a check for whether or not the position changed. Idk if there's a neater way to put it tho tbh.

### DIFF
--- a/src/main/java/carpet/mixins/Entity_scarpetEventsMixin.java
+++ b/src/main/java/carpet/mixins/Entity_scarpetEventsMixin.java
@@ -94,6 +94,7 @@ public abstract class Entity_scarpetEventsMixin implements EntityInterface
     @Inject(method = "setPos", at = @At("TAIL"))
     private void secondPos(CallbackInfo ci)
     {
-        events.onEvent(EntityEventsGroup.Event.ON_MOVE, motion, pos1, this.pos);
+        if(pos1!=this.pos)
+            events.onEvent(EntityEventsGroup.Event.ON_MOVE, motion, pos1, this.pos);
     }
 }


CommandLeo:
Will this be backported to 1.17?

James103:
> Will this be backported to 1.17?

1.17.1 is no longer supported by Carpet mod.

ch-yx:
im new to java. 
could any one tell me why dont you use `pos1.equals(this.pos)` here?

CommandLeo:
> > Will this be backported to 1.17?
> 
> 1.17.1 is no longer supported by Carpet mod.

@James103 
It was said some time ago in Carpet Discord that important bugfixes would be ported to 1.17 even after moving to 1.18.

altrisi:
You can always apply the change in your own build of the mod for 1.17 if you need it, or make a PR to the 1.17 branch and ask for another release with it, and see if gnembon accepts it.

PR #1290, opened 2022-01-25
adding a new function

ch-yx:


src/main/java/carpet/script/language/Loops.java
```diff
@@ -140,7 +143,45 @@ public static void apply(Expression expression)
...
+                    
+
+                    @Override
+                    public boolean hasNext() {
+                        return a.stream().allMatch(Iterator::hasNext);
+                    }
+
+                    @Override
+                    public Value next() {
+                        return ListValue.wrap(a.stream().map(Iterator::next));
```
ch-yx, line 162: worrying if some of the element of this list is the same iterator.
Then these iterators run the risk of overstepping the bounds.
But it is a classic use of zip to put the same object in the argument multiple times.
If it does go out of bounds I can't squeeze the elements I've taken out back into the iterator 🤔

gnembon: valid concern, because peeps might want to use it like a = range(10); zip(a, a); btw - that works in python

src/main/java/carpet/script/language/Loops.java
```diff
@@ -140,7 +143,45 @@ public static void apply(Expression expression)
             c.setVariable("_i", _iter);
             return (cc, tt) ->  ret;
         });
+//=========
+        expression.addFunction("zip", (lv) ->
+        {
+            if (lv.stream().allMatch(v->(v instanceof AbstractListValue))){
+                return new LazyListValue(){
+                    List<AbstractListValue> b=lv.stream().map(v->(AbstractListValue)v).collect(Collectors.toList());
```
gnembon, line 151: I would name these better

src/main/java/carpet/script/language/Loops.java
```diff
@@ -140,7 +143,45 @@ public static void apply(Expression expression)
...
+                    List<Iterator<Value>> a=lv.stream().map(v->((AbstractListValue)v).iterator()).collect(Collectors.toList());
+                    
+
+                    @Override
+                    public boolean hasNext() {
+                        return a.stream().allMatch(Iterator::hasNext);
```
gnembon, line 157: I would not use streams. zip seems could be used in mission critical situations and could be operating on basic types, like numbers

gnembon:
Great idea. needs documentation, API entries, and primarily rethinking of the consecutive use of iterators

ch-yx:
here is a example of the new added generator creater function.
the generator has a Value to store its Status which is Null at the beginning. 
the first argument is a expr that returns [value to be yielded, new_status]
the second arguments is a expr that returns if the generator hasnext.

/script run g=generator(if(_==null,status={'now'->0,'end'->5},status=_);status:'now'+=1;[status:'now'-1,status],_:'now'<0+_:'end');[g]

src/main/java/carpet/script/language/Loops.java
ch-yx: i changed it to zip_longest and fill the void with Null...😅

PR #1291, opened 2022-01-25
Allow serializing FormattedText values to JSON in scarpet

replaceitem:
The `toJson` method is not overridden in the `FormattedTextValue`.
Enabling that can be useful for getting the json strings from a text value.

*I thought about adding the same but for deserializing a json string to a text. However, i could not think of a place this would fit in scarpet. Adding a new function just for this seems overkill, but there is nothing that could be used instead (as far as i can tell).*


gnembon:
conflicts - should be trivial to rename

gnembon (approved):

Issue #613, opened 2020-12-06
is __config -> stay_loaded really needed? Should autoload be even there and off by default?


gnembon:
With the addition of storing scripts in .minecraft/config/carpet/scripts folder, does it still makes sense to have that option in config, and not assuming if a script is in the world scripts folder it needs to be loaded.

scripts in the appstore all tend to have that set, possibly to avoid - I put it in the world files, but script doesn't load. - yeah, you not only need to enable scripts loading in the carpet settings, but also edit that file I told you to download to load automatically.

On the same note - scripts autoloading should be on by default? if you don't have any scripts it won't have any effect so its still voluntary and have no effect by default, and scripts from shared .minecraft/config won't be loading automatically.

The main reason for that, tbh, is that I typically had several scripts I was testing at all times and having all of them load automatically would be super annoying, but now with a better location for shared and non-automatic apps location - it makes sense that whatever you put in the world/scripts floder, just works if you put it there?

Firigion:
It makes sense for survival, but no so much for tools, I'm not sure I'm convinced. Maybe set the carpetrule autoload to true by default, but sill have a way for scripts in the scripts folder of the world not automaticaly load. Maybe make the config work by the negative? `__config() -> {'Autoload'-> false}` is required for it not to be automaticaly loaded?

BisUmTo:
Honestly I always "hated" the fact of having to specify a script to load itself automatically and load it for the first time.
80% of the scripts always need to be loaded, as they are used as "mods / datapacks" and if you had lots of them and you wanted to start a new world with the same scripts you always need to load them for the first time...

In my opinion commands like `/script install <name> [globally true|false]` and `/script uninstall <name> [globally true|false]` might be really useful!
This ipotetic command can load/unload permanently (install/uninstall) a script that has `__config->{'installable'->true}` and it will stay loaded.

The scripts in the world folder could automatically install for the current world (if they are `installable`); stored scripts could need to be installed manually in the world instead.

The `globally` boolean could be useful to install a stored script in all your worlds.
[If in a LAN server, only the owner can manipulate this bool]

With those new commands the `/carpet ScriptAutoload` rule could be removed, on my opinion...

altrisi:
[...]

PR #1295, opened 2022-01-26
Fix `stay_loaded` being `false` in apps with no config

altrisi:
Does that.

Shouldn't conflict with the mojmap refactor.


gnembon:
yeah but that's intended right?

gnembon:
its been changed a long time ago, that apps will only stay loaded if they request to do so not to pollute the app space from stuff that you left in the apps folder.

altrisi:
No, it was made true by default, but that `true` default value is only applied if a `__config` function is found.

Issue was #613, commit e8a6b89.

gnembon (approved):

Issue #1272, opened 2022-01-14
[Internals Suggestion] Using MixinExtras


altrisi:
[MixinExtras](https://github.com/LlamaLad7/MixinExtras) is an "addon" to Mixin that provides two new features: `ModifyExpressionValue` and `WrapWithCondition`.

The reason I believe it'd be a good idea to use them is because unlike using `Redirect`, those can be chained with the ones from other mods in a compatible way, allowing multiple mods to add to a condition. The author claims they are even compatible with existing redirects from other mods to the same target.

Not only that, they're obviously more expressive in the code given they have the specific purpose, and it's not a general redirect.

So let's get to what they do (probably better explained [on their wiki](https://github.com/LlamaLad7/MixinExtras/wiki)):
## [`ModifyExpressionValue`](https://github.com/LlamaLad7/MixinExtras/wiki/ModifyExpressionValue)

This basically allows to add to a condition without having to make a `Redirect` to the complete method call. So for example, all those redirect mixins in the creative no clip mixin (https://github.com/gnembon/fabric-carpet/blob/af368a159d2c3c287acbd22d9a9c0f0094519c56/src/main/java/carpet/mixins/PlayerEntity_creativeNoClipMixin.java), they'd be `ModifyExpressionValue` that would take the original result of the expression, and Carpet could then add an `original || (CarpetSettings.creativeNoClip && playerEntity.isCreative() && playerEntity.getAbilities().flying)`. While still being compatible with any mod redirecting the same expression.

Something like this (not tested directly):

https://github.com/gnembon/fabric-carpet/blob/af368a159d2c3c287acbd22d9a9c0f0094519c56/src/main/java/carpet/mixins/PlayerEntity_creativeNoClipMixin.java#L20-L28

to

```java
@ModifyExpressionValue(
    method = "tick", 
    at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/player/PlayerEntity;isSpectator()Z")
)
private boolean canClipTroughWorld(boolean original)
{
    return original || (CarpetSettings.creativeNoClip && playerEntity.isCreative() && playerEntity.getAbilities().flying);
}
```

This is done _a lot_ in the carpet codebase.

## [`WrapWithCondition`](https://github.com/LlamaLad7/MixinExtras/wiki/WrapWithCondition)

This one wouldn't be used as much, but it's still nice that it makes those compatible with other mods doing similar things.

[...]

PR #1310, opened 2022-02-03
Improve compatibility with other mods by using MixinExtras and reimplementing the shapes compatibility layer

### LINKED ISSUE #1309: Scarpet shapes are back to being incompatible with Canvas
The compatibility layer was commented out.

altrisi:
This PR intends to improve compatibility with other mods _without_ loosing functionality (no "`require = 0` compatibility", this actually removes two of those). This should also fix the worldedit conflict without needing them to disable features with Carpet.

- Adds (and uses it for most things it can be) Mixin Extras (implements #1272)

  Things I didn't change: the ones around movable BEs (just in case), the ones around `CarpetSettings.skipGenerationChecks` (there was a comment somewhere that made me not want to change those), the ones that would slightly decrease performance (such as dispenser qc, given the original expression is evaluated regardless of order)

- Readds the shapes compatibility layer with different rendering engines (Canvas) (fixes #1309)

  It was removed in a (snapshot?) commit, but the issue still occurs without it.

Pending stuff (will do likely tomorrow, feel free to test though):
- ~~Assert the Mixin Extras dependency is bundled for production~~ checked
- Assert it works in production
- Maybe further testing in production? Just in case, idk
- Assert compatibility with removed `require = 0`s is still a thing
  I'll likely need to backport those to test, since these mods aren't in snapshots


gnembon:
question - why replace all these redirects, only do that for mixins that are questionable?

Scenario - what if suddenly something stop being supported in this external library?

altrisi:
The scenario is already solved by "we bundle the library", so we are in control of the version and loaded classes, and it only depends on Mixin's API, which we already depend on and is unlikely to change (and even if it did, given fabric loader is compatible with all mc versions their fork also ensures compatibility with API changes in Mixin, like it happened during the loader 0.12 update, that the Mixin API behaviour that broke was presented changed only via opt-in).

As to why replace them all? I don't really know, but my intention was to prevent future issues and provide the compatible way directly (you can't really know what other modders will want to do in the future, and many users will just think "it crashes" and remove one of the mods), and in some cases (IMO) having the annotation say exactly what it does makes it easier to read (`WrapWithCondition` vs `Redirect`). It also makes sure that if there's a subtle change in vanilla, those would be caught here too (example a method that called `something(variable)` starts calling `something(variable+2)` wouldn't be updated in Carpet's mixin if using a Redirect, but would with this. Not really sure if that can actually happen, but still).

And because I had some free time with nothing else to do really.

gnembon:
My gut feeling tells me its better to stick to standards and implement 'new shiny things' only where it really matters. 

And with things getting incompatible - its fine. its better to get a bug report and ponder what is the best way to solve it, rather than have a solution that would silently don't work as intended.

gnembon:
Too is great, but its non standard. 

Should only be used in places known for causing compat issues, and only there.

altrisi:
Will close it given it's got full of conflicts, will open a new one if I redo it at some point.

PR #1312, opened 2022-02-04
Added __on_player_message event

### LINKED ISSUE #937: [Scarpet][Feature Request] Intercepting and reacting to player chat
Hey :)

I was browsing through the Scarpet API docs and noticed, that there is no event for when a player sends a chat message.  
Would it be possible to add that or is the only way for player chat to be processed by Scarpet via the command system?  
Regarding that, can one define more than one top-level command per app? For instance `/myapp` and `/myapp_admin`?

replaceitem:
Adds an `__on_player_message(player,message)` event, which triggers when a player sends a chat message or runs a command.

This is part of #937, and the interception part will be achievable when event cancellation is a thing.

### DIFF
--- a/docs/scarpet/api/Events.md
+++ b/docs/scarpet/api/Events.md
@@ -275,6 +275,9 @@ Triggered when the player has successfully logged in and was placed in the game.
 ### `__on_player_disconnects(player, reason)`
 Triggered when a player sends a disconnect package or is forcefully disconnected from the server.
 
+### `__on_player_sends_message(player, message)`
+Triggered when a player sends a chat message or runs a command.
+
 ### `__on_statistic(player, category, event, value)`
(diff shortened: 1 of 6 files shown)


gnembon:
would `__on_player_message` be sufficient?

gnembon:
oh, and please add to editor support files

docs/scarpet/api/Events.md
```diff
@@ -275,6 +275,9 @@ Triggered when the player has successfully logged in and was placed in the game.
 ### `__on_player_disconnects(player, reason)`
 Triggered when a player sends a disconnect package or is forcefully disconnected from the server.
 
+### `__on_player_sends_message(player, message)`
```
gnembon, line 278:
```suggestion
### `__on_player_message(player, message)`
```

gnembon (approved):

Issue #1074, opened 2021-08-22
Scarpet suggestion: Ability to use localized text in `format`ted strings


James103:
Currently, in order to use localized text anywhere from within Scarpet, you would need to use a command to set the text. For example, if you wanted to print "Shears" in English and "Tijeras" in Spanish, you would need to run the following command, even if it is done from within Scarpet:
`tellraw @s {"translate":"item.minecraft.shears"}`

In order to avoid having to run commands every tick while still allowing for per-player localized text, a new component type would need to be introduced into the `format` function:
* `j`: Translate (the following word including periods is treated as a string ID for which to turn into localized text).

This would now allow you to run something like `script run print('j item.minecraft.shears')` to do the same as above.

BisUmTo:
What about translation with arguments?
`script run print(format('j multiplayer.player.joined Herobrine'))` would be a nice solution, but what happens if the arg has a space?

James103:
If player want to use the localization system to its full potential, a new function may be required and it would be called `translate`:
### `translate(format, id, ...args)`
Returns a localized `FormattedText` object. `format` is the format specifiers to apply on the full string. `id` is the translation ID of the string to use. `...args` can be either strings or `FormattedText` objects, which get inserted as translation arguments.

For example, `translate('b', 'death.attack.mob', format('i A'), 'B')` would return a Formatted Text corresponding to
(in English) _A_ **was slain by** B
(in Spanish) _A_ **fue asesinado/a por** B

Another example: `format(translate('b', 'death.attack.mob', format('i A'), 'B'), 'g \ (x3)'` would output:
(in English) _A_ **was slain by** B (x3)
(in Spanish) _A_ **fue asesinado/a por** B (x3)
with the "(x3)" in gray.

Note that going the functional route will eliminate the need for using `j` as a format specifier for localized text.

PR #1315, opened 2022-02-06
(solve #1074)make it able to create TranslatableComponent or KeybindComponent

ch-yx:
* `UPDATE`
seems that gnembon really want a function called translate
so i add one. 
https://github.com/gnembon/fabric-carpet/pull/1315/commits/bef2339686648107362af864051dc2680d7911f1

------------------------

I&#39;ve completely rewritten it.
I wrote a method called Messenger2.c() . Hopefully it will replace the existing Messenger.c() . Because the existing one has some flaws. (such as the parameter...
it is the same as newformat, but adding a 'j' at the front of the first argument.

I've completely rewritten it. 

I wrote a method called Messenger2.c() . Hopefully it will replace the existing Messenger.c() . Because the existing one has some flaws. (such as the parameter can only be string value, and cannot be text value & hover text cannot be complex.)I also managed to make the new function fully compatible with the way to use the old one.

and add a scarpet function which is nothing but a warper of it for testing.i named it "newformat" temporary as i do not want to remove the old one too early to compare.

example:
`/script run print(newformat('bj death.fell.assist.item','e good guy',p,p~'holds','|','^','_ key.command',newformat('e g','b g')))`

'j' (maybe '%' is better) from the first arg marks that it is a translation component. args between it and next '|' ( or #EOF) will be its inputs.
'^' marks that next args will be hover text. Current '^b eee' format is allowed too.(maintain compatibility) '^b eee' == '^','b eee'
'_' marks that it is a keyboard component. (key.command is '/' for me)
图片

This is a structure that allows recursion. It can be nested.

inner newformat function is not necessary. but u can put it here too.

and entities/item\`s len3 vectors are allowed too.

### DIFF
--- a/src/main/java/carpet/utils/Messenger2.java
+++ b/src/main/java/carpet/utils/Messenger2.java
@@ -100,10 +100,7 @@ public class Messenger2 {
         c = 0;
     }
 
-    /*
-     * composes single line, multicomponent message, and returns as one chat
-     * messagge
-     */
+
     public Component getSomeComponents(){
         return getSomeComponents("|");
     }
@@ -136,7 +133,7 @@ public class Messenger2 {
                 case '!' -> previousStyle
                         .withClickEvent(new ClickEvent(ClickEvent.Action.RUN_COMMAND, message.substring(1)));
                 case '^' -> previousStyle
-                        .withHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, getSomeComponents()));
+                        .withHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, message.equals("^")?getARawComponent():c(message.substring(1))));
                 case '@' -> previousStyle
                         .withClickEvent(new ClickEvent(ClickEvent.Action.OPEN_URL, message.substring(1)));
                 case '&' -> previousStyle
@@ -242,6 +239,6 @@ public class Messenger2 {
         }
     }
 
-    // simple text
+
 
 }
--- a/src/main/java/carpet/utils/Messenger2.java
+++ b/src/main/java/carpet/utils/Messenger2.java
@@ -1,0 +1,238 @@
+package carpet.utils;
+import java.util.regex.Pattern;
+import net.minecraft.ChatFormatting;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.BaseComponent;
+import net.minecraft.network.chat.ClickEvent;
+import net.minecraft.network.chat.HoverEvent;
+import net.minecraft.network.chat.KeybindComponent;
+import net.minecraft.network.chat.MutableComponent;
+import net.minecraft.network.chat.Style;
+import net.minecraft.network.chat.TextColor;
+import net.minecraft.network.chat.TextComponent;
+import net.minecraft.network.chat.TranslatableComponent;
+
+import net.minecraft.world.entity.Entity;
+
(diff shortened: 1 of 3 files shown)


ch-yx:
i still want that it is using `format` too………
may I use `. `?

ch-yx:
image_179
image_180
allowing translation text support styles.

ch-yx:
thank @commandleo for his pointing out that this fixed #1074

ch-yx:
as fallenbreath is focusing on translating thing, i want to re write it.

ch-yx:
example:
`/script run print(newformat('bj death.fell.assist.item','e good guy',p,p~'holds','|','^','_ key.command',newformat('e g','b g')))`

'j' from the first arg marks that it is a translation component. args between it and next '|' ( or #EOF) will be its inputs.
'^' marks that args between it and next '|' (or #EOF) will be hover text.
'_' marks that it is a keyboard component.

This is a structure that allows recursion. It can be nested

src/main/java/carpet/utils/Messenger2.java
```diff
@@ -0,0 +1,247 @@
...
+        c += 1;
+        return (base, message) -> {
+            Style previousStyle = base.getStyle();
+            base.setStyle(switch (ch) {
+                case '?' -> previousStyle
+                        .withClickEvent(new ClickEvent(ClickEvent.Action.SUGGEST_COMMAND, message.substring(1)));
+                case '!' -> previousStyle
+                        .withClickEvent(new ClickEvent(ClickEvent.Action.RUN_COMMAND, message.substring(1)));
+                case '^' -> previousStyle
+                        .withHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, getSomeComponents()));
```
ch-yx, line 139: maybe i should use getAComponent() to replace getSomeComponents()
dont know which is better.

src/main/java/carpet/utils/Messenger2.java
```diff
@@ -0,0 +1,245 @@
...
+            res= (ent.getDisplayName());
+            c += 1;return res;
+        }
+        if (now() instanceof String message) {
+            if (message.equalsIgnoreCase("")) {
+                c+=1;
+                return new TextComponent("");
+            }
+            if (message.equals("[")){
+                c+=1;
```
ch-yx, line 179: to be honest, that is not necessary and kind of ugly, as u can simply put a Component as a arguments. 

but maybe some will enjoy in it as it is sometimes useful if u want to add click events to some components together.

src/main/java/carpet/utils/Messenger2.java
```diff
@@ -0,0 +1,245 @@
...
+        c += 1;
+        return (base, message) -> {
+            Style previousStyle = base.getStyle();
+            base.setStyle(switch (ch) {
+                case '?' -> previousStyle
+                        .withClickEvent(new ClickEvent(ClickEvent.Action.SUGGEST_COMMAND, message.substring(1)));
+                case '!' -> previousStyle
+                        .withClickEvent(new ClickEvent(ClickEvent.Action.RUN_COMMAND, message.substring(1)));
+                case '^' -> previousStyle
+                        .withHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, message.equals("^")?getARawComponent():c(message.substring(1))));
```
ch-yx, line 137: because hover text is not click able, i use getARawComponent here rather than getAComponent.

PR #1318, opened 2022-02-07
Missing EntityValue documentation

carseatheadrest:
Adds a few values from `src/main/java/carpet/script/value/EntityValue.java` that were missing in the documentation. Also, fixes some typos.


gnembon:
Thanks. Full.md is autogenerated from other .mds so no need to edit that

PR #1321, opened 2022-02-09
Allow carpet extensions to extend the settings manager with ease

fxmorin:
I want to be able to modify the settings manager & customize it. Thanks


fxmorin:
I'll wait for #1121 although this has been bugging me for a while...

PR #1324, opened 2022-02-10
adding structure block support

ch-yx:
Because of remap, create this new pr.
Hope all goes well.

### DIFF
--- a/src/main/java/carpet/script/api/WorldAccess.java
+++ b/src/main/java/carpet/script/api/WorldAccess.java
@@ -59,6 +59,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Random;
 import java.util.function.BiFunction;
 import java.util.function.BiPredicate;
 import java.util.function.Predicate;
@@ -66,6 +67,10 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import net.minecraft.commands.arguments.item.ItemInput;
+import net.minecraft.core.BlockPos;
+import net.minecraft.core.Direction;
+import net.minecraft.core.QuartPos;
+import net.minecraft.core.Registry;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.nbt.Tag;
 import net.minecraft.network.protocol.game.ClientboundExplodePacket;
@@ -75,6 +80,7 @@ import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.server.level.Ticket;
 import net.minecraft.server.level.TicketType;
 import net.minecraft.sounds.SoundSource;
+import net.minecraft.tags.TagContainer;
 import net.minecraft.util.Mth;
 import net.minecraft.util.SortedArraySet;
 import net.minecraft.world.Clearable;
@@ -107,6 +113,8 @@ import net.minecraft.world.level.biome.Climate;
 import net.minecraft.world.level.biome.MultiNoiseBiomeSource;
 import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.Blocks;
+import net.minecraft.world.level.block.Mirror;
+import net.minecraft.world.level.block.Rotation;
 import net.minecraft.world.level.block.SoundType;
 import net.minecraft.world.level.block.entity.BlockEntity;
 import net.minecraft.world.level.block.state.BlockState;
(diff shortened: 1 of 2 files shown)


src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -1585,7 +1597,134 @@ else if (ticket == TicketType.POST_TELEPORT) // post teleport
...
+                    return Value.NULL;
+                } // if(lv.size()==2&&lv.get(1).isNull()){lv3.remove(struident);return
+                  // Value.NULL;}
+                  theStructure = structureManager.getOrCreate(struident);
+            } else {
+                theStructure = new StructureTemplate();
+            }
+            BlockArgument start = BlockArgument.findIn((CarpetContext) c, lv, 1);
+            BlockArgument end = BlockArgument.findIn((CarpetContext) c, lv, start.offset);
+            BlockArgument ignoredblock = BlockArgument.findIn((CarpetContext) c, lv, end.offset + 1, true, true, false);
```
gnembon, line 1624: ignored block should for sure be optional - most usecases wouldn't require it, so please count your lv's and only read it when needed.

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -1585,7 +1597,134 @@ else if (ticket == TicketType.POST_TELEPORT) // post teleport
...
+            BlockPos size = endBlockPos.subtract(startBlockPos).offset(1, 1, 1);
+            // name,start,end, ignoreEntities, ignoredBlock,~~author~~,disk
+            theStructure.fillFromWorld(world,
+                    startBlockPos,
+                    size,
+                    !lv.get(end.offset).getBoolean(),
+                    ignoredblock.block == null ? null : ignoredblock.block.getBlockState().getBlock());
+            // lv6.setAuthor(lv.get(9).getString());//MC-140821
+            if (!returnnbt) {
+                if (lv.get(ignoredblock.offset).getBoolean()) {
```
gnembon, line 1643: same for saving parameter - honestly why it even is there? Why would you want to fill from world, and not save it? To save on write time? then you should use the structure and pass it to the variable?

Please give a usecase where save_structure('name', .... false) would ever make any sense?

(relying on structure manager to hold structures makes 0 sense to me by the way - you get variables for it).

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -1585,7 +1597,134 @@ else if (ticket == TicketType.POST_TELEPORT) // post teleport
...
+            StructureManager structureManager = world.getStructureManager();
+            StructureTemplate theStructure;
+            String name;
+
+            boolean returnnbt = lv.get(0).isNull();
+            ResourceLocation struident = null;
+            if (!returnnbt) {
+                name = lv.get(0).getString();
+                struident = ResourceLocation.tryParse(name);
+                if (name.isEmpty() || name.endsWith(":") || struident == null) {
```
gnembon, line 1614: seems like there is more to proper validation of structure names. maybe some convention should be used here that is acceptable on the client?

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -1585,7 +1597,134 @@ else if (ticket == TicketType.POST_TELEPORT) // post teleport
...
+            BlockArgument start = BlockArgument.findIn((CarpetContext) c, lv, 1);
+            // name,pos,
+            // ignoreEntities,integrity,awake,noupdate,fluid,gravity,rotation,mirror
+            StructurePlaceSettings lv4 = new StructurePlaceSettings();
+            lv4.clearProcessors();
+
+            lv4.setIgnoreEntities(start.offset <lv.size()?lv.get(start.offset).getBoolean():false)
+                    .setFinalizeEntities(start.offset + 2<lv.size()?lv.get(start.offset + 2).getBoolean():false)
+                    .setKnownShape(start.offset + 3<lv.size()?lv.get(start.offset + 3).getBoolean():true)
+                    .setKeepLiquids(start.offset + 4<lv.size()?lv.get(start.offset + 4).getBoolean():false)
```
gnembon, line 1689: please read and name all the options where you read arguments for clarity. 
It is very difficult to discern what is going on here - and the args don't match those available in the docs

src/main/java/carpet/script/api/WorldAccess.java
ch-yx: there are more checks in 'tryParse'.
i think that is enough.

I didn't do any more validation at all at first and just used `tryParse`. but was surprised to find out during testing that the path is empty actually bypasses this function's check. That's how I added the rule that it can't end with a colon.

src/main/java/carpet/script/api/WorldAccess.java
ch-yx: they are all optional now.

src/main/java/carpet/script/api/WorldAccess.java
ch-yx: i add some comments for the args .

src/main/java/carpet/script/api/WorldAccess.java
ch-yx: > relying on structure manager to hold structures makes 0 sense to me


Because I don't think it makes sense to serialize to/deserialize from a nbt value every time you load/unload. And it might be necessary to load content that the user saved manually with a structure block.

src/main/java/carpet/script/api/WorldAccess.java
gnembon: use InputValidator::identifierOf since its throws proper exception without you needing to handle it. 

If you have found some extra usecases, you can update InputValidator::identifierOf to include these extra cases

ch-yx: > use InputValidator::identifierOf since its throws proper exception without you needing to handle it.
> 
> If you have found some extra usecases, you can update InputValidator::identifierOf to include these extra cases

ok. i start using identifierOf and add a small check in identifierOf.

PR #1330, opened 2022-02-13
Settings rework and encapsulation

### LINKED ISSUE #1091: [Internal] Convert ParsedRule into an implementation detail

altrisi:
Tested full binary (and functionality) compatibility (as long as extensions weren't using hacks like mixing into private methods) via some extensions that update to snapshots (like TIS addition) and a mod I made for this: [ExampleMod.java](https://github.com/gnembon/fabric-carpet/files/8102287/ExampleMod.txt) and [MahRules.java](https://github.com/gnembon/fabric-carpet/files/8102385/MahRules.txt) (tested the mod against 22w06a, the version this PR was based on).

This PR is a complete overhaul to settings and rules. The most important change is that ParsedRule is no longer the only implementation, it's just an implementation of CarpetRule, which provides a base contract that other extensions can implement freely and add to their settings managers. All of the API classes related to settings have also been moved to a new `api` package to tell extensions they can use them freely and without binary compatibility risks. This PR also prepares the ground for Fallen's translation refactor PR (#1338), where rule descriptions and extra info no longer come from the annotation, but from the language file, therefore extensions that migrate to the new system must provide those via the language system (via CarpetExtension, up to each extension how that's implemented). This PR doesn't change the translation system but adds a small layer to it to allow that to work with legacy rules.

Enumeration of changes:
- Moved everything settings related to `carpet.api.settings` and separated class meanings from implementations (including moving standard validators to a new class instead of Validator)
- Extracted non-implementation dependant functionality of rules to a new class RuleHelper
- Add API methods to add arbitrary-implementation CarpetRules to SettingsManagers
- Deprecate rule observers depending on log4j and use a new RuleObserver interface instead
- Changed translation keys to have the carpet prefix, specify missing information to prevent deduplication (manager) and to use more readable names
- Added fallback paths to keep all of the existing systems from extensions working
- Fixed a few unhandled exceptions when parsing rule values
- Changed all Carpet usages (except Rule annotation) to the new systems and APIs
  This is the bulk of the PR files, 2-4 line changes for imports or ParsedRule->CarpetRule

For later PRs:
- Migrate all Carpet's rule annotations to the new system (the rest of Carpet uses the new system)

Ready for review!

Supersedes #1121, fixes #1091.


src/main/java/carpet/api/settings/SettingsManager.java
```diff
@@ -0,0 +1,781 @@
...
+     * @see SettingsManager#addRuleObserver(RuleObserver)
+     * @see SettingsManager#addGlobalRuleObserver(RuleObserver)
+     */
+    @FunctionalInterface
+    public static interface RuleObserver {
+        /**
+         * @param source The {@link CommandSourceStack} that likely originated this change, and should be the notified source for further
+         *               messages. Can be {@code null} if there was none and the operation shouldn't send feedback.
+         * @param changedRule The {@link CarpetRule} that changed. Use {@link CarpetRule#value() changedRule.value()} to get the rule's value,
+         *                    and pass it to {@link RuleHelper#toRuleString(Object)} to get the {@link String} version of it
```
gnembon, line 96: you have missed the point of the third parameter - the string value. Its not the machine translated and understood value set for this rule, but the actual string that player has typed.

I don't see a reason for removing it

altrisi: I wanted to keep it at the start, but ended up not doing so because a technical issue.

If you add the third parameter, then you have the original (fallback) `addRuleObserver(TriConsumer)` and the new `addRuleObserver(RuleObserver)`. While that's not a compile issue for the `SettingsManager` class as two methods can have the same name as long as they have different parameters, then consumers of the class can't call it with a lambda, as both the old and new have the same simple signature (3 params) and the method is ambiguous to the compiler. It may work if they specify the parameter types when declaring the lambda, but it makes it less easy to use, and would break source compatibility (existing binaries would be fine). Other alternatives were renaming the method, but I couldn't think of how to call it.

At the end of the day I thought it wasn't _that_ useful so I decided removing it was the easiest way (at the end of the day, this only fires if the rule was changed successfully, and then most of the time the value as a string will be the input).

How could I call the method then?

altrisi: Fixed by renaming to registerRuleObserver.

src/main/java/carpet/CarpetServer.java
```diff
@@ -43,7 +45,7 @@ public class CarpetServer // static for now - easier to handle all around the co
     public static MinecraftServer minecraft_server;
     private static CommandDispatcher<CommandSourceStack> currentCommandDispatcher;
     public static CarpetScriptServer scriptServer;
-    public static SettingsManager settingsManager;
+    public static SettingsManager settingsManager; // to change type to api type, can't change right now because of binary and source compat
```
gnembon, line 48: why not?

src/main/java/carpet/CarpetServer.java
```diff
@@ -92,7 +94,7 @@ public static void onServerLoaded(MinecraftServer server)
 
         settingsManager.attachServer(server);
         extensions.forEach(e -> {
-            SettingsManager sm = e.customSettingsManager();
+            var sm = e.extensionSettingsManager();
```
gnembon, line 97: why var? Why binary compat is important?

altrisi:
Binary compat -> Mods built depending on previous carpet versions work without having to be rebuilt targeting only newer versions.

In the case of `CarpetSettings.settingsManager`, this is because most extensions add their settings to it (I think this started because it's what extra does). Java doesn't only care about the name even if that's all you see, it's built against the exact type (`carpet.settings.SettingsManager`) so mods referencing this will error and crash with something like `NoSuchFieldError` (not even `Exception` as it's a `LinkageError`) and therefore not work.

Source compat in this case is because the deprecated methods from the old `SettingsManager` are not in the moved one, so if some extension was trying to call for example the old `CarpetSettings.settingsManager.addRuleObserver(log4j.TriConsumer)`, `...(ParsedRule)getRule`, `..notifyPlayersCommandsChanged`, etc, it would not only not work from a jar built against an older version, but also fail to compile directly when updating the version, without any indication to what to do.

And I think binary compat is important in this area because well, Carpet extensions (other than scarpet only extensions) _all_ use stuff around settings and carpet rules, so giving them warnings, instructions and time to update is nice.
And _all_ (but two or three max) use _directly_ `CarpetSettings.settingsManager`, so that's one to keep.

And tbh, because I like keeping binary compat while reasonable.

Changed the `var`s, there were just to not have the fully qualified name (`carpet.settings.SettingsManager`) and similar where there are two but it was obvious from the context.

gnembon:
Ok, I have a different opinion about it -, especially that external changes happened that breaks compat anyways (log4j, mc api changes), you as a mod that decided on dependency on carpet - might as well adjust anyways. Mod made for 1.18 doesn't have to work with 1.19 for example even if it could. I can see that extra baggage to keep for compat reasons as unnecessary maintenance. Moving stuff to its own package for these reasons is also something that I woudn't do.

Having said that - it is 'just' an opinion and someone else may have a different ones and they would all be valid. I can see your point of view and (while I think its unnecessary) I am accepting it, especially that you have done all this work to connect it all together. I will continue to review it with this assumption

src/main/java/carpet/CarpetSettings.java
```diff
@@ -57,29 +58,25 @@
     public static int updateSuppressionBlockSetting = -1;
 
     private static class LanguageValidator extends Validator<String> {
-        @Override public String validate(CommandSourceStack source, ParsedRule<String> currentRule, String newValue, String string) {
-            if (currentRule.get().equals(newValue) || source == null)
```
gnembon, line 61: was that necessary for startup? Why is that removed now?

src/main/java/carpet/CarpetSettings.java
```diff
@@ -803,42 +814,31 @@ public static void changeSpawnSize(int size)
...
                 return null;
             }
-            if (currentRule.get().intValue() == newValue.intValue())
+            if (currentRule.value().intValue() == newValue.intValue())
             {
                 //must been some startup thing
                 return newValue;
             }
             if (CarpetServer.minecraft_server == null) return newValue;
-
```
gnembon, line 829: nice change

src/main/java/carpet/api/settings/Rule.java
```diff
@@ -0,0 +1,82 @@
...
+    boolean strict() default true;
+    
+    /**
+     * If specified, the rule will automatically enable or disable 
+     * a builtin Scarpet Rule App with this name.
+     */
+    String appSource() default "";
+
+    /**
+     * The class of the validator checked when the rule is changed.
```
gnembon, line 54:
```suggestion
     * The class of the validator checked right before the rule is changed, using the returned value as the new value to set, or cancel the change if null is returned.
```

src/main/java/carpet/api/settings/Rule.java
```diff
@@ -0,0 +1,82 @@
...
+    Class<? extends Condition>[] conditions() default {};
+    
+    /**
+     * <p>Represents a condition that must be met for a rule to be registered in a {@link SettingsManager} via
+     * {@link SettingsManager#parseSettingsClass(Class)}</p>
+     * 
+     * @see Rule#conditions()
+     * @see #shouldRegister()
+     */
+    interface Condition {
```
gnembon, line 75: Can lambda be used in annotations? I woudn't be surprized if that's not the case, but technicall Condition is the same as BooleanSupplier, right?

src/main/java/carpet/api/settings/RuleHelper.java
```diff
@@ -0,0 +1,112 @@
...
+     * @param rule The {@link CarpetRule} to reset to its default value
+     * @param source A {@link ServerCommandSource} to notify about this change, or {@code null}
+     * 
+     * @param <T> The type of the {@link CarpetRule}
+     */
+    public static <T> void resetToDefault(CarpetRule<T> rule, CommandSourceStack source) {
+        try {
+            rule.set(source, rule.defaultValue());
+        } catch (InvalidRuleValueException e) {
+            throw new IllegalStateException("Rule couldn't be set to default value!", e);
```
gnembon, line 71: Thought - would that be even possible? Would be great to mention that situation in the javadoc above if that's the case.

src/main/java/carpet/CarpetSettings.java
altrisi: This is because I've tried to shape this a bit for fallen's translation rework system, and started moving a few things to translations (to not have half of it deprecated right after release).

Previously carpet would just use the fallbacks from code in default language, now we always use the language system (the legacy, deprecated fallbacks are injected into the language system when registering), so we have to init it even with the default value.

src/main/java/carpet/api/settings/Rule.java
altrisi: No, annotations can only reference constants, so you'd still have to make a full class implementing the lambda.
BooleanSupplier is the same, yes, but this explains what exactly it is, and doesn't really cost anything.

src/main/java/carpet/api/settings/RuleHelper.java
altrisi: `ParsedRule` by itself should never, a broken implementation may (broken validator maybe). It should never happen, but this is here so just in case it does, we don't continue with broken state, and the user hopefully gets useful debug from the exception. And the try-catch is necessary anyway because `InvalidRuleValueException` is checked.

(`CarpetRule#defaultValue()` contract states it must always be a valid value for `set`).

Part of the convenience of the method is hiding the try-catch.

src/main/java/carpet/api/settings/RuleHelper.java
```diff
@@ -0,0 +1,112 @@
...
+        return Translations.hasTranslation(key) ? Translations.tr(key) + String.format(" (%s)", rule.name()): rule.name();
+    }
+    
+    /**
+     * @param rule The {@link CarpetRule} to get the translated description of
+     * @return A {@link String} being the translated description of this rule, in Carpet's configured language.
+     * 
+     * @apiNote This method isn't stable API yet and may change or be removed in binary incompatible ways in later Carpet versions
+     */
+    @ApiStatus.Experimental
```
gnembon, line 96: I know its jetbrains and feels pretty standard, but log4j also felt a staple at the time and now we need to suffer due to autocomplete decision most likely. That annotation is probably only for information purposes, so maybe not really needed? I wish java had some standard common annotations for these purposes.

altrisi: Not really, I'll remove them.

Didn't think it was bad given it's only kept up to class data and discarded at runtime, Java ignores missing annotations if they were to be removed, and loom/mc/some mc dep adds it to the compile classpath anyway. And I think with it some (jetbrains) IDEs provide warnings about them when being referenced, which is nice.

In vanilla Java this would be `@Deprecated` iinw ("_discouraged from using_"), but I thought it didn't really express what it meant given it's normally used as "this will be removed".

src/main/java/carpet/api/settings/SettingsManager.java
```diff
@@ -0,0 +1,796 @@
...
+        }
+    }
+
+    private void writeSettingsToConf(ConfigFileData data)
+    {
+        if (locked)
+            return;
+        try (BufferedWriter fw = Files.newBufferedWriter(getFile()))
+        {
+            for (String comment : data.comments())
```
gnembon, line 380: That's why I think worrying about commands is not what we should worry at this point. If you have mixed in rules and comments they will be all smooshed together at this point. Its better IMO to be explicit that comments are preserved only when config is locked. If its locked its manually curated, so comments make sense. If they are automatically rendered, this means comments should not exist anyways, since its the game that edits them.

If we would want to do this properly saving the rules should be done via re-writing of the conf, skipping rules that are removed, updating rules that changed and appending new rules at the end if they are added. But that would definitely be a different PR

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -1495,7 +1497,7 @@ else throw new InternalExpressionException("Attacking entity needs to be a livin
 
             boolean success = WorldTools.createWorld(cc.s.getServer(), worldKey, seed);
             if (!success) return Value.FALSE;
-            CarpetServer.settingsManager.notifyPlayersCommandsChanged();
+            CommandHelper.notifyPlayersCommandsChanged(cc.s.getServer());
             return Value.TRUE;
```
gnembon, line 1501: no need to keep it - custom_dimension is not going back, at least until adding dimensions mid flight is not internally supported.

src/main/resources/assets/carpet/lang/en_us.json
```diff
@@ -1,54 +1,36 @@
...
-  "ui.change_permanently": "Change permanently",
-  "ui.click_to_keep_the_settings_in_%(conf)s_to_save_across_restarts": "click to keep the settings in %s to save across restarts",
-  "ui.rule_%(rule)s_will_now_default_to_%(value)s": "Rule %s will now default to %s",
-  "ui.rule_%(rule)s_not_set_restart": "Rule %s will now not be set on restart",
-  "ui.current_value": "current value"
+  // TODO Rules
+  
+  // Categories:
+
+  "carpet.settings.carpet.settings.category.bugfix": "bugfix",
```
gnembon, line 6: not too many carpet.settings?

src/main/java/carpet/utils/TypedField.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.commons.lang3.ClassUtils;
+
+/**
+ * <p>A wrapper around a {@link Field} that uses a {@link VarHandle} and {@link MethodHandle}s to retrieve and set the contents of the field</p>
+ * 
+ * <p>It keeps the memory semantics of {@code volatile} if the field had that modifier.</p>
+ *
+ * @param <T> The {@link Field}'s type
+ */
+public final class TypedField<T> {
```
gnembon, line 18: Not sure about the use of this class

src/main/java/carpet/script/api/WorldAccess.java
altrisi: Change was made before it got commented, I think it's out of scope to remove it here.

src/main/java/carpet/utils/TypedField.java
altrisi: I originally wanted to use it for some other stuff, but ended up just for this, true that it's not that useful for just one thing, I'll revert it to a regular field.

The thing was that this class not only kept the `<T>` instead of returning/accepting an untyped `Object`, but also checked access and everything at creation, and given it used Method/VarHandle it should be a tiny bit faster (though it's not used in any performance-heavy area so it doesn't matter).

src/main/java/carpet/api/settings/SettingsManager.java
altrisi: I totally forgot I kept this in here when moving the class between packages, I'll revert it back.

altrisi: Removed saving comments for later, kept ignoring them so it can work for locked configs and default files.

altrisi:
Ran extra testing, fixed last issues/tweaks, I'd say this is ready to merge!

PR #1331, opened 2022-02-13
allows rendering arbitrary polygon

ch-yx:
图片
have those special arguments
* `points` is a list of points.
* argument "mode" will decide how to use argument "points"

if mode == polygon, 
图片
if mode == strip
图片
if mode == triangles
图片
* argument "relative" is a list of bool. it is the used in follow mode. (which vertex going with the entity)

~i have to make a alternative method for un-carpet client next.~ i have finished it.
~~**this mode is only for faces now.  if you want to have edges, use `line`.**~~ i have finished it too.

### DIFF
--- a/src/main/java/carpet/script/utils/ShapeDispatcher.java
+++ b/src/main/java/carpet/script/utils/ShapeDispatcher.java
@@ -648,7 +648,7 @@ public class ShapeDispatcher
             return hash;
         }
         ArrayList<Vec3> alter_point=null;
-        Random random=new Random();
+        final Random random=new Random();
         ArrayList<Vec3> alter_point(ServerPlayer p){
             if (alter_point!=null){
                 return alter_point;
@@ -738,10 +738,11 @@ public class ShapeDispatcher
         }
         @Override
         public Consumer<ServerPlayer> alternative() {
+            final ParticleOptions locparticledata = getParticleData(String.format(Locale.ROOT ,"dust %.1f %.1f %.1f %.1f", fr, fg, fb, fa));
             return p->{
+                if(p.level.dimension() != this.shapeDimension){return;}
                 for(Vec3 v : alter_point(p)){
-                    if(p.level.dimension() != this.shapeDimension){continue;}
-                    p.getLevel().sendParticles(p, getParticleData(String.format(Locale.ROOT ,"dust %.1f %.1f %.1f %.1f", fr, fg, fb, fa)), true,
+                    p.getLevel().sendParticles(p,locparticledata , true,
                     v.x, v.y, v.z, 1,
                     0.0, 0.0, 0.0, 0.0);
                 }
@@ -773,6 +774,8 @@ public class ShapeDispatcher
             inneredges = options.getOrDefault("inneredges",optional.get("inneredges")).getBoolean();
             if(vertex_list.size()<3){
                 throw new IllegalArgumentException("Unexpected vertex list size: " + vertex_list.size());
+            }else if(vertex_list.size()<4){
+                inneredges=false;
             }
             if("polygon".equals(_mode)){
                 mode=VertexFormat.Mode.TRIANGLE_FAN;
--- a/src/main/java/carpet/script/utils/ShapeDispatcher.java
+++ b/src/main/java/carpet/script/utils/ShapeDispatcher.java
@@ -126,7 +126,7 @@ public class ShapeDispatcher
         List<ServerPlayer> alternativePlayers = new ArrayList<>();
         for (ServerPlayer player : players)
         {
-            if (ServerNetworkHandler.isValidCarpetPlayer(player))
+            if (ServerNetworkHandler.isValidCarpetPlayer(player)&&!CarpetSettings.superSecretSetting)//for debug
             {
                 clientPlayers.add(player);
             }
@@ -741,6 +741,7 @@ public class ShapeDispatcher
             final ParticleOptions locparticledata = getParticleData(String.format(Locale.ROOT ,"dust %.1f %.1f %.1f %.1f", fr, fg, fb, fa));
             return p->{
                 if(p.level.dimension() != this.shapeDimension){return;}
+                if(!(fa>0.0f)){return;}
                 for(Vec3 v : alter_point(p)){
                     p.getLevel().sendParticles(p,locparticledata , true,
                     v.x, v.y, v.z, 1,
--- a/src/main/java/carpet/script/utils/ShapeDispatcher.java
+++ b/src/main/java/carpet/script/utils/ShapeDispatcher.java
@@ -647,9 +647,105 @@ public class ShapeDispatcher
             hash ^= Boolean.hashCode(inneredges);    hash *= 1099511628211L;
             return hash;
         }
+        ArrayList<Vec3> alter_point=null;
+        Random random=new Random();
+        ArrayList<Vec3> alter_point(ServerPlayer p){
+            if (alter_point!=null){
+                return alter_point;
+            }
+            alter_point=new ArrayList<>();
+            switch (mode) {
+                case TRIANGLES:
+                    for (int i=0;i<vertex_list.size();i++){
+                        Vec3 vecA=vertex_list.get(i);
+                        if(relative.get(i)){
+                            vecA=relativiseRender(p.level, vecA, 0);
+                        }
+                        i++;
+                        Vec3 vecB=vertex_list.get(i);
+                        if(relative.get(i)){
+                            vecB=relativiseRender(p.level, vecB, 0);
+                        }
+                        i++;
+                        Vec3 vecC=vertex_list.get(i);
+                        if(relative.get(i)){
+                            vecC=relativiseRender(p.level, vecC, 0);
+                        }
+                        alter_draw_triangles(vecA, vecB, vecC);
+                    }
+                    break;
+                case TRIANGLE_FAN:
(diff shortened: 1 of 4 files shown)


src/main/java/carpet/script/utils/ShapeDispatcher.java
```diff
@@ -625,6 +629,69 @@ public static int particleMesh(List<ServerPlayer> playerList, ParticleOptions pa
...
+            long hash = super.calcKey();
+            hash ^= 2;                     hash *= 1099511628211L;
+            hash ^= Byte.hashCode(mode);   hash *= 1099511628211L;
+            hash ^= going.hashCode();      hash *= 1099511628211L;
+            for (Vec3 i :vertex_list){
+                hash ^= vec3dhash(i);
+                hash *= 1099511628211L;
+            }
+            hash ^= vertex_list.size();    hash *= 1099511628211L;
+            hash ^=(vertex_list.size()+37)*(vertex_list.size()+3721)*4224696537L+30692*vertex_list.size();
```
gnembon, line 645: why this?

gnembon:
docs are missing - kinda important point. 

Unclear use of 'going' parameter - name is also weird and unclear as well how it interfaces with 'snap' and 'follow' - you should explain that in the docs.

I like it is flexible with mixin stationary and relative vertices, but it needs to be communicated in the docs. Name is also meh. 'relative' or 'rel' would be better I think

src/main/java/carpet/script/utils/ShapeDispatcher.java
```diff
@@ -625,6 +629,69 @@ public static int particleMesh(List<ServerPlayer> playerList, ParticleOptions pa
...
+            return hash;
+        }
+        @Override
+        public Consumer<ServerPlayer> alternative() {
+            return p->{};
+        }
+        private final Set<String> required = Set.of("vertex");
+        private final Map<String, Value> optional = Map.ofEntries(
+            entry("going",Value.NULL),
```
gnembon, line 653: same 'going' -> 'rel' or 'relative' (if follow not defined it would mean all false, otherwise all true by default)

src/main/java/carpet/script/utils/ShapeDispatcher.java
```diff
@@ -625,6 +629,69 @@ public static int particleMesh(List<ServerPlayer> playerList, ParticleOptions pa
...
+            entry("mode", new StringValue("polygon"))
```
gnembon, line 654: 'mode' -> 'type'?

src/main/java/carpet/script/utils/ShapeDispatcher.java
ch-yx: removed.
some excitement came on as I was writing.🙃🤦‍♂️

src/main/java/carpet/script/utils/ShapeDispatcher.java
```diff
@@ -858,6 +925,33 @@ public long calcKey()
     public static abstract class Param
     {
         public static Map<String, Param> of = new HashMap<String, Param>(){{
+            put("vertex", new PointsParam("vertex"){
```
gnembon, line 927: Points param is unused, so you can change it to not use expensive x, y, z fields

src/main/java/carpet/script/utils/ShapeDispatcher.java
```diff
@@ -625,6 +629,68 @@ public static int particleMesh(List<ServerPlayer> playerList, ParticleOptions pa
         }
     }
 
+    public static class Polyface extends ExpiringShape
+    {
+        @Override
+        public long calcKey(){
+            long hash = super.calcKey();
+            hash ^= 2;                     hash *= 1099511628211L;
```
gnembon, line 637: use a unique ID here - 2 is used by a line. not that it would matter.

I need to write a proper builder for this hasher

src/main/java/carpet/script/utils/ShapesRenderer.java
```diff
@@ -386,7 +387,34 @@ public void renderLines(PoseStack matrices, Tesselator tessellator, BufferBuilde
...
+    public static class RenderedPolyface extends RenderedShape<ShapeDispatcher.Polyface>
+    {
+        public RenderedPolyface(Minecraft client, ShapeDispatcher.ExpiringShape shape)
+        {
+            super(client, (ShapeDispatcher.Polyface)shape);
+        }
+        @Override
+        public void renderFaces(Tesselator tessellator, BufferBuilder bufferBuilder, double cx, double cy, double cz, float partialTick)
+        {
+                bufferBuilder.begin(switch(shape.mode){
```
gnembon, line 399: you should decode the mode in the constructor so you don't have so switch every render

src/main/java/carpet/script/utils/ShapesRenderer.java
```diff
@@ -386,7 +387,34 @@ public void renderLines(PoseStack matrices, Tesselator tessellator, BufferBuilde
...
+                    if(shape.relative.get(i)){
+                        vec=shape.relativiseRender(client.level, vec, partialTick);
+                    }
+                    bufferBuilder.vertex(vec.x()-cx, vec.y()-cy, vec.z()-cz).color(shape.fr, shape.fg, shape.fb, shape.fa).endVertex();
+                }
+                tessellator.end();
+                
+        }
+        @Override
+        public void renderLines(PoseStack matrices, Tesselator tessellator, BufferBuilder builder, double cx, double cy,
```
gnembon, line 415: Don't forget to render lines too - they accept different colors and should be rendered independently from faces.

src/main/java/carpet/script/utils/ShapeDispatcher.java
```diff
@@ -625,6 +629,68 @@ public static int particleMesh(List<ServerPlayer> playerList, ParticleOptions pa
...
+            for (Vec3 i :vertex_list){
+                hash ^= vec3dhash(i);
+                hash *= 1099511628211L;
+            }
+            hash ^= vertex_list.size();    hash *= 1099511628211L;
+            return hash;
+        }
+        @Override
+        public Consumer<ServerPlayer> alternative() {
+            return p->{};
```
gnembon, line 649: well, draw the lines at least

src/main/java/carpet/script/utils/ShapesRenderer.java
ch-yx: done.

src/main/java/carpet/script/utils/ShapesRenderer.java
ch-yx: done.

src/main/java/carpet/script/utils/ShapeDispatcher.java
```diff
@@ -647,9 +647,105 @@ public long calcKey(){
...
+            }
+            
+        }
         @Override
         public Consumer<ServerPlayer> alternative() {
-            return p->{};
+            return p->{
+                for(Vec3 v : alter_point(p)){
+                    if(p.level.dimension() != this.shapeDimension){continue;}
+                    p.getLevel().sendParticles(p, getParticleData(String.format(Locale.ROOT ,"dust %.1f %.1f %.1f %.1f", fr, fg, fb, fa)), true,
```
ch-yx, line 744: what did i just write...... why i put that check in the loop???

src/main/java/carpet/script/utils/ShapeDispatcher.java
ch-yx: done.

src/main/java/carpet/script/utils/ShapeDispatcher.java
ch-yx: done

src/main/java/carpet/script/utils/ShapeDispatcher.java
ch-yx: I personally feel that mode is a little better. It is a little more specific.

src/main/java/carpet/script/utils/ShapeDispatcher.java
ch-yx: ah i ve already fixed it

ch-yx:
i d like to repost some image of this pr from
https://discord.com/channels/882822986795716608/942620305589960715/942710228225523742

polygon mode.with and without inner edge:
图片
图片
strip mode.with and without inner edge:
图片
图片
particle ver. for un-carpet user:

src/main/java/carpet/script/utils/ShapesRenderer.java
```diff
@@ -386,7 +390,155 @@ public void renderLines(PoseStack matrices, Tesselator tessellator, BufferBuilde
...
+            if(shape.fa==0){return;}
+            if (shape.doublesided)
+                RenderSystem.disableCull();
+            else
+                RenderSystem.enableCull();
+            
+            RenderSystem.depthMask(false);
+            RenderSystem.depthFunc(516);
+            RenderSystem.enableBlend();
+            RenderSystem.blendFunc(GlStateManager.SourceFactor.ONE_MINUS_DST_ALPHA, GlStateManager.DestFactor.DST_ALPHA);
```
ch-yx, line 411: should i use `blendFuncSeparate` here instead? 
Not Sure.

If no one makes a complaint, I won't change it.

ch-yx:
note: contained by https://github.com/gnembon/fabric-carpet/pull/1365 now.

ch-yx:
as it is included in that pr, i ll close this one.

PR #1332, opened 2022-02-13
Allow Messenger to create `TranslatableComponent`s and minor refactor to make it more flexible for future additions

altrisi:
(badly copied description from a Discord message because I'm lazy to make a proper PR body rn, even more since this is likely going to not get in anyway in favour of ch-yx PR)

I made an internal message composer that has three separate components, what it will apply events to, what will be done with the next component and what will provide a new ready to append component (and whether it's ready). For the api it works the same way, just if you pass it a translatable (starting with j prefix as proposed in the original issue) then that and the following n components (n being the variables it takes) will go to the translatable.

### DIFF
--- a/src/main/java/carpet/utils/Messenger.java
+++ b/src/main/java/carpet/utils/Messenger.java
@@ -4,12 +4,14 @@ import net.minecraft.ChatFormatting;
 import net.minecraft.Util;
 import net.minecraft.commands.CommandSourceStack;
 import net.minecraft.core.BlockPos;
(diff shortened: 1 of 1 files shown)


gnembon:
not sure if that is meant to be used as a reference, or to be merged? I assume its just an idea for ch-yx

altrisi:
Yes, I was asked for how my implementation was on discord: https://discord.com/channels/882822986795716608/882826588348092457/942444797988311050

PR #1333, opened 2022-02-14
Print a warning when an extension registers late

altrisi:
This PR makes Carpet print a warning if extensions register themselves late, with the idea of (maybe) later making that throw an exception and later just not have the fallback if they register late.

There's multiple methods extensions can use to register themselves in time, they should make sure they are registering correctly IMO and Carpet shouldn't be checking that.

### DIFF
--- a/src/main/java/carpet/CarpetServer.java
+++ b/src/main/java/carpet/CarpetServer.java
@@ -66,6 +66,8 @@ public class CarpetServer // static for now - easier to handle all around the co
         if (currentCommandDispatcher != null)
         {
             extension.registerCommands(currentCommandDispatcher);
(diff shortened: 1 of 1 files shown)


gnembon:
Maybe it should actually include a message what to do, i.e. add mod dependency on carpet - because that's what's its all about right?

gnembon:
so when a developer sees it, they would know exactly what to do.

gnembon (approved):

altrisi:
> Maybe it should actually include a message what to do, i.e. add mod dependency on carpet - because that's what's its all about right?

No, if they don't have a dependency on Carpet they can't call that method, this is just if someone is calling it later than server startup.

PR #1334, opened 2022-02-14
Remove log4j as much as possible

altrisi:
This PR removes as many log4j references as possible from the Carpet codebase, given it has been noted by some people that it could be swapped via the launcher at a later date, given all references to it have been changed to slf4j in Minecraft, not referencing log4j at all anymore.

Remaining issues:
- Scarpet's `logger` supported `fatal` log level, slf4j doesn't. How can that be deprecated there? Impure functions don't take a context to get a host to issue deprecations on, and I don't know of the equivalences or a way to do it
- SettingsManager API relies on one log4j type (TriConsumer), in a method that is actually relatively used by extensions. This will be solved with the settings refactor in #1330 and the deprecated versions are removed, given observers are changed to use their own type, `RuleObserver`.

This PR also prepares a loom flag that completely removes log4j from the compile classpath to both assert that no code is referencing it and as a bonus make IDEs not suggest it either. However, given SettingsManager API depends on it, it isn't turned on yet.


src/main/java/carpet/script/api/Auxiliary.java
```diff
@@ -830,11 +830,14 @@ else if(lv.size()==2)
...
                 res = lv.get(1);
                 switch(level){
-                    case "error": CarpetScriptServer.LOG.error(res.getString()); break;
-                    case "warn":  CarpetScriptServer.LOG.warn(res.getString());  break;
                     case "debug": CarpetScriptServer.LOG.debug(res.getString()); break;
-                    case "fatal": CarpetScriptServer.LOG.fatal(res.getString()); break;
+                    case "warn":  CarpetScriptServer.LOG.warn(res.getString());  break;
                     case "info":  CarpetScriptServer.LOG.info(res.getString());  break;
+                    case "fatal":
+                        // Somehow issue deprecation
```
gnembon, line 837: yeah, go for it.

build.gradle
```diff
@@ -13,6 +13,7 @@ group = project.maven_group
 
 loom {
 	accessWidenerPath = file("src/main/resources/carpet.accesswidener")
+	runtimeOnlyLog4j = false // Can be turned on after TriConsumer is no longer present in public API
```
gnembon, line 16: what this have to do with the TriConsumer in Setting Manager?

gnembon:
Awesome,

If it breaks compatibility, maybe like with settings rework - wait for 1.18.2 and do this after. I don't think we need to maintain 1.18.2 for longer than one or two releases.

altrisi:
This PR doesn't, I just mention that it is not fully gone yet (other than not being able to print FATAL via scarpet, but that's probably uncommon enough and doesn't break compat, just logs as error instead for now). Log4j's TriConsumer is kept for now.

build.gradle
gnembon: Oh, didn't realized that it was from log4j. I would say burn it with fire? question when.

altrisi: Neither did I, until gradle complained.

Yeeting it right now would make any already compiled extension that uses rule observers crash on launch, I think it's fine to wait a bit while it's deprecated and then removed.

gnembon (approved):

PR #1336, opened 2022-02-16
Update Full.md

LilTheFool:
Added a ToC for accessibility


altrisi:
Small thing:
Full.md is automatically generated from the rest, this would get removed on updates if not added via the script that generates it.

However, Github already provides a (better, dynamic and searchable) table of contents in the top left of the file header, so I don't think this is really needed.

gnembon:
It is very exhaustive though - this is a nice summary. I woudn't trash it - maybe it would be beneficial to add this to the top of the overview - problem is maintainability, and this looses badly here. Thoughts?

PR #1338, opened 2022-02-16
Added server-side translation system

Fallen-Breath:
## What current translaion system does

Current carpet's translation system translate string based on the rule value of `language`. Everyone in the same server will received the same language carpet is translating to. It also doesn't support text components as trasnlation parameters

## What will this PR do

1. Implement a complete server-side translation system that applies translation based on player's client setting language and support text component as translation parameter
All translaion action happen on the server side so the clients needs 0 modification to read those translated texts. What it will do is iterating through the text components and its siblings, replaced all `TranslatableComponent` with translated text components, before sending to a player

2. Deprecate the `desc` and `extra` fields in `@Rule` annotation and migrate them all into `en_us.json` translation file. `desc` and `extra` will still funtions as a backup translation for `en_us` for extention backwards compatibility

3. Replaced the previous String-based translation with the new TextComponent-based translation (It's now only used in `SettingsManager` for `/carpet` commands), and apply translation to more places

## Progress

- [x] Server-side translation framework
- [x] Fallback rule translation logic of the `desc` and `extra` fields in `@Rule`
- [ ] Migrate carpet rule translation texts into `en_us.json
- [ ] Apply translation to all carpet commands
  - [x] `/carpet` command
- [ ] (?) Scarpet API support (I don't know how do scarpet things tho) 

## To be discuss

There are things to be discussed for this PR. One is for the changes made to `Messenger` in this PR. I'm not sure if there are any better implementation of this

Another is for translation key format. A conclusion should be made before I start to apply translation to all carpet commands

### Messenger

Besides adding a `tr` method for easier `TranslatableComponent` creation (just like `s`), this PR also changes some implementation logic of `getChatComponentFromDesc` and `c`

It's required for converting the let `Messenger#c` supports formatting text component using its `desc` strings. Text component support is necessary since now we use `TranslatableComponent` for server-side translation instead of simple strings now, and we needs to translate stuffs like these using text component:

```java
Messenger.c("w some info " + name, "rb another text")
Messenger.c("w some text", String.format("^g hover text with parameter %s", param))
```

Quick view of what new you can do now:

- `Messenger.c("rb", someText)`, where `someText` is a text component and it will change it's style to red and bold
- `Messenger.c(baseText, "^", hoverText)`, where `hoverText` will be used as the hovering text of `baseText`

### `Messenger#c`

1. You can pass a style modifier string singly to apply style to the next object. For example, the following codes generates the same text:

```java
Messenger.c("g text")
Messenger.c("g", "text")
Messenger.c("g", Messenger.s("text"))
```

2. You can use a single "^" as a modifier and let the next object to become a hovering text of the previous text. For example, the following codes generates the same text:

```java
Messenger.c("w text", "^r hover")
Messenger.c("w text", "^", "r hover")
Messenger.c("w text", "^", Messenger.s("hover", "r"))
```

3. Strings with no space charactor inside will be recognized as a `desc` formatting modifier for the next object. This might causing some behavior changes if someone is using `Messenger.c("abc")`, but passing a string without space character into `Messenger#c` is already not correct isn't it?

#### `Messenger#getChatComponentFromDesc`

Let's says for a input string "wb sometext", "wb" is the `desc` and "sometext" is the `message`

1. Now it will not append a "w" prefix if the `desc` is empty. So now you can create a text without explicitly declare it to be white. Text with no explicitly color declaring will inherit it's parent component's color, so a single `desc` can apply to all components in a composed text
2. For input string with a space character as the prefix, no formatting will be applied to

### Translation key format

Should we add an `carpet.` prefix or something to all or any of translation entries? Although with fabric carpet only, the translation files will not be accessed by outside carpet, but if fabric-resource-loader is loaded then the translation file will be applied in the game globally, which means that there's a chance of translation key conflict. 

I'll suggest not to change rule translations though, since it might break translation backwards compatibility for extensions, and prefix `rule.someRuleName` is not that likeyly to be used somewhere else


Fallen-Breath:
Here's an idea for the translation key format:

- All entries except rule translations will have the prefix `carpet.`. If an extension passed a translation key without `carpet.` prefix then a warning will be thrown and the `carpet.` prefix will be added automatically
- Entries with key starts with `//` will be ignored when storing
- For rules, named like `<settingManagerId>.rule.ruleName.desc`. For fabric carpet the `<settingManagerId>` will be `carpet`. So if extension passed a rule translation without an settingManagerId prefix (by testing if it simply starts with `rule.`) and carpet's setting manager id (`carpet`) will be used as fallback
- For categroies, named like `<settingManagerId>.category.categoryName`. Backwards compatiblity logic is the same as rule translations
- For carpet commands, named like `carpet.command.<commandPrefix>.blablabla`
- For carpet loggers, named liked `carpet.logger.<loggerName>.blabla`
- For extensions, it's suggested to named their translation entiry keys as `<modId>.blabla`. People can use whatever format they want, and carpet's server-side translation system is capable to handle them well

Fallen-Breath:
By applying translation logic to `carpet.settings.SettingsManager#printAllRulesToLog`, It's now possible to dump setting ducument in any language via single `/carpet language` command

Fallen-Breath:
Tested and it's compatible with Carpet-TIS-Addition without any changes required. All received legacy translation key and `@Rule` fields (`desc`, `extra`) from the extension are transformed and all work correctly

PR #1340, opened 2022-02-17
Added lazydfu mod in dev runtime for faster startup

Fallen-Breath:
Data fixer initialization loves to eat all cpu during Minecraft startup (100% CPU usage for >1min on my i7-8750H) and that's really annoying

Here comes the fix for that: [LazyDFU](https://github.com/astei/lazydfu) mod at runtime

## Compatibility

LazyDFU only has 1 single mixin and it works from 1.14 to current 1.18.2 snapshots, and it's likely to work in future versions without changes

## Stability

For the mod, LazyDFU changed nothing about game behaviors

For the maven, jitpack is used as the maven repository of LazyDFU which is stable

## Dependency

If somehow LazyDFU doesn't work anymore in the future, you can simply removed it from runtime (basically revert this commit) and nothing will get hurted

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,6 +4,12 @@ plugins {
 	id 'maven-publish'
 }
 
+repositories {
+	maven {
+		url 'https://www.jitpack.io'
+	}
+}
+
 sourceCompatibility = JavaVersion.VERSION_17
 targetCompatibility = JavaVersion.VERSION_17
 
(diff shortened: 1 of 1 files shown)


altrisi:
Why not just put it in your mods folder?

Fallen-Breath:
fair

PR #1345, opened 2022-02-19
Fix a few IDE warnings

altrisi:
This PR fixes multiple IDE warnings, such as raw types, unused/unnecessary code/imports or serializable warnings (because of double brace map init). Tried not to touch files relating worldgen because of the changes going on in snapshots.

Map.ofEntries() is also more memory efficient than HashMap, so why not (same for Set.of IIRC, likely even higher savings).

Was mostly a bit bored.


src/main/java/carpet/helpers/FeatureGenerator.java
```diff
@@ -63,7 +63,7 @@
 
 public class FeatureGenerator
 {
-    public static final Object boo = new Object();
+	public static final Object boo = new Object();
```
Fallen-Breath, line 66: tab?

altrisi: How did that happen?????

src/main/java/carpet/mixins/ServerChunkCacheMixin.java
```diff
@@ -53,7 +53,7 @@ private int setupTracking(DistanceManager chunkTicketManager)
             SpawnReporter.first_chunk_marker = new HashSet<>();
             for (MobCategory cat : MobCategory.values())
             {
-                Pair key = Pair.of(dim, cat);
+                var key = Pair.of(dim, cat);
```
gnembon, line 56: please don't use vars. You an use ide to resolve the types for you.

PR #1353, opened 2022-03-04
Fixed execution offset in screens callback context

BisUmTo:
Wrong origin in screen callbacks


gnembon:
Why?

BisUmTo:
scan()
rect()
diamond()

Don't work correctly.

More info https://discord.com/channels/882822986795716608/882825965108068412/949430483081637938

altrisi:
Wondering if `ScreenValue` should instead save the origin of the creating context, in case the app was running under a different one. But probably not worth it given there's discussion on custom origins being removed anyway, and it can always be changed later if it becomes an issue.

PR #1356, opened 2022-03-09
Fix more warnings and do some small optimizations

altrisi:
No functional differences.

This PR fixes some more IDE warnings and applies some small (very small most of them) optimizations.

It's also a bit of a follow-up to #1009 in that it changes some map/set initializations to use newer syntax.

The optimizations are mostly memory, here's all the changes:
- `Set.of()` (and `Collectors.toUnmodifiableSet()`) use less memory than `HashSet` (or `Sets.newHashSet()`)
- `Set.of`/`List.of()` is also more efficient with a small number of elements (<= 2 iinw) given it doesn't even allocate an array, unlike some of the previously used solutions

  Those also use constant arrays (hotspot implementation only feature), so they _can_ have better performance (if the jit decides to optimize them).
- ~`new NumericValue()` is primitive specialized, and while it doesn't improve runtime performance given it only runs once where it has been replaced, given I changed the ones for integer values those will have them available for possible optimizations~ This is now only used in one place
- Enums already save their names, so I changed the enum around MobAI to use it directly, and therefore byName is no longer needed given enums have a method for that as long as you toUpperCase the input
- DataStructures: Changed `decode_json` to use the deserializer via gson, as it should be used, instead of nulling out parameters that could be needed at some point
- Made vanilla slot result argument map be primitive specific for better performance and memory usage and made sure it doesn't reuse the same mutable result for multiple apps.
  New: It now also uses a new (more compact) data type for storage in the map that creates a list instead of saving composed Value instances.
- Changed some tabs to spaces I found in WoolTool when fixing mine (tell me if I should revert those)


src/main/java/carpet/script/CarpetEventServer.java
```diff
@@ -983,11 +982,9 @@ public static String getEntityLoadEventName(EntityType<? extends Entity> et)
         }
 
         @Deprecated
-        public static final Map<EntityType<? extends Entity>, Event> ENTITY_LOAD= new HashMap<>()
-        {{
-            EntityType.byString("zombie");
```
gnembon, line 988: I encountered class-loading errors unless entity type was called statically like here. This might have changed

PR #1360, opened 2022-03-13
Docs edit, noted double backslash requirement for some regex operators

rv3r:
It appears that `replace` and `replace_first` require double backslashes to use some regex operators. Since I don't think this is a bug, I've added a note about it to the description of the functions in the docs. I also added an example that is similar to the originals. I don't know if docs have a style guide, but if they do, I probably failed to follow it.


James103:
Full.md is automatically generated from the component docs files.
Please make the change in `docs/scarpet/language/SystemFunctions.md` instead.

rv3r:
Can do. Is there a note about that somewhere?

altrisi:
I don't think so... You can only kind-of assume it by finding the `merge-docs` scripts at the root or by looking at the release actions, but nowhere else really.

PR #1361, opened 2022-03-13
Docs edit, noted double backslash requirement for some regex operators

rv3r:
It appears that `replace` and `replace_first` require double backslashes to use some regex operators, probably because it seems that Java also requires this. I've added a note about it to the description of the functions in the docs and also added an example that is similar to the originals. If this requirement is considered a bug and is fixed soon, this pr can be ignored.

### DIFF
--- a/docs/scarpet/language/SystemFunctions.md
+++ b/docs/scarpet/language/SystemFunctions.md
@@ -206,9 +206,11 @@ title('aBc') => 'Abc'
 
 Replaces all, or first occurence of a regular expression in the string with `repl` expression, 
 or nothing, if not specified
+or nothing, if not specified. To use escape characters(`\(`,`\+`,...), metacharacters(`\d`,`\w`,...), or position anchors(`\b`,`\z`,...) in your regular expression, use two backslashes
 
 <pre>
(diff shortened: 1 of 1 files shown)


docs/scarpet/language/SystemFunctions.md
```diff
@@ -206,9 +206,11 @@ title('aBc') => 'Abc'
 
 Replaces all, or first occurence of a regular expression in the string with `repl` expression, 
 or nothing, if not specified
+or nothing, if not specified. To use escape characters(`\(`,`\+`,...), metacharacters(`\d`,`\w`,...), or position anchors(`\b`,`\z`,...) in your regular expression, use two backslashes
```
James103, line 209:
```suggestion
or nothing, if not specified. To use escape characters (`\(`,`\+`,...), metacharacters (`\d`,`\w`,...), or position anchors (`\b`,`\z`,...) in your regular expression, use two backslashes.
```

ch-yx:
isn't that more misleading than before…

BisUmTo (approved):

ch-yx:
ah.... i think is it better said somethink like:
`do not forget to escape escape characters in string literal again, which means that you may want to use double....`

PR #1364, opened 2022-03-13
add some getter/setter as requests in discord.

ch-yx:
this pr adds following getter/setter

~~carrying_item :   cursor stack  replaced by `get_current_screen`~~
skull_dangerous :    skulls blueness
offer_flower : ironman have flower or not
item : could works on itemframs as well
item_display_name : display name of item.
real_light : real light

### DIFF
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -268,6 +268,9 @@ Example usages:
   // not a problem in apps
 </pre>
 
+### `item_display_name(item, hasbracket?)`
+ returns the `display_name` of the item as a Text Value. if hasbracket is true, there will be a bracket around it.
+
 ### `display_title(players, type, text?, fadeInTicks?, stayTicks?, fadeOutTicks),`
 
 Sends the player (or players if `players` is a list) a title of a specific type, with optionally some times.
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -268,11 +268,11 @@ Example usages:
   // not a problem in apps
 </pre>
 
-### `item_display_name(item )`
+### `item_display_name(item)`
  Returns the name of the item as a Text Value. `item` should be a list of `[item_name, count, nbt]`, or just an item name.
  
- Please note that it is a translated value, and the result of it converting to a string will using your servers language, but when you use print() to broadcast it to players, it will use each player's own language.
-
+ Please note that it is a translated value. treating it like a string (eg.slicing, breaking, changing its case) will turn it back into a normal string without translatable properties. just like a colorful formatted text loose its color. And the result of it converting to a string will use en-us (in a server) or your single player's language, but when you use print() or others functions that accept a text value to broadcast it to players, it will use each player's own language.
+ 
  If the item is renamed, it will also be reflected in the results.
 
 
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -270,7 +270,9 @@ Example usages:
 
 ### `item_display_name(item )`
  Returns the name of the item as a Text Value. `item` should be a list of `[item_name, count, nbt]`, or just an item name.
- Please note that it is a translated value, and the result of it converting to a string may be affacted by your language.
+ 
+ Please note that it is a translated value, and the result of it converting to a string will using your servers language, but when you use print() to broadcast it to players, it will use each player's own language.
+
  If the item is renamed, it will also be reflected in the results.
 
 
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -268,7 +268,7 @@ Example usages:
   // not a problem in apps
 </pre>
 
-### `item_name_text(item )`
+### `item_display_name(item )`
  returns the name of the item as a Text Value. `item` should be a list of `[item_name, count, nbt]`, or just an item name.
 
 
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -269,7 +269,8 @@ Example usages:
 </pre>
 
 ### `item_name_text(item )`
- returns the name of the item as a Text Value. 
+ returns the name of the item as a Text Value. `item` should be a list of `[item_name, count, nbt]`, or just an item name.
+
 
 ### `display_title(players, type, text?, fadeInTicks?, stayTicks?, fadeOutTicks),`
 
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -268,8 +268,8 @@ Example usages:
   // not a problem in apps
 </pre>
 
-### `item_display_name(item, hasbracket?)`
- returns the `display_name` of the item as a Text Value. if hasbracket is true, there will be a bracket around it.
+### `item_name_text(item )`
+ returns the name of the item as a Text Value. 
 
 ### `display_title(players, type, text?, fadeInTicks?, stayTicks?, fadeOutTicks),`
 
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -269,7 +269,9 @@ Example usages:
 </pre>
 
 ### `item_display_name(item )`
- returns the name of the item as a Text Value. `item` should be a list of `[item_name, count, nbt]`, or just an item name.
+ Returns the name of the item as a Text Value. `item` should be a list of `[item_name, count, nbt]`, or just an item name.
+ Please note that it is a translated value, and the result of it converting to a string may be affacted by your language.
+ If the item is renamed, it will also be reflected in the results.
 
 
 ### `display_title(players, type, text?, fadeInTicks?, stayTicks?, fadeOutTicks),`
--- a/docs/scarpet/api/BlocksAndWorldAccess.md
+++ b/docs/scarpet/api/BlocksAndWorldAccess.md
@@ -421,7 +421,7 @@ Numeric function, returning the sky light at position (from sky access).
 
 ### `effective_light (pos)`
 
-Numeric function, returning the "real" light at position. affected by time and weather.
+Numeric function, returning the "real" light at position, which is affected by time and weather.
 
 ### `see_sky(pos)`
 
--- a/docs/scarpet/api/BlocksAndWorldAccess.md
+++ b/docs/scarpet/api/BlocksAndWorldAccess.md
@@ -419,9 +419,9 @@ Numeric function, returning the block light at position (from torches and other 
 
 Numeric function, returning the sky light at position (from sky access).
 
-### `real_light(pos)`
+### `effective_light (pos)`
 
-Numeric function, returning the real light at position.
+Numeric function, returning the "real" light at position. affected by time and weather.
 
 ### `see_sky(pos)`
 
--- a/docs/scarpet/api/BlocksAndWorldAccess.md
+++ b/docs/scarpet/api/BlocksAndWorldAccess.md
@@ -421,7 +421,7 @@ Numeric function, returning the sky light at position (from sky access).
 
 ### `effective_light (pos)`
 
-Numeric function, returning the "real" light at position, which is affected by time and weather.
+Numeric function, returning the "real" light at position, which is affected by time and weather. which also affects mobs spawning, frosted ice blocks melting.
 
 ### `see_sky(pos)`
 
--- a/docs/scarpet/api/BlocksAndWorldAccess.md
+++ b/docs/scarpet/api/BlocksAndWorldAccess.md
@@ -419,6 +419,10 @@ Numeric function, returning the block light at position (from torches and other 
 
 Numeric function, returning the sky light at position (from sky access).
 
+### `real_light(pos)`
+
+Numeric function, returning the real light at position.
+
 ### `see_sky(pos)`
 
 Boolean function, returning true if the block can see sky.
--- a/docs/scarpet/api/BlocksAndWorldAccess.md
+++ b/docs/scarpet/api/BlocksAndWorldAccess.md
@@ -419,7 +419,7 @@ Numeric function, returning the block light at position (from torches and other 
 
 Numeric function, returning the sky light at position (from sky access).
 
-### `effective_light (pos)`
+### `effective_light(pos)`
 
 Numeric function, returning the "real" light at position, which is affected by time and weather. which also affects mobs spawning, frosted ice blocks melting.
 
--- a/docs/scarpet/api/Entities.md
+++ b/docs/scarpet/api/Entities.md
@@ -308,7 +308,7 @@ The item triple (name, count, nbt) if its an item or item frame entity, `null` o
 
 ### `query(e, 'offering_flower')`
 
-The iron golem have red flower in hand or not.
+Whether the given iron golem has a red flower in their hand.
 
 ### `query(e, 'blue_skull')`
 
--- a/docs/scarpet/api/Entities.md
+++ b/docs/scarpet/api/Entities.md
@@ -306,11 +306,11 @@ Number of ticks an entity sits in a portal.
 
 The item triple (name, count, nbt) if its an item or item frame entity, `null` otherwise.
 
-### `query(e, 'offer_flower')`
+### `query(e, 'offering_flower')`
 
 The iron golem have red flower in hand or not.
 
-### `query(e, 'skull_dangerous')`
+### `query(e, 'blue_skull')`
 
 The skull is blue or not.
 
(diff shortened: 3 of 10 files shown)


ch-yx:
about the cursor stack:
although i write it, i dont think that that is the best way.

could respectable @replaceitem write a getter to get players current vanilla screen? 
get_current_screen(player) to warp the vanilla screen to a screen handler and retuen it.
then we can use screen values method to get cursor stack.
this is better than entity property.

replaceitem:
I never thought about doing that, not sure if thats easily possible, but if it is, i could add it.
However i think it would make sense to use `player~'current_screen'` instead?

BisUmTo:
Missing documentation 😉

replaceitem (commented):
The `get_current_screen` (improvement over `player~'current_screen'` to allow for attaching a callback) should be a seperate PR, since this one only targets some missing entity query/modify properties.

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -1668,12 +1699,69 @@ else if (v instanceof ListValue)
...
+            }
+            
+        });
+        put("item", (e, v) -> {
+                ItemStack item=getItemStackFromValue(v).copy();
+                if(e instanceof ItemEntity iteme)            
+                    iteme.setItem(item);
+                if(e instanceof ItemFrame iteme)
+                    iteme.setItem(item);  
+        });
```
Ghoulboy78, line 1731: I can't comment below cos its not modified, but you can remove the checkbox for item (line 1733)

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -1668,12 +1699,69 @@ else if (v instanceof ListValue)
...
         // "dimension"      []
         // "item"           []
```
Ghoulboy78, line 1733:
```suggestion
```

Ghoulboy78: Cos uve added it.

src/main/java/carpet/script/api/Auxiliary.java
```diff
@@ -276,6 +276,14 @@ public static void apply(Expression expression)
             ));
         });
 
+        expression.addContextFunction("item_display_name", -1, (c, t, lv)->{
```
Ghoulboy78, line 279: Why not jut merge this into display_name? Like is there a specific reason?

Ghoulboy78:
I think you can add in the docs now

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -1668,12 +1699,69 @@ else if (v instanceof ListValue)
             }
         });
 
+        //setter of ironman flower, blueness of skull, carryingitem
```
Ghoulboy78, line 1702: Same here, you can remove this

src/main/java/carpet/script/api/Auxiliary.java
ch-yx: e has to be an entity value there.

docs/scarpet/api/BlocksAndWorldAccess.md
```diff
@@ -419,6 +419,10 @@ Numeric function, returning the block light at position (from torches and other
 
 Numeric function, returning the sky light at position (from sky access).
 
+### `effective_light (pos)`
+
+Numeric function, returning the "real" light at position. affected by time and weather.
```
Ghoulboy78, line 424:
```suggestion
Numeric function, returning the "real" light at position, which is affected by time and weather.
```
This may be clearer to understand

docs/scarpet/api/Entities.md
```diff
@@ -670,6 +678,18 @@ if the custom name is always visible, even through blocks.
 Sets the entity persistence tag to `true` (default) or `false`. Only affects mobs. Persistent mobs
 don't despawn and don't count towards the mobcap.
 
+### `modify(e, 'item', item_triple)`
+
+Sets the item for the item or item frame entity.
```
Ghoulboy78, line 683: You could mention something like (The item triple is a list of `[item_name, count, nbt]`

docs/scarpet/api/Entities.md
```diff
@@ -304,7 +304,15 @@ Number of ticks an entity sits in a portal.
...
-The item triple (name, count, nbt) if its an item entity, `null` otherwise.
+The item triple (name, count, nbt) if its an item or item frame entity, `null` otherwise.
+
+### `query(e, 'offering_flower')`
+
+The iron golem have red flower in hand or not.
+
+### `query(e, 'blue_skull')`
+
+The skull is blue or not.
```
Ghoulboy78, line 315: I think it's worth mentioning that it's a wither skull

docs/scarpet/api/Entities.md
```diff
@@ -670,6 +678,18 @@ if the custom name is always visible, even through blocks.
...
+
+Sets the item for the item or item frame entity.
+
+### `modify(e, 'offering_flower', bool)`
+
+Sets if the iron golem have red flower in hand.
+
+### `modify(e, 'blue_skull', bool)`
+
+Sets the skull is blue or not.
```
Ghoulboy78, line 691: Same here for the wither skull

docs/scarpet/api/Auxiliary.md
```diff
@@ -268,6 +268,9 @@ Example usages:
   // not a problem in apps
 </pre>
 
+### `item_display_name(item, hasbracket?)`
+ returns the `display_name` of the item as a Text Value. if hasbracket is true, there will be a bracket around it.
+
```
Ghoulboy78, line 273: IMO you should mention that `item` can be either an item triplet of `[item_name, count, nbt]`, or an item name.

docs/scarpet/api/BlocksAndWorldAccess.md
```diff
@@ -419,6 +419,10 @@ Numeric function, returning the block light at position (from torches and other
 
 Numeric function, returning the sky light at position (from sky access).
 
+### `effective_light (pos)`
+
+Numeric function, returning the "real" light at position, which is affected by time and weather.
```
altrisi, line 424: What is this light level used for (by the game)?

I think that should be specified, for example something like used for spawning or whatever it's used for.

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -1668,12 +1690,57 @@ else if (v instanceof ListValue)
...
+                if(e instanceof ItemFrame iteme)
+                    iteme.setItem(item);  
+        });
         // "dimension"      []
-        // "item"           []
         // "count",         []
         // "effect_"name    []
     }};
 
+    static public ItemStack getItemStackFromValue(Value value){
```
altrisi, line 1717: Question (I don't know), isn't there something to make this already from a triple? I'd assume there'd be other functions doing similar stuff.

src/main/java/carpet/script/api/Auxiliary.java
```diff
@@ -276,6 +276,14 @@ public static void apply(Expression expression)
             ));
         });
 
+        expression.addContextFunction("item_name_text", -1, (c, t, lv)->{
+            /*if(lv.size()==1 || lv.get(1).getBoolean()){
+                return new FormattedTextValue(EntityValue.getItemStackFromValue(lv.get(0)).getDisplayName());
+            }*/
```
altrisi, line 282: dead code

src/main/java/carpet/script/api/Auxiliary.java
```diff
@@ -276,6 +276,14 @@ public static void apply(Expression expression)
             ));
         });
 
+        expression.addContextFunction("item_name_text", -1, (c, t, lv)->{
```
altrisi, line 279: `item_name_text` sounds strange, I'd suggest something more like `item_display_name` (or maybe `item_name` if there's nothing like that already, though not sure), because it being a text isn't something worth noting in the method name IMO (given it's interchangeable with a string).

altrisi: Actually wasn't this already `item_display_name` before?

src/main/java/carpet/script/value/EntityValue.java
altrisi: Note that I haven't really reviewed this method yet.

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -1668,12 +1690,57 @@ else if (v instanceof ListValue)
...
+                nbt=nbtv.getCompoundTag();
+            }
+            else{
+                nbt=new NBTSerializableValue(items.get(2).getString()).getCompoundTag();
+            }
+        }
+        try{
+            return NBTSerializableValue.parseItem(id,nbt).createItemStack(count,false);
+        } catch (CommandSyntaxException e1) {
+            e1.printStackTrace();
```
altrisi, line 1740: Don't do `printStackTrace`. Instead, if the exception is expected to happen for some reason, handle it. Else, pass it through `CarpetSettings.LOG` with a message saying what was going on and possibly throw an exception unless it's not that bad.

Btw why is a `CommandSyntaxException` possible here? Seems like you may be calling something to do with commands? Though idk, as mentioned I haven't really checked this method yet.

altrisi (commented):
Pending any kind of testing.

Also code style is quite different from what's around, though I don't think that matters _that_ much.

docs/scarpet/api/BlocksAndWorldAccess.md
ch-yx: it is used in monster.Monster.isDarkEnoughToSpawn. i think that is why monster only spawn in night

src/main/java/carpet/script/value/EntityValue.java
Ghoulboy78: imo this method fits in better in ValueConversions

src/main/java/carpet/script/api/Auxiliary.java
```diff
@@ -276,6 +276,11 @@ public static void apply(Expression expression)
             ));
         });
 
+        expression.addContextFunction("item_display_name", -1, (c, t, lv)->{
+            return new FormattedTextValue(EntityValue.getItemStackFromValue(lv.get(0)).getHoverName());
+
+        });
```
altrisi, line 282: I don't fully like this function. It has too much of implementation detail.
Here's my issues (maybe not everyone sees all of them as issues, so IMO should be discussed with others too possibly on discord):

The function returns a translated text. This is an issue first of all because it's not documented in the docs (and it should maybe even be in the name? idk), but also because that implies other details and gets more error prone.

While on singleplayer it's simple: it'll translate to the client languages, it will be unexpected that on servers it will be in english (server only loads the english lang file). This is error prone, because someone may just test it in singleplayer and assume it'll always be translated, or only in multi and assume it'll never be.

Because of that, it also limits what can be done with it. Say you want to get the color of a wool block (or something, assume you can't use the id for whatever reason): You can't reliably remove the "bed" suffix from it given it'll not have the same length in all languages, and it may not even be a suffix in some languages!

At the same I also see how the translated name could be useful though, so idk, that's why I think it should be discussed. Though maybe (all of this tbd) it could be left for a more generic function to get a translated thing, since (I think?) the translated name can be gotten easily from id.

altrisi: Oh and docs should also mention more about it, like nbt having custom name will return that instead, I think they're too short for a function where details are IMO relevant.

altrisi (commented):
Few small things and one that I think needs discussing further with others (and especially missing details in docs).

altrisi (approved):

docs/scarpet/api/Auxiliary.md
```diff
@@ -268,6 +268,14 @@ Example usages:
   // not a problem in apps
 </pre>
 
+### `item_display_name(item)`
+ Returns the name of the item as a Text Value. `item` should be a list of `[item_name, count, nbt]`, or just an item name.
```
gnembon, line 272: `/script run item_display_name(['iron_nugget', 0, null])` returns 'Air' instead of null or raising UNKNOWN_ITEM, and this piece of documentation says it would handle that. 

I assume the only reason to support item triples is to handle cases where items are renamed in nbt, correct?

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -1668,12 +1690,56 @@ else if (v instanceof ListValue)
...
     }};
 
+    static public ItemStack getItemStackFromValue(Value value){
+        List<Value> items;
+        if(value instanceof ListValue blv && blv.length()==3){
+            items = blv.getItems();
+        }else if(value.isNull()){
+            return ItemStack.EMPTY;
+        }else{
+            items= List.of(StringValue.of(value.getString()),Value.ONE,Value.NULL);
```
gnembon, line 1724: this seems little excessive, if only string value is provided, then you could grab the item from registries and make a single stack out of it. check crafting_remaining_item and then item.getDefaultInstance()

gnembon: ok, for the sake of completeness, I guess it works with 
`item_display_name('iron_nugget{RepairCost:0,display:{Name:\'{"text":"IroNug"}\'}}')`
it does which is nice, but having a shortcut of submitting just registry valid name makes sense to me from the performance perspective

src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -1668,12 +1690,56 @@ else if (v instanceof ListValue)
...
+                nbt=nbtv.getCompoundTag();
+            }
+            else{
+                nbt=new NBTSerializableValue(items.get(2).getString()).getCompoundTag();
+            }
+        }
+        try{
+            return NBTSerializableValue.parseItem(id,nbt).createItemStack(count,false);
+        } catch (CommandSyntaxException e1) {
+            throw new IllegalStateException("Unexpected exception while creating item stack", e1);
```
gnembon, line 1740: yeah, would throw INVALID_ITEM instead of illegal state. illegal state should not be thrown from invalid text input, allowing script writer to validate user input for example

src/main/java/carpet/script/value/EntityValue.java
gnembon: it does fit better in value conversions - not in entity code

gnembon: value conversions in general convert the other way. There is NBTSerializableValue.parseItem that returns one without count, which could be used instead of this method (it does the same thing, just cached)

src/main/java/carpet/script/value/EntityValue.java
gnembon: ok, since NBTSerializableValue.parseItem already does that disregard this comment and use that instead.

docs/scarpet/api/Auxiliary.md
```diff
@@ -268,6 +268,14 @@ Example usages:
   // not a problem in apps
 </pre>
 
+### `item_display_name(item)`
```
gnembon, line 271: new functions need to be added to the supported editor keywords

docs/scarpet/api/BlocksAndWorldAccess.md
```diff
@@ -419,6 +419,10 @@ Numeric function, returning the block light at position (from torches and other
 
 Numeric function, returning the sky light at position (from sky access).
 
+### `effective_light (pos)`
```
gnembon, line 422: space

docs/scarpet/api/Entities.md
```diff
@@ -304,7 +304,15 @@ Number of ticks an entity sits in a portal.
...
-The item triple (name, count, nbt) if its an item entity, `null` otherwise.
+The item triple (name, count, nbt) if its an item or item frame entity, `null` otherwise.
+
+### `query(e, 'offering_flower')`
+
+Whether the given iron golem has a red flower in their hand.
+
+### `query(e, 'blue_skull')`
+
+Whether the given wither skull entity is blue.
```
gnembon, line 315: returns null for all other entities

ch-yx:
i moved getItemStackFromValue to NBTSerializableValue as everyone starts thinking it fit there.

docs/scarpet/api/Auxiliary.md
ch-yx: i m sorry that i dont know how editor file works... 
and npp is a illegal software in my country. i dont know if i allowed to do so.

src/main/java/carpet/script/value/EntityValue.java
ch-yx: this method IS using said NBTSerializableValue.parseItem to do so.
and i move it to ValueConversions as every one thinks it fit there better.

ch-yx:
i moved getItemStackFromValue to ValueConversions as every one thinks it fit there better.

docs/scarpet/api/Auxiliary.md
ch-yx: so i added the idea`s one now.

PR #1365, opened 2022-03-16
adding a shape that can display a item or a block

ch-yx:
**base on the polygon one.**
**images of the polygons are in that pr**
this pr is to add a shape that can display a item or a block
 * `'block_displayer'` \ `'item_displayer'`:
   * Required attributes:
     * `pos` - position of the object.
   * Optional attributes:
     * `needf3b` - if True, can toggle visibility by using F3+B. just like entities' hitbox. (maybe call it easytoggle?)
     * `facing` - object's direction, where its facing. Possible options are: `player` (default, it
     always rotates to face the player), `north`, `south`, `east`, `west`, `up`, `down`
     * `tilt`, `lean`, `turn` - additional rotations along all three axis
     * `item` - the object to show. could have NBT. is a item tuple or a string-id (exclusive to `'item_displayer'`)
     * `block` - the object to show. could have NBT. is a block value or a string-id (exclusive to `'block_displayer'`)
     * `height`, `width`, `obj_size` - size of it. should be numbers
     * `light_fromsky`, `light_fromblock` - light level. omit it to use local light level. should between 0~15
(After testing, it is sad to find that one light level is difficult to simulate the effect in the game. Had to split into two)(maybe its better to call them sky_light/block_light)

here is a preview. dont know why but firefox cannot see it. edge can.

https://user-images.githubusercontent.com/45825511/165358852-3b4f564e-d13f-41ff-b672-8ef90298966b.mp4

upload the same video again. firefox can see it, but edge cant....

https://user-images.githubusercontent.com/45825511/181075383-5d63935c-1898-4430-bab4-bf4730ed2aef.mp4


docs/scarpet/api/Auxiliary.md
```diff
@@ -142,6 +142,16 @@ Available shapes:
      * `height` - height of the cyllinder, defaults to `0`, so flat disk. Can be negative.
      * `level` - level of details, see `'sphere'`.
 
+ * `'poly'`:
+   * Required attributes:
+     * `points` - vertices of the polygon
```
Ghoulboy78, line 147: I recommend changing to `list of points defining vertices of the polygon`

docs/scarpet/api/Auxiliary.md
```diff
@@ -142,6 +142,16 @@ Available shapes:
...
+   * Optional attributes:
+     * `relative` - list of bools. vertices of the polygont that affected by 'follow'. Default means that every point is affacted.
```
Ghoulboy78, line 149: polygon, not polygont

docs/scarpet/api/Auxiliary.md
```diff
@@ -142,6 +142,16 @@ Available shapes:
...
+     * `mode` - how those points are connected. may be "polygon"(default),"strip" or "triangles". "polygon" means that it will be viewed as vertices of a polygon center on the first one. "strip" means that it will be viewed as a triangles strip. "triangles" means that it will be viewed as some triangles that are not related to each other.
```
Ghoulboy78, line 150: You should specify that for `triangles` mode, you need to input a triplet of vertices for each triangle

src/main/java/carpet/script/utils/ShapeDispatcher.java
```diff
@@ -121,7 +133,7 @@ public static void sendShape(Collection<ServerPlayer> players, List<ShapeWithCon
         List<ServerPlayer> alternativePlayers = new ArrayList<>();
         for (ServerPlayer player : players)
         {
-            if (ServerNetworkHandler.isValidCarpetPlayer(player))
+            if (ServerNetworkHandler.isValidCarpetPlayer(player)&&!CarpetSettings.superSecretSetting)//for debug
```
Ghoulboy78, line 136: Correct me if I'm wrong, but I thought that we filter out carpet client players because non-carpet clients can't receive the shapes correctly?

src/main/java/carpet/script/utils/ShapeDispatcher.java
```diff
@@ -249,6 +261,8 @@ public static ExpiringShape fromTag(CompoundTag tag)
             put("sphere", creator(Sphere::new));
             put("cylinder", creator(Cylinder::new));
             put("label", creator(DisplayedText::new));
+            put("poly",creator(Polyface::new));
+            put("blockoritemdisplayer",creator(DisplayedItem::new));
```
Ghoulboy78, line 265: Have you added this to the docs?

Ghoulboy78 (approved):
This looks really amazing, can't wait to play about with this in Scarpet Edit. My mind has immediately jumped to a marching cubes implementation in scarpet (which was possible with previous draw_shapes stuff, but more cumbersome).
Also I wonder if we could draw shapes by using a combination of this and shapes.scl

src/main/java/carpet/script/utils/ShapeDispatcher.java
ch-yx: not yet. i only write the polygon`s doc. 
block or item displayer`s one is not written yet.

ch-yx: @Ghoulboy78 one of the reasons is i dont know if those names of arguments are nice for English speakers.

forexample, i dont think "blockoritemdisplayer" is a good name. but i cannot find a better one.
i want to get some feed back and decide those name first.

Ghoulboy78: true that.

ch-yx: "fix"ed by split it into`'block_displayer'` and `'item_displayer'`:

Ghoulboy78:
Btw, is there by any chance a way to make the shapes more or less transparent? (Just curious, cos it would be cool for something I might do with them)

ch-yx:
should there be a item's model overwrite (for clock/compass) and/or use item_frames code for filledmap?

docs/scarpet/api/Auxiliary.md
```diff
@@ -142,6 +142,33 @@ Available shapes:
...
 
+ * `'poly'`:
+   * Required attributes:
+     * `points` - list of points defining vertices of the polygon
+   * Optional attributes:
+     * `relative` - list of bools. vertices of the polygon that affected by 'follow'. Default means that every point is affacted.
+     * `mode` - how those points are connected. may be "polygon"(default),"strip" or "triangles". "polygon" means that it will be viewed as vertices of a polygon center on the first one. "strip" means that it will be viewed as a triangles strip. "triangles" means that it will be viewed as some triangles that are not related to each other (therefor length of `points` in this mode have to be a multiple of 3).
+     * `inneredges` - if `true` it will make the inner edges be drawn as well. 
+     * `doublesided` - if `true` it will make the shapes visible from the back as well. Default is `true`. 
+     * `needf3b` - if True, can toggle visibility by using F3+B. just like entities' hitbox.
```
gnembon, line 153: this option sounds very useful, not just for these shapes but for any shape as a general option. But I would definitely change its name. If we want a shape to appear only when hitboxes option is toggled, then I would maybe call it 'toggleable', or 'debug' indicating it would be used with F3B debug mode. thoughts on name?

docs/scarpet/api/Auxiliary.md
```diff
@@ -142,6 +142,33 @@ Available shapes:
      * `height` - height of the cyllinder, defaults to `0`, so flat disk. Can be negative.
      * `level` - level of details, see `'sphere'`.
 
+ * `'poly'`:
+   * Required attributes:
+     * `points` - list of points defining vertices of the polygon
+   * Optional attributes:
+     * `relative` - list of bools. vertices of the polygon that affected by 'follow'. Default means that every point is affacted.
+     * `mode` - how those points are connected. may be "polygon"(default),"strip" or "triangles". "polygon" means that it will be viewed as vertices of a polygon center on the first one. "strip" means that it will be viewed as a triangles strip. "triangles" means that it will be viewed as some triangles that are not related to each other (therefor length of `points` in this mode have to be a multiple of 3).
+     * `inneredges` - if `true` it will make the inner edges be drawn as well.
```
gnembon, line 151: inner? (options should be as short and clear skipping unimportant details, like 'edges')

docs/scarpet/api/Auxiliary.md
```diff
@@ -142,6 +142,33 @@ Available shapes:
      * `height` - height of the cyllinder, defaults to `0`, so flat disk. Can be negative.
      * `level` - level of details, see `'sphere'`.
 
+ * `'poly'`:
+   * Required attributes:
+     * `points` - list of points defining vertices of the polygon
+   * Optional attributes:
+     * `relative` - list of bools. vertices of the polygon that affected by 'follow'. Default means that every point is affacted.
```
gnembon, line 149: I would guess, could be just a single bool to affect all points, right?

docs/scarpet/api/Auxiliary.md
```diff
@@ -142,6 +142,33 @@ Available shapes:
...
+     * `mode` - how those points are connected. may be "polygon"(default),"strip" or "triangles". "polygon" means that it will be viewed as vertices of a polygon center on the first one. "strip" means that it will be viewed as a triangles strip. "triangles" means that it will be viewed as some triangles that are not related to each other (therefor length of `points` in this mode have to be a multiple of 3).
+     * `inneredges` - if `true` it will make the inner edges be drawn as well. 
+     * `doublesided` - if `true` it will make the shapes visible from the back as well. Default is `true`. 
+     * `needf3b` - if True, can toggle visibility by using F3+B. just like entities' hitbox.
+
+ * `'block_displayer'` \ `'item_displayer'`:
```
gnembon, line 155: 'block' and 'item'

docs/scarpet/api/Auxiliary.md
```diff
@@ -142,6 +142,33 @@ Available shapes:
...
+   * Required attributes:
+     * `pos` - position of the object.
+     * `item` - the object to show. could have NBT. is a item tuple or a string-id (exclusive to `'item_displayer'`)
```
gnembon, line 158: option still could be shared between 'block' and 'item'. call it 'id' or 'value', or 'name' or something else. 'id' feels best because is also very short and saves bandwidth, but you may find other arguments for or agains.

docs/scarpet/api/Auxiliary.md
```diff
@@ -142,6 +142,33 @@ Available shapes:
...
+     * `block` - the object to show. could have NBT. is a block value or a string-id (exclusive to `'block_displayer'`)
+   * Optional attributes:
+     * `needf3b` - if True, can toggle visibility by using F3+B. just like entities' hitbox.
+     * `facing` - object's direction, where its facing. Possible options are: `player` (default, it
+     always rotates to face the player), `north`, `south`, `east`, `west`, `up`, `down`
+     * `tilt`, `lean`, `turn` - additional rotations along all three axis
+     * `height`, `width`, `obj_size` - size of it. should be numbers
```
gnembon, line 165: 'size' - for all three dimensions, or accepting a triple of [x, y, z] size. This can be one parameter

docs/scarpet/api/Auxiliary.md
```diff
@@ -142,6 +142,33 @@ Available shapes:
...
+     * `light_fromsky`, `light_fromblock` - light level. omit it to use local light level. should between 0~15.
```
gnembon, line 166: 'skylight' and 'blocklight', or 'skylight' and 'light' identifying blocklight as a default light type.

docs/scarpet/api/Auxiliary.md
```diff
@@ -142,6 +142,33 @@ Available shapes:
...
+     * `item_transform_type` - one of "NONE","THIRD_PERSON_LEFT_HAND","THIRD_PERSON_RIGHT_HAND","FIRST_PERSON_LEFT_HAND",
```
gnembon, line 167: 'version' ? 'variant' ? 'transformation'?

docs/scarpet/api/Auxiliary.md
```diff
@@ -142,6 +142,33 @@ Available shapes:
...
+     "FIRST_PERSON_RIGHT_HAND","HEAD","GUI","GROUND","FIXED". In addition to the literal meaning, it can also be used to use special models of tridents and telescopes. (exclusive to `'item_displayer'`)
```
gnembon, line 168: please use lower case snake case constants like the rest of the scarpet API, so 'first_person_right_hand'

docs/scarpet/api/Auxiliary.md
```diff
@@ -142,6 +142,33 @@ Available shapes:
      * `height` - height of the cyllinder, defaults to `0`, so flat disk. Can be negative.
      * `level` - level of details, see `'sphere'`.
 
+ * `'poly'`:
```
gnembon, line 145: 'polygon'

src/main/java/carpet/mixins/ShulkerBoxBlockEntity_renderFixMixin.java
```diff
@@ -0,0 +1,30 @@
+package carpet.mixins;
```
gnembon, line 1: why do you need that? Can you explain what's wrong with it?

src/main/java/carpet/script/utils/ShapeDispatcher.java
gnembon: isValidCarpetPlayer is already checking for super secret settings - that is not needed

src/main/java/carpet/script/utils/ShapeDispatcher.java
```diff
@@ -546,6 +563,129 @@ public long calcKey()
...
+        CompoundTag blockEntity;
+        BlockState blockState;
+        ItemStack item = null;
+        TransformType item_transform_type;
+
+        @Override
+        protected void init(Map<String, Value> options) {
+            super.init(options);
+            pos = vecFromValue(options.get("pos"));
+            blockState = ((BlockValue) options.getOrDefault("block", BlockValue.AIR)).getBlockState();
```
gnembon, line 618: getting the block should happen once and reused.

gnembon: and as suggested make it one argument and make it mandatory, so not expecting AIR as default

src/main/java/carpet/script/utils/ShapeDispatcher.java
```diff
@@ -546,6 +563,129 @@ public long calcKey()
...
+            blockEntity = ((BlockValue) options.getOrDefault("block", BlockValue.AIR)).getData();
+            NBTSerializableValue item_ = (NBTSerializableValue) options.getOrDefault("item", null);
```
gnembon, line 620: what is that _ in the name?

src/main/java/carpet/script/utils/ShapeDispatcher.java
```diff
@@ -546,6 +563,129 @@ public long calcKey()
...
+            light_fromblock = NumericValue
+                    .asNumber(options.getOrDefault("light_fromblock", optional.get("light_fromblock"))).getInt();
+            if (light_fromblock > 15)
+                light_fromblock = 15;
+            light_fromsky = NumericValue.asNumber(options.getOrDefault("light_fromsky", optional.get("light_fromsky")))
+                    .getInt();
+            if (light_fromsky > 15)
+                light_fromsky = 15;
+
+            item_transform_type = TransformType.valueOf(options.getOrDefault("item_transform_type",new StringValue("GUI")).getString());
```
gnembon, line 633: this is typically not specified, so make a shortcut rather than creating a custom objects.

src/main/java/carpet/script/utils/ShapeDispatcher.java
```diff
@@ -546,6 +563,129 @@ public long calcKey()
...
+            lean = NumericValue.asNumber(options.getOrDefault("lean", optional.get("lean"))).getFloat();
+            turn = NumericValue.asNumber(options.getOrDefault("turn", optional.get("turn"))).getFloat();
+            height = NumericValue.asNumber(options.getOrDefault("height", optional.get("height"))).getFloat();
+            width = NumericValue.asNumber(options.getOrDefault("width", optional.get("width"))).getFloat();
+            size = NumericValue.asNumber(options.getOrDefault("obj_size", optional.get("obj_size"))).getFloat();
+        }
+
+        @Override
+        public Consumer<ServerPlayer> alternative() {
+            return s -> {};
```
gnembon, line 662: for blocks, technically you can fetch its map color and use that for a particle in the box its rendered on, or display just the block particle, which is vanilla.

gnembon: same for items that are block items

docs/scarpet/api/Auxiliary.md
ch-yx: ok. i will call it 'toggleable' instead.

docs/scarpet/api/Auxiliary.md
ch-yx: yes it is inner edges indeed.

said if i draw a square with inner edge, it will be like 〼..... the polygon is made by triangles. if inner edge is on, edges of triangles will be drawn as well. that is usefull when those triangles are not in the same plane.

docs/scarpet/api/Auxiliary.md
ch-yx: the name "size" is already be used in other shape. and it is a int argument.
that is why i have to rename it to obj_size.

docs/scarpet/api/Auxiliary.md
ch-yx: lower case is allowed too in my code. maybe i should change them into lower case in document.

src/main/java/carpet/mixins/ShulkerBoxBlockEntity_renderFixMixin.java
ch-yx: here is some shulkers block s render code in  net.minecraft.client.renderer.blockentity.ShulkerBoxRenderer
it get its block state from its level.
```
	public void render(ShulkerBoxBlockEntity shulkerBoxBlockEntity, float f, PoseStack poseStack, MultiBufferSource multiBufferSource, int i, int j) {
		Direction direction = Direction.UP;
		if (shulkerBoxBlockEntity.hasLevel()) {
			BlockState blockState = shulkerBoxBlockEntity.getLevel().getBlockState(shulkerBoxBlockEntity.getBlockPos());
			if (blockState.getBlock() instanceof ShulkerBoxBlock) {
				direction = blockState.getValue(ShulkerBoxBlock.FACING);
			}
		}
```

but other block entities get it more directly:
here is beds
```
	public void render(BedBlockEntity bedBlockEntity, float f, PoseStack poseStack, MultiBufferSource multiBufferSource, int i, int j) {
		Material material = Sheets.BED_TEXTURES[bedBlockEntity.getColor().getId()];
		Level level = bedBlockEntity.getLevel();
		if (level != null) {
			BlockState blockState = bedBlockEntity.getBlockState();
```
so i have to use a mixin to fix that. as there is not a real shulker box in the world.

ch-yx: i read others block entities code. only shulker box write it that way.

docs/scarpet/api/Auxiliary.md
ch-yx: ah.... i ve already wroten block and item argument..... which support id state and nbt...

docs/scarpet/api/Auxiliary.md
ch-yx: done.

docs/scarpet/api/Auxiliary.md
ch-yx: no. but default means that every point is affacted. if 'follow' is omited, then every point is not affacted. so i dont see why it need to support a single bool.

but if you really want, i could add it.

docs/scarpet/api/Auxiliary.md
ch-yx: done

src/main/java/carpet/script/utils/ShapeDispatcher.java
ch-yx: > for blocks, technically you can fetch its map color and use that for a particle in the box its rendered on, or display just the block particle, which is vanilla.

i get its map color to implent its alternative.

src/main/java/carpet/script/utils/ShapeDispatcher.java
ch-yx: ok. i fix it.

src/main/java/carpet/script/utils/ShapeDispatcher.java
ch-yx: fixed

docs/scarpet/api/Auxiliary.md
ch-yx: now the name is bound with its data type. cannot have different type in the same name.

src/main/java/carpet/script/utils/ShapeDispatcher.java
ch-yx: i did not get what you mean.....
what object is the custom objects you said?

docs/scarpet/api/Auxiliary.md
ch-yx: i dont know which is better. I feel that these words are not very precise.  maybe 'item_display_context'?

docs/scarpet/api/Auxiliary.md
ch-yx: and if that is off. only outer edge are drawn. that is usefull when those triangles are in the same plane to creat a polygon.

docs/scarpet/api/Auxiliary.md
ch-yx: and another difficulty is block could accept a block value, but item cannot. if they have the same name, then it have to share the same value type.....

src/main/java/carpet/mixins/ShulkerBoxBlockEntity_renderFixMixin.java
ch-yx: (if only you can change shulker box`s code, then this file is no longer needed.)

src/main/java/carpet/script/utils/ShapeDispatcher.java
ch-yx: i fixed it in my understanding

docs/scarpet/api/Auxiliary.md
ch-yx: done

src/main/java/carpet/mixins/ShulkerBoxBlockEntity_renderFixMixin.java
gnembon: problem is you are changing the vanilla code for some use of some mod. If you change this, does it mess up rendering of shulker boxes in some other places? what about in the inventory? All in all, carpet should strive to preserve vanilla implementation at all cost, even if it ends up in ugly code, or causing things to not be perfect.

ch-yx: i tried. and i did not find any problem in rendering at all.

and i think is the one who write that code forget that he could get it directly. XD

src/main/java/carpet/script/utils/ShapeDispatcher.java
```diff
@@ -546,6 +564,150 @@ public long calcKey()
...
+            lean = NumericValue.asNumber(options.getOrDefault("lean", optional.get("lean"))).getFloat();
+            turn = NumericValue.asNumber(options.getOrDefault("turn", optional.get("turn"))).getFloat();
+            List<Value> scale = ((ListValue) options.getOrDefault("scale", optional.get("scale"))).unpack();
+            scale_y = NumericValue.asNumber(scale.get(1)).getFloat();
+            scale_x = NumericValue.asNumber(scale.get(0)).getFloat();
+            scale_z = NumericValue.asNumber(scale.get(2)).getFloat();
+        }
+
+        @Override
+        public Consumer<ServerPlayer> alternative() {
```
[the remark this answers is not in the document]
ch-yx: It is my fault. I have switched to block_marker.

src/main/java/carpet/mixins/ShulkerBoxBlockEntity_renderFixMixin.java
ch-yx: fixed this in another way.
this time no vanilla code changed.

docs/scarpet/api/Auxiliary.md
ch-yx: > I would guess, could be just a single bool to affect all points, right?

ok. i let it could be a single bool to affect all points now.

docs/scarpet/api/Auxiliary.md
ch-yx: > inner? (options should be as short and clear skipping unimportant details, like 'edges')

i renamed it to inner.

src/main/java/carpet/script/utils/ShapeDispatcher.java
ch-yx: fixed

src/main/java/carpet/script/utils/ShapeDispatcher.java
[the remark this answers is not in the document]
ch-yx: i thought 'block' when i write it the first time. which is bad.

src/main/java/carpet/script/utils/ShapesRenderer.java
```diff
@@ -233,6 +287,181 @@ public void promoteWith(RenderedShape<?> rshape)
...
+            RenderSystem.depthMask(true);
+            RenderSystem.enableCull();
+            RenderSystem.enableDepthTest();
+
+            // matrices.scale(-1, 1, 1);
+
+            blockPos = new BlockPos(v1);
+            int light;
+            light = LightTexture.pack(
+                    shape.light_fromblock == -999 ? client.level.getBrightness(LightLayer.BLOCK, blockPos)
```
gnembon, line 370: detail, nit, but I woudn't use -999 for a specific use, any value outside of 0-15 would need a default. This way the default could be just -1, right?

src/main/java/carpet/script/utils/ShapesRenderer.java
```diff
@@ -386,7 +615,165 @@ public void renderLines(PoseStack matrices, Tesselator tessellator, BufferBuilde
...
+    {
+        public RenderedPolyface(Minecraft client, ShapeDispatcher.ExpiringShape shape)
+        {
+            super(client, (ShapeDispatcher.Polyface)shape);
+        }
+        @Override
+        public void renderFaces(Tesselator tessellator, BufferBuilder bufferBuilder, double cx, double cy, double cz, float partialTick)
+        {       
+            if(shape.fa==0){return;}
+            if (shape.needf3b && !client.getEntityRenderDispatcher().shouldRenderHitBoxes()){
```
gnembon, line 628: I higly suggest you move the F3b option up to the any shape category and add that as an extra condition to render any shape.

src/main/java/carpet/script/utils/ShapesRenderer.java
```diff
@@ -386,7 +615,165 @@ public void renderLines(PoseStack matrices, Tesselator tessellator, BufferBuilde
...
+                    Vec3 vec=shape.vertex_list.get(i);
+                    if(shape.relative.get(i)){
+                        vec=shape.relativiseRender(client.level, vec, partialTick);
+                    }
+                    bufferBuilder.vertex(vec.x()-cx, vec.y()-cy, vec.z()-cz).color(shape.fr, shape.fg, shape.fb, shape.fa).endVertex();
+                }
+                tessellator.end();
 
+            /*
+                RenderSystem.depthFunc(515);
```
gnembon, line 653: there is still a lot of commented code here and there indicating you are still testing something. Please cleanup :)

ch-yx: ok XD

src/main/java/carpet/script/utils/ShapesRenderer.java
ch-yx: 🆗 i moved it to its father class

src/main/java/carpet/script/utils/ShapesRenderer.java
ch-yx: i change it to -1

docs/scarpet/api/Auxiliary.md
```diff
@@ -142,6 +143,31 @@ Available shapes:
...
+     * `inner` - if `true` it will make the inner edges be drawn as well. 
+     * `doublesided` - if `true` it will make the shapes visible from the back as well. Default is `true`. 
+
+ * `'block'` \ `'item'`:
+   * Required attributes:
+     * `pos` - position of the object.
+     * `item` - the object to show. could have NBT. is a item tuple or a string-id (exclusive to `'item'`)
+     * `block` - the object to show. could have NBT. is a block value or a string-id (exclusive to `'block'`)
+   * Optional attributes:
+     * `facing` - object's direction, where its facing. Possible options are: `player` (default, it
```
gnembon, line 161: the default should be 'north' for blocks, and 'player' facing makes the blocks rotate when looking at blocks off axis. It does not rotate in 3rd person mode indicating something its flipped with the angles with 'player' facing setting.

gnembon: with items, we could leave default 'north' for consistency.

gnembon:
there is different shading that includes different ambience occlusion happening to blocks. Solving this is not super crucial, but could be good to figure out whats up

docs/scarpet/api/Auxiliary.md
```diff
@@ -142,6 +143,31 @@ Available shapes:
...
+     always rotates to face the player), `north`, `south`, `east`, `west`, `up`, `down`
+     * `tilt`, `lean`, `turn` - additional rotations along all three axis
+     * `scale` - scale of it in 3 axis-direction. should be a number or a list of 3 numbers (x,y,z).
+     * `skylight`, `blocklight` - light level. omit it to use local light level. should between 0~15.
+     * `variant` - one of "none","third_person_left_hand","third_person_right_hand","first_person_left_hand",
+     "first_person_right_hand","head","gui","ground","fixed". In addition to the literal meaning, it can also be used to use special models of tridents and telescopes. (exclusive to `'item'`)
```
gnembon, line 167: shoudn't 'none' be the default? it isn't

src/main/java/carpet/script/utils/ShapeDispatcher.java
```diff
@@ -858,11 +1204,55 @@ public long calcKey()
     public static abstract class Param
     {
         public static Map<String, Param> of = new HashMap<String, Param>(){{
+            /*put("vertex", new PointsParam("vertex"){
```
gnembon, line 1207: please clean up unused code

src/main/java/carpet/script/utils/ShapesRenderer.java
```diff
@@ -233,6 +288,179 @@ public void promoteWith(RenderedShape<?> rshape)
...
+            // shape.size*0.0025f);
+            if (shape.tilt != 0.0f)
+                matrices.mulPose(Vector3f.ZP.rotationDegrees(shape.tilt));
+            if (shape.lean != 0.0f)
+                matrices.mulPose(Vector3f.XP.rotationDegrees(shape.lean));
+            if (shape.turn != 0.0f)
+                matrices.mulPose(Vector3f.YP.rotationDegrees(shape.turn));
+            matrices.scale(shape.scale_x, shape.scale_y, shape.scale_z);
+
+            if (!isitem)// blocks should use its center as the origin
```
gnembon, line 357: you could have mentioned that in the docs - seems important

src/main/java/carpet/script/utils/ShapesRenderer.java
```diff
@@ -233,6 +288,179 @@ public void promoteWith(RenderedShape<?> rshape)
...
+                matrices.translate(-0.5, -0.5, -0.5);
+            // matrices.scale(-1, 1, 1);
+            RenderSystem.depthMask(true);
+            RenderSystem.enableCull();
+            RenderSystem.enableDepthTest();
+
+            // matrices.scale(-1, 1, 1);
```
gnembon, line 364: please remove all that unused code, or if there is a reason to keep it - let us know why you left it.

src/main/java/carpet/script/utils/ShapesRenderer.java
```diff
@@ -233,6 +288,179 @@ public void promoteWith(RenderedShape<?> rshape)
...
+
+            blockPos = new BlockPos(v1);
+            int light;
+            light = LightTexture.pack(
+                    shape.light_fromblock == -1 ? client.level.getBrightness(LightLayer.BLOCK, blockPos)
```
gnembon, line 369: check < 0 rather - that will cover some wierd values, like -2

src/main/java/carpet/script/utils/ShapesRenderer.java
```diff
@@ -233,6 +288,179 @@ public void promoteWith(RenderedShape<?> rshape)
...
+            RenderSystem.depthMask(false);
+
+        }
+
+        @Override
+        public boolean stageDeux() {
+            return true;
+        }
+
+        public void sbrender(ShulkerBoxBlockEntity shulkerBoxBlockEntity, float f, PoseStack poseStack, MultiBufferSource multiBufferSource, int i, int j) {
```
gnembon, line 431: if you copy - paste- modify this method from vanilla code, make sure to include the reference in a comment to allow for simpler tracking in case that stops working or becomes weird

src/main/resources/carpet.accesswidener
```diff
@@ -7,3 +7,4 @@ accessible class net/minecraft/world/level/lighting/BlockLightSectionStorage$Blo
 accessible class net/minecraft/server/MinecraftServer$ReloadableResources
 
 accessible field net/minecraft/world/level/block/state/BlockBehaviour UPDATE_SHAPE_ORDER [Lnet/minecraft/core/Direction;
+accessible field net/minecraft/client/renderer/blockentity/ShulkerBoxRenderer model Lnet/minecraft/client/model/ShulkerModel;
```
gnembon, line 10: you could do that with a mixin - mixins are better / clearer than accessWideners which should only be used if there is no other way.

ch-yx: > you could do that with a mixin - mixins are better / clearer than accessWideners which should only be used if there is no other way.



i was using a mixin....but you said dont touch vanilla code.... so i change to use accessWideners. 
i m new to mixin. if there is a way to use mixin without touch vanilla code, please tell me.

docs/scarpet/api/Auxiliary.md
ch-yx: ok

src/main/java/carpet/script/utils/ShapeDispatcher.java
ch-yx: done.

src/main/java/carpet/script/utils/ShapesRenderer.java
ch-yx: mentioned

src/main/java/carpet/script/utils/ShapesRenderer.java
ch-yx: ok

src/main/java/carpet/script/utils/ShapesRenderer.java
ch-yx: i added a comment to mark its origin

src/main/resources/carpet.accesswidener
ch-yx: I dug into it for a while and it seems to be successfully implemented with a mixin. But I don't know if I am using it correctly. Please review it carefully.

PR #1366, opened 2022-03-17
Ease concrete restrictions on Annotated Scarpet API and make instantiating lazier

altrisi:
Prevents instantiating classes with all annotated methods static, making it also more flexible in that you can put annotated functions near to other API that may be initialized in special ways.

The docs really need rewriting, but that's for another day, this should be trivial.

This is motivated after discovering that Java doesn't throw binary compatibility errors if an annotation type is missing, just ignores those annotations.

Therefore other mods should be able annotate their already API methods directly without fear of binary issues when carpet isn't present, as long as the call to parseFunctionsClass is wrapped inside a condition. This PR improves this by not adding strange behaviour to parsing a class. Further PRs will try to improve this by adding utils like allowing to change the method name from the annotation. Even considered making a "scarpet API lite" jar with just the annotation, Context.Type and a stub for parseFunctionsClass to ease it even further by not requiring the whole carpet package at compile time, but that's way more long term and not sure if really worth it.

### DIFF
--- a/src/main/java/carpet/script/annotation/AnnotationParser.java
+++ b/src/main/java/carpet/script/annotation/AnnotationParser.java
@@ -12,10 +12,13 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


altrisi:
@gnembon Is it ok if I merge PRs on this area after tested against API consumers?

PR #1368, opened 2022-03-21
one-line fix for a minor bug about encoding 

ch-yx:
i think it is necessary to make encoding and decoding use the same codec

it is ok that both of them use local encoding.
it is ok that both of them use the same specific encoding.
it is not ok that only one of them use local encoding.

### DIFF
--- a/src/main/java/carpet/script/language/DataStructures.java
+++ b/src/main/java/carpet/script/language/DataStructures.java
@@ -329,10 +329,10 @@ public class DataStructures {
             return (cc, tt) -> ret;
         });
 
-        expression.addUnaryFunction("encode_b64", v -> StringValue.of(Base64.getEncoder().encodeToString(v.getString().getBytes())));
+        expression.addUnaryFunction("encode_b64", v -> StringValue.of(Base64.getEncoder().encodeToString(v.getString().getBytes(StandardCharsets.UTF_8))));
         expression.addUnaryFunction("decode_b64", v -> {
             try {
-                return StringValue.of(new String(Base64.getDecoder().decode(v.getString()), StandardCharsets.ISO_8859_1));
+                return StringValue.of(new String(Base64.getDecoder().decode(v.getString()), StandardCharsets.UTF_8));
             } catch (IllegalArgumentException iae){
                 throw new ThrowStatement("Invalid b64 string: " + v.getString(), Throwables.B64_ERROR);
             }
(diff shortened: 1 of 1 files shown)


Ghoulboy78:
IMO it's better to use some common encoding for both, like UTF-8, so that people can send messages in base64 and then decode them and have them make sense. I'm also a little embarrassed that I didn't notice this when I wrote this code.

Ghoulboy78:
I would also recommend changing a few lines down in the json section. I have noticed that we use a deprecated method to parse Json, so I would recommend to replace line 344 which currently is:
`return new ScarpetJsonDeserializer().deserialize(new JsonParser().parse(v.getString()), null, null);`
With:
`return new ScarpetJsonDeserializer().deserialize(JsonParser.parseString(v.getString()), null, null);`

src/main/java/carpet/script/language/DataStructures.java
```diff
@@ -341,7 +341,7 @@ else if (lv.size() == 2)
         expression.addUnaryFunction("encode_json", v -> StringValue.of(v.toJson().toString()));
         expression.addUnaryFunction("decode_json", v -> {
             try {
-                return new ScarpetJsonDeserializer().deserialize(new JsonParser().parse(v.getString()), null, null);
+                return new ScarpetJsonDeserializer().deserialize(JsonParser.parseString(v.getString()), null, null);
```
altrisi, line 344: This is wrong, the deserializer shouldn't be called by just nulling out parameters that may be required later. Not only that, this warning is already fixed in another PR.

ch-yx: ok, i revert the latest commit.

altrisi (approved):

ch-yx:
thank you a lot for your approve!
now, does it only need gnembon's approve to be merged?

gnembon:
b64 is not using anything outside of ascii so encoding don't matter, but hey. If it doesn't matter it can be utf_8

ch-yx:
> b64 is not using anything outside of ascii so encoding don't matter, but hey. If it doesn't matter it can be utf_8

b64 itself ofc is ascii only. but the string the function received/yield could be out of ascii.

before that is merged, decode_b64(encode_b64('我')) is not ‘我’ and the result is depends on the os `s default encoding.

surely that is a bug. and im glad that is fixed.

PR #1375, opened 2022-03-26
Profile categories splitting and descriptions.

Crec0:
- Splits `Spawning and Random Ticks` into `Spawning`, ~~`Ice and Snow Ticks`~~ and `Random Ticks`
- Splits `Blocks` into `Tile ticks` and `Block Events`


Crec0:
Also fixed the order in profiler mixins. Mojang moved ChunkSource down in ServerLevel#tick() in 1.17.1 and carpet was still working like pre 1.17.1

Sideeffect: Environment included raid and tile ticks inside it. One whole category was missed.

src/main/java/carpet/utils/CarpetProfiler.java
```diff
@@ -31,16 +33,38 @@
...
-    private static final String[] SCARPET_SECTIONS = {
-            "Scarpet run", "Scarpet events", "Scarpet schedule",
-            "Scarpet command", "Scarpet load", "Scarpet app data", "Scarpet client"
-    };
-    private static final String[] SECTIONS = {
-            "Spawning and Random Ticks", "Ticket Manager","Unloading",
-            "Blocks", "Entities", "Block Entities",
-            "Entities (Client)", "Block Entities (Client)",
-            "Village", "Environment"};
+    private static final ImmutableMap<String, String> GENERAL_SECTIONS = ImmutableMap.of(
```
altrisi, line 36: No need to use guava here:

```suggestion
    private static final Map<String, String> GENERAL_SECTIONS = Map.of(
```

Same for the other ones. You can use Map.ofEntries(Entry...) if there's more than the `of()` static methods allow (with static import of `Map.entry()` for convenience).

Crec0:
Did the requested `Map` changes

Renamed `Tile ticks` to `Schedule ticks` because that's what they really are

Crec0:
Don't merge it yet. There's an Issue of negative ms being shown in `whatever the rest might be` category. I will fix it later today

altrisi:
Is the rest being negative still an issue or is this rfr?

Crec0:
Yes still an issue. will fix today.
Yes those changes are intentional. it was targeting both `tick` methods therefore injecting in both.

Crec0:
I think this is ready to be reviewed and merged

altrisi (approved):

PR #1380, opened 2022-03-30
Sculk sensor range

fxmorin:
Rule to change the sculk sensor range. Much like powered rail limit


src/main/java/carpet/CarpetSettings.java
```diff
@@ -1102,4 +1102,13 @@ public String description() {
...
     public static boolean creativePlayersLoadChunks = true;
 
+    @Rule(
+            desc = "Customizable sculk sensor range",
+            options = {"8", "16", "32"},
+            category = CREATIVE,
+            strict = false,
+            validate = PushLimitLimits.class
+    )
+    public static int sculkSensorRange = 8;
```
gnembon, line 1112: 32? interesting.

Dev note - the performance will be terrible above 16, but hay, its for experiments, right?

PR #1390, opened 2022-04-02
New Logger: `updateStackCount`

fxmorin:
It will tell you a stacks (block updates & shape updates) Chained Neighbor Update count.
Only does this if the chain is over 25
Moving from #1378, to the master branch

### DIFF
--- a/src/main/java/carpet/logging/LoggerRegistry.java
+++ b/src/main/java/carpet/logging/LoggerRegistry.java
@@ -28,6 +28,7 @@ public class LoggerRegistry
     public static boolean __pathfinding;
     public static boolean __explosions;
     public static boolean __updateSuppressedCrashes;
+    public static boolean __updateStackCount;
 
     public static void initLoggers()
     {
@@ -48,7 +49,7 @@ public class LoggerRegistry
         registerLogger("mobcaps", HUDLogger.stardardHUDLogger("mobcaps", "dynamic",new String[]{"dynamic", "overworld", "nether","end"}));
         registerLogger("explosions", Logger.stardardLogger("explosions", "brief",new String[]{"brief", "full"}, true));
         registerLogger("updateSuppressedCrashes", Logger.stardardLogger("updateSuppressedCrashes", null,null));
-
+        registerLogger("updateStackCount", Logger.stardardLogger("updateStackCount", null,null));
     }
 
     /**
--- a/src/main/java/carpet/mixins/CollectingNeighborUpdater_loggerMixin.java
+++ b/src/main/java/carpet/mixins/CollectingNeighborUpdater_loggerMixin.java
@@ -1,0 +1,38 @@
+package carpet.mixins;
+
+import carpet.logging.LoggerRegistry;
+import carpet.utils.Messenger;
+import net.minecraft.network.chat.BaseComponent;
+import net.minecraft.world.level.redstone.CollectingNeighborUpdater;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(CollectingNeighborUpdater.class)
+public class CollectingNeighborUpdater_loggerMixin {
+
+    @Shadow
+    private int count;
+
+
+    @Inject(
+            method = "runUpdates()V",
+            at = @At(
+                    value = "INVOKE",
+                    target = "Ljava/util/ArrayDeque;clear()V",
+                    shift = At.Shift.BEFORE,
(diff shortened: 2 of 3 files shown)


gnembon:
An idea - why not it taking a parameter defining the initial depth? 25 hardcoded doesn't feel like useful for everybody?

fxmorin:
There we go, now it can be limited to any depth you want

fxmorin:
To think that me wanting to add a simple logger resulted in me making a logger for type caching. Alright time to fix a couple other logger issues in another pr I guess

src/main/java/carpet/logging/TypeLogger.java
```diff
@@ -0,0 +1,141 @@
...
+        }
+        else if (type.isEnum())
+        {
+            String ucValue = value.toUpperCase(Locale.ROOT);
+            return (T) Enum.valueOf((Class<? extends Enum>) type, ucValue);
+        }
+        else
+        {
+            return null;
+        }
```
altrisi, line 139: What if the converting function could be specified when constructing? For thought, though definitely not blocking this PR, can be done later.

src/main/java/carpet/logging/TypeLogger.java
```diff
@@ -0,0 +1,141 @@
...
+        parsedPlayerOptions.remove(playerName);
+        super.removePlayer(playerName);
+    }
+
+    /**
+     * serves messages to players fetching them from the promise
+     * will repeat invocation for players that share the same option
+     */
+    @FunctionalInterface
+    public interface typeMessage<T> { BaseComponent [] get(T playerOption, Player player);}
```
altrisi, line 72: Naming

src/main/java/carpet/logging/TypeLogger.java
```diff
@@ -0,0 +1,141 @@
...
+            }
+        }
+    }
+
+    /**
+     * guarantees that each message for each option will be evaluated once from the promise
+     * and served the same way to all other players subscribed to the same option
+     */
+    @FunctionalInterface
+    public interface typeMessageIgnorePlayer<T> { BaseComponent [] get(T playerOption);}
```
altrisi, line 92: Naming

src/main/java/carpet/logging/logHelpers/PathfindingVisualizer.java
```diff
@@ -2,37 +2,29 @@
...
 import carpet.helpers.ParticleDisplay;
 import carpet.logging.LoggerRegistry;
+import carpet.logging.TypeLogger;
 import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.phys.Vec3;
 
 public class PathfindingVisualizer
 {
+    @SuppressWarnings("unchecked")
```
altrisi, line 12: Ugh, big block for suppressing... Isn't there a more specific place to put it? I know on local declarations you can put it, not sure if in calls?

src/main/java/carpet/logging/logHelpers/UpdateStackCountHelper.java
```diff
@@ -0,0 +1,29 @@
...
+                    (TypeLogger.typeMessageIgnorePlayer<Integer>) (option) -> {
+                try {
+                    if (count >= option) {
+                        return new BaseComponent[]{Messenger.c(
+                                "w Stack finished with: " + count + " updates"
+                        )};
+                    }
+                } catch (NumberFormatException ex) {
+                    return null;
+                }
```
altrisi, line 24: Can anything throw this anymore?

altrisi (commented):
Only gave it a quick look.

For a later refactor I feel this should be the other way around: Logger would extend TypeLogger. But at that time we'd have given them better names, don't think potentially breaking binary compat now is worth for this now when there's a few more things that could be changed.

fxmorin:
> Only gave it a quick look.
> 
> For a later refactor I feel this should be the other way around: Logger would extend TypeLogger. But at that time we'd have given them better names, don't think potentially breaking binary compat now is worth for this now when there's a few more things that could be changed.

I suggest we go for it and make Logger extend TypeLogger, this would also fix the suppressed warning blocks.
As for `parse()` in `TypeLogger`, do what you want with it. It's currently temporary and waiting for the settings refractor #1330 to instead use the new `FromStringConverter`

src/main/java/carpet/logging/logHelpers/PathfindingVisualizer.java
fxmorin: It's not worth changing, since this method is basically just the log() and I would need to separate everything individually.

altrisi:
I believe you forgot to add the mixin to the mixin json...

gnembon:
Would it make sense to make all standardLoggers essentially TypeLogger<String> or enum? Thoughts?

fxmorin:
Ya we've talked about it. I just don't have the time to do that currently.
If you want to do that, go for it

PR #1392, opened 2022-04-03
Make scarpet internal and control flow exceptions stackless

### LINKED ISSUE #1391: [Scarpet Internals] Disable control flow exceptions from creating stacktraces
I've recently learnt that making exceptions not fill the stacktrace does actually increase performance, it's not just placebo (confirmed by jdk people in a place they use exceptions internally, [openjdk/jdk#8072](https://togithub.com/openjdk/jdk/pull/8072)).

Therefore I think we should disable scarpet's control flow exceptions from filling stack traces, given the use of exceptions does seem to make those code paths slower.

The only issue from this: If an exception leaks, we won't know where does it come from. However, scarpet exceptions are not supposed to leak, and we can always remove the code that prevents filling the stacktraces if we need to reproduce a bug with leaking exceptions.

altrisi:
Resolves #1391.

Does that by introducing an abstract class that doesn't do anything when `fillInStacktrace()` is called, `StacklessRuntimeException`. See the issue for more details.

### DIFF
--- a/src/main/java/carpet/script/exception/CarpetExpressionException.java
+++ b/src/main/java/carpet/script/exception/CarpetExpressionException.java
@@ -5,7 +5,7 @@ import carpet.utils.Messenger;
 import java.util.List;
 import net.minecraft.commands.CommandSourceStack;
 
-public class CarpetExpressionException extends RuntimeException implements ResolvedException
+public class CarpetExpressionException extends StacklessRuntimeException implements ResolvedException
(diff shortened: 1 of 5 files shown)


altrisi:
Gonna merge this, it worked in the small testing I did, can always be reverted if it ends up causing issues.

Not going to add the toggle yet, can always be added later if we actually get some exception leaking.

PR #1397, opened 2022-04-04
Show local type in exception hover tooltips

altrisi:
Adds the type of the value to the hover over tooltips from the error snooper.

imagen


altrisi:
Has been working fine in my testing, merging this as it's a useful debug tool.

PR #1399, opened 2022-04-04
Logger Improvements

fxmorin:


src/main/java/carpet/logging/HUDController.java
```diff
@@ -92,27 +92,21 @@ public static void update_hud(MinecraftServer server, List<ServerPlayer> force)
...
-                        dim = Level.OVERWORLD; // OW
-                        break;
-                    case "nether":
-                        dim = Level.NETHER; // nether
-                        break;
-                    case "end":
-                        dim = Level.END; // end
-                        break;
-                }
+                player.level.dimension();
```
altrisi, line 95: Dead call

src/main/java/carpet/logging/Logger.java
```diff
@@ -74,11 +74,7 @@ public String getDefault()
     }
     public String [] getOptions()
     {
-        if (options == null)
-        {
-            return new String[0];
-        }
-        return options;
+        return options == null ? new String[0] : options;
```
altrisi, line 77: Could also set the array to an empty one at constructor if it's null instead of checking every time tbh. Not really important.

src/main/java/carpet/logging/Logger.java
```diff
@@ -219,9 +215,13 @@ public void onPlayerConnect(Player player, boolean firstTime)
...
         {
             Set<String> loggingOptions = new HashSet<>(Arrays.asList(CarpetSettings.defaultLoggers.split(",")));
-            if (loggingOptions.contains(getLogName()))
-            {
-                LoggerRegistry.subscribePlayer(playerName, getLogName(), getDefault());
+            String logName = getLogName();
+            for (String str : loggingOptions) {
+                String[] vars = str.split(" ");
+                if (vars[0].equals(logName)) {
+                    LoggerRegistry.subscribePlayer(playerName, getLogName(), vars.length == 1 ? getDefault() : vars[1]);
```
altrisi, line 222: It's probably not supported/not a thing anywhere, but this still wouldn't work if a logger has parameters with spaces, as only the first parameter would be passed (`vars[1]`). I think it can be solved with:
```suggestion
                String[] vars = str.split(" ", 2);
                if (vars[0].equals(logName)) {
                    LoggerRegistry.subscribePlayer(playerName, getLogName(), vars.length == 1 ? getDefault() : vars[1]);
```

src/main/java/carpet/logging/Logger.java
```diff
@@ -241,15 +241,15 @@ public void onPlayerDisconnect(Player player)
 
     public String getAcceptedOption(String arg)
     {
-        if (options != null && Arrays.asList(options).contains(arg)) return arg;
+        if (Arrays.asList(this.getOptions()).contains(arg)) return arg;
```
altrisi, line 244: This will NPE if `this.getOptions()` is null.

src/main/java/carpet/logging/Logger.java
```diff
@@ -241,15 +241,15 @@ public void onPlayerDisconnect(Player player)
...
         return null;
     }
 
     public boolean isOptionValid(String option) {
         if (strictOptions)
         {
-            return Arrays.asList(this.options).contains(option);
+            return Arrays.asList(this.getOptions()).contains(option);
```
altrisi, line 251: Same here.

altrisi (changes requested):
Some small things, after that pending testing.

src/main/java/carpet/logging/Logger.java
fxmorin: `this.getOptions()` can't return null

altrisi:
Fixed Component conflicts from 22w16b.

altrisi (approved):

altrisi:
Crashes Carpet TIS Addition hard... Well, it's not the only thing in this update I guess... (by far)

PR #1401, opened 2022-04-04
Update `updateSuppressionBlock`  for 1.19

fxmorin:
Very basic switch from the old update suppression block to the new one


src/main/java/carpet/mixins/LevelAccessor.java
```diff
@@ -0,0 +1,12 @@
+package carpet.mixins;
+
+import net.minecraft.world.level.Level;
+import net.minecraft.world.level.redstone.NeighborUpdater;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+@Mixin(Level.class)
+public interface LevelAccessor {
```
gnembon, line 9: I suggest you add that to the existing interface class. I know its called WorldInterface - feel free to rename it.

fxmorin: omg that's funny, I checked if a levelAccessor already existed. Ya ill rename the current one

src/main/java/carpet/CarpetSettings.java
```diff
@@ -1045,53 +1044,27 @@ public String validate(CommandSourceStack source, ParsedRule<String> currentRule
...
-    public static int getInteger(String s) {
-        try {
-            return Integer.parseInt(s);
-        } catch(NumberFormatException e) {
-            return -1;
-        }
-    }
-
-    private static class updateSuppressionBlockModes extends Validator<String> {
+    private static class updateSuppressionBlockModes extends Validator<Integer> {
```
altrisi, line 1062: Should've been caught earlier, class naming.

src/main/java/carpet/CarpetSettings.java
```diff
@@ -1045,53 +1044,27 @@ public String validate(CommandSourceStack source, ParsedRule<String> currentRule
...
-                }
-            }
+        public Integer validate(CommandSourceStack source, ParsedRule<Integer> currentRule, Integer newValue, String string) {
+            if (newValue < -1) newValue = -1;
             return newValue;
         }
-        @Override
-        public String description() {
-            return "Cannot be negative, can be true, false, or # > 0";
-        }
```
altrisi, line 1094: Maybe keep some description, so there's some feedback to the user.

src/main/java/carpet/CarpetSettings.java
```diff
@@ -1045,53 +1044,27 @@ public String validate(CommandSourceStack source, ParsedRule<String> currentRule
...
-                    if (parsedInt <= 0) {
-                        updateSuppressionBlockSetting = -1;
-                        return "false";
-                    } else {
-                        updateSuppressionBlockSetting = parsedInt;
-                    }
-                }
-            }
+        public Integer validate(CommandSourceStack source, ParsedRule<Integer> currentRule, Integer newValue, String string) {
+            if (newValue < -1) newValue = -1;
```
altrisi, line 1065: Return `null` if invalid, else users will expect their value to be set. `null` tells them it's an invalid value and cancels the operation.

src/main/java/carpet/mixins/CollectingNeighborUpdaterAccessor.java
```diff
@@ -0,0 +1,13 @@
...
+import net.minecraft.world.level.redstone.CollectingNeighborUpdater;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+@Mixin(CollectingNeighborUpdater.class)
+public interface CollectingNeighborUpdaterAccessor {
+    @Accessor("count")
+    void setCount(int count);
+    @Accessor("maxChainedNeighborUpdates")
+    int getMaxChainedNeighborUpdates();
```
altrisi, line 12: Aren't the annotation parameters implicit by the method names? At the end of the day mixin will have to add it to the refmap anyway.

src/main/java/carpet/mixins/Level_tickMixin.java
```diff
@@ -19,14 +20,20 @@
...
+    @Shadow @Final protected NeighborUpdater neighborUpdater;
     CarpetProfiler.ProfilerToken currentSection;
     CarpetProfiler.ProfilerToken entitySection;
 
     Map<EntityType<?>, Entity> precookedMobs = new HashMap<>();
 
+    @Override
+    public NeighborUpdater getNeighborUpdater() {
+        return this.neighborUpdater;
+    }
```
altrisi, line 35:
```suggestion
    @Override
    @Unique
    public NeighborUpdater getNeighborUpdater() {
        return this.neighborUpdater;
    }
```
I think Unique can/should go there, if so why not, may save us the mixin if mojang adds it some day.

Needs an import.

altrisi (commented):
Looks good other than some small things, pending testing from my side.

altrisi:
Update suppression/update order noob trying to test update suppression related PR. Result: ??? (got the log message about suppression, got no idea how to make that stop a portal from being destroyed).

Code makes sense though.

altrisi (approved):

PR #1404, opened 2022-04-08
Improve error message when commands fail to be added and make commands using the new system check for masking

altrisi:
The command no longer can only come from vanilla. This should make it clearer for users, given it's not that uncommon of an issue, especially when worldedit has a `cs` command and a lot of people like to call their cam app `cs`. Made command errors get prefixed with "Failed to load app 'name':" as well.

Also fixes an issue with that checker not running unless adding a command via the legacy system (else it merges the trees).


altrisi:
TBD: Seems like some apps are now depending on "adding" to existing commands, at least with the `allow_command_conflicts` flag. Wondering what should be done with this change.

PR #1409, opened 2022-04-15
Edited scarpet's documentation

XavierWah:
Edited filenames in scarpet's documentation so they look better.


Ghoulboy78:
You would also have to modify mergedoc.bat and mergedoc.sh so that the automatic concatenation works

altrisi:
The `.bat` one is not really used (or usable iirc), but the sh one yes. About the changes, right now most of those are named after their implementation (`Auxiliary.md` -> `Auxiliary.java`, `Scoreboard.md` -> `Scoreboards.java`), changing this may make it slightly more difficult to find some. Not against it though, up to gnembon to decide.

altrisi:
It's fine, it's not very obvious.

PR #1411, opened 2022-04-16
Fix bad inject in PistonHandler_customStickyMixin for production

KJP12:
This fixes a bug that can cause the production environment to crash or fail to push blocks.

Note, using pistons in the development environment is broken as a result of this change for unknown reasons; is the LVT mangled by Loader?


KJP12:
Hm, this is interesting; this is actually a mod compatibility problem?

Since this doesn't affect pure vanilla, closing.

KJP12:
After some investigating, it appears that the problem happens with certain mass ASM transformers such as [the experimental McUnsafe](https://github.com/the-glitch-network/McUnsafe), but oddly, doesn't happen just GudASM or Plymouth Anti-Xray applied. As this problem seems to be only for specific environments, this PR will be useless in 99.99% of cases. I just happened to hit the 0.01% where such a nuance exists.

Issue #1386, opened 2022-04-02
[Commands] [Scarpet] Port `/info` to a Scarpet app


altrisi:
Scarpet apps are more maintainable given they don't depend on the game source directly (so only the language has to be updated once for all apps to work) and having Carpet functionality as scarpet apps also works as tests for the scarpet API.

I propose `/info` to be reimplemented as a Scarpet app, given it is quite a simple concept if scarpet has access to all the data `/info` provides, and if it doesn't, this can be a good tracking issue for adding them.

gnembon:
yes. This just shows the gap in the API. There are more 'exotic' measures, like random wandering chances, but that could be measured in scanning the area for blocks, so not a biggie..

Another vector would be informative aspects of /spawn - could also be scarpet only.

PR #1414, opened 2022-04-19
Adding `provides_power` and `srong_power` functions

Ghoulboy78:
This is a step towards accomplishing #1386. It fulfils the second and third checkboxes from the bottom.

### DIFF
--- a/docs/scarpet/api/BlocksAndWorldAccess.md
+++ b/docs/scarpet/api/BlocksAndWorldAccess.md
@@ -460,6 +460,15 @@ Boolean function, true if the block causes suffocation.
 
 Numeric function, returning redstone power level at position.
 
+### `provides_power(pos)`
+
+Boolean function, true if the block can provide power.
+
+### `strong_power(pos)`
+
+Numeric function, returning strong redstone power level at position
+(i.e. via observer, repeater, comparator or torch).
+
 ### `ticks_randomly(pos)`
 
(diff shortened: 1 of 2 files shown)


ekardnam:
Wouldn't `power_level`/`strong_power_level`/`direct_power_level` be a bit more explanatory as a name?

Ghoulboy78:
there is already a `power()` function, so this fits in with the existing function. Plus, the docs are always there cos people may still be confused as to the difference, cos these functions both work non-intuitively (at least for first-time users).

Ghoulboy78:
`direct_power` is power coming from a redstone component pointing into a block, or from a redstone component itself. This is different from `power` which can also be power from an adjacent powered block.
Ik that the names are confusing, but these is the "official" terminology

Ghoulboy78:
> How to call it, idrk tbh.

That was another issue I faced, which is why I used mojmap name. Also, I explained it in the docs, which is why I used that name.
IMO we should ask in the discord whether or not people find it excessively confusing, and if they can suggest any alternative names. Cos they're the people that r gonna be using it anyways.

Ghoulboy78:
@altrisi I think we can go ahead with this naming then.

altrisi:
Don't think it was never asked in discord

Ghoulboy78:
It was, just no one responded so I forgot about it. I'll ask again.

PR #1415, opened 2022-04-19
Updating Scarpet docs for Auxiliary.md

Ghoulboy78:
Given that no one seems to be taking this up, and given that I have a bunch of free time, I might as well do this myself. This is a step towards accomplishing the goals outlined in #837.
I will basically go in order down the list and submit prs for all of them.


docs/scarpet/api/Auxiliary.md
```diff
@@ -4,33 +4,48 @@ Collection of other methods that control smaller, yet still important aspects of
...
+<pre>
+//Plays the sound of a zombie breaking down a door, with 4x the volume and 1/4th the pitch
+//Turning down ambient music in sound settings will reduce the volume of this sound
+sound('entity.zombie.break_wooden_door', [x, y, z], 4, 0.25, 'ambient')
+</pre>
 
 ## Particles
 
+Functions which allow to send particles of various forms to the player. These allow to highlight zones for all players,
+including those with vanilla clients (unlike with `draw_shape` function which only works for players with carpet clients)
```
altrisi, line 29: Not 100% true, many shapes have fallbacks, there's just a few that don't (most if not all _could_ have a fallback, just it's too complicated/no one bothered). IMO this comment should be in `draw_shape` if it's not there.

Ghoulboy78: ye true

Ghoulboy78:
oh shit, I did not realise what merging with upstream would do. Imma just make a new PR, this is a mess...

PR #1423, opened 2022-04-21
Fix "Unable to determine descriptor" warnings by setting `remap=false` at some locations.

Crec0:
Tested in prod. Didn't crash. Loaded everything fine and worked without issues.


altrisi:
I don't use idea so I don't know, doesn't this mean the `@SuppressWarnings("UnresolvedMixinReference")` annotations are no longer needed in these?

Crec0:
I believe so. Will remove them.

altrisi (approved):
Latest commit also works fine in prod 🚀

PR #1433, opened 2022-04-25
Making deepslate generate below y=0, not y=16

Ghoulboy78:
Cos generating below y=16 was only for 1.17


Ghoulboy78:
Good pint, that

altrisi (approved):

altrisi:
Why not, let's get it in. It was warned already, even if the change comes later than expected.

PR #1443, opened 2022-05-16
Prevent assigning variables to `UNDEF`

### LINKED ISSUE #1439: Assigning an undefined variable to another variable should throw an exception
For example `a = b` (where `b` is undefined) should throw an exception instead of being legal.

(RFC)

altrisi:
Closes #1439.

Testing wanted! Especially if you have `strict` scripts with loops type `map`, `filter`, `reduce`, etc.

Even though I mentioned my rationale on why it should work, I'm still not 100% sure `UNDEF` can't get into one of those places. All I know is randomly I've seen stuff like `Value.TRUE` or `Value.NULL` with a `boundVariable`.

This has had some successful testing, but sending as a draft because it still looks fragile to me. I believe a single script managing to set a `boundVariable` to `Value.UNDEF` could cause all apps in `strict` mode to stop working.

### DIFF
--- a/src/main/java/carpet/script/value/UndefValue.java
+++ b/src/main/java/carpet/script/value/UndefValue.java
@@ -137,4 +137,18 @@ public class UndefValue extends NullValue {
         throw getError();
     }
 
+    @Override
+    public Value reboundedTo(String var) {
+        if (getVariable() != null) {
+            // Here we're making a few assumptions:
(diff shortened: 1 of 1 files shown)


altrisi:
If I get bound variables outside, this is no longer the way to do it...

But then there's a better way for this, throwing/giving a throwing lv when getting empty from the context. And UNDEF itself would be gone.

altrisi:
Going to close this to wait for the bound variables refactor, where it's much simpler and actually gets rid of undef.

Issue #1403, opened 2022-04-07
Scarpet: Add `__on_chunk_unloaded(x, z)`


James103:
As of Carpet mod 1.4.70, `__on_chunk_loaded` exists as a bindable event for a Scarpet app, but there is no equivalent `__on_chunk_unloaded` event.

If and when `__on_chunk_unloaded` is implemented, it would take in `x` and `z` parameters which are the lowest x and z coordinates of the chunk that is about to be unloaded. The event would be called right before the chunk unloads, allowing an app to save some info related to the chunk such as when it was last loaded and unloaded.

For example, an app needs to calculate offline progress on a chunk by chunk basis because it, for example, massively slows down crop growth, therefore requiring extra random ticks when the chunk loads in order to catch up. Without this event, said app would need to update the last loaded time of each chunk every game tick or every second, which would impact server performance more than the proposed event.

Ghoulboy78:
This could be done by mixing into `ServerLevel::unload()` function.

PR #1451, opened 2022-05-21
Added `__on_chunk_unloaded` scarpet event

ekardnam:
Hello it is my first PR on here (hopefully first of many). I just added a `__on_chunk_unloaded` event as requested in #1403

### DIFF
--- a/docs/scarpet/api/Events.md
+++ b/docs/scarpet/api/Events.md
@@ -103,6 +103,9 @@ this should not happen, but let you be warned.
 Called right after a chunk at a given coordinate is loaded. All newly generated chunks are considered loaded as well.
  `x` and `z` correspond to the lowest x and z coordinates in the chunk.
 
+### `__on_chunk_unloaded(x, z)`
+Called right before a chunk at the given coordinates is unloaded. `x` and `z` correspond to the lowest x and z coordinates in the chunk.
+
 ### `__on_lightning(block, mode)`
 Triggered right after a lightning strikes. Lightning entity as well as potential horseman trap would 
 already be spawned at that point. `mode` is `true` if the lightning did cause a trap to spawn. 
--- a/src/main/java/carpet/mixins/ServerLevel_scarpetMixin.java
+++ b/src/main/java/carpet/mixins/ServerLevel_scarpetMixin.java
@@ -26,6 +26,8 @@ import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
 
 import static carpet.script.CarpetEventServer.Event.EXPLOSION;
 import static carpet.script.CarpetEventServer.Event.LIGHTNING;
+import static carpet.script.CarpetEventServer.Event.CHUNK_UNLOADED;
+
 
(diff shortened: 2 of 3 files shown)


docs/scarpet/Full.md
```diff
@@ -4756,6 +4756,9 @@ this should not happen, but let you be warned.
 Called right after a chunk at a given coordinate is loaded. All newly generated chunks are considered loaded as well.
  `x` and `z` correspond to the lowest x and z coordinates in the chunk.
 
+### `__on_chunk_unloaded(x, z)`
+Called right before a chunk at the given coordinates is unloaded. `x` and `z` correspond to the lowest x and z coordinates in the chunk.
+
```
Ghoulboy78, line 4761: You don't need to modify Full.md, it gets done automatically

Ghoulboy78:
Also if you add "Fixes #1403" to the first message then that issue will be automatically be closed when this gets merged

docs/scarpet/Full.md
ekardnam: Thank you for the nudge and sorry. I was thinking it could have been the case, but I didnt find anything quickly looking thru actions, my bad

ekardnam:
Sure I am going to force push and change the commit message tonight

ekardnam:
What would be a good way to test this?

Ghoulboy78:
You could add the event to a script like this:
`test.sc`:
```
__on_chunk_unloaded(x, y)->print(str('Unloaded chunk (%s, %s)', x, y));
```

Then fly slowly in a cardinal direction (for eg. north), and when you cross a chunk border you should get that message printed a few times. Exactly how many times depends on the render distance, and is going to be 2*render distance + 1

ekardnam:
That is kinda what I did, except I did not count the times. It was not being triggered in the beginning, but I think that is because I would spawn in spawn chunks.

ekardnam:
Oh thanks, good to know. Was thinking of doing a scarpet app that pushes loaded chunks in a list when they are loaded and removes them when unloaded and I could use code from `overlay.sc` to see them in game. It feels like a more reliable way to test then just counting the times the event is triggered

ekardnam:
Gonna implement that script asap, I'm in uni exam session so I dont have too much time. Also don't forget my implementation follows your suggestion so dont be so harsh on yourself :).
For your question i dont really know

altrisi:
Don't worry about implementing a complex test script for the event, as long as it works and testing is simple it's fine. If needed it can be always added later.

There's not really a way to automatically test scarpet rn anyway.

With this in mind, would you say this is ready?

ekardnam:
> Don't worry about implementing a complex test script for the event, as long as it works and testing is simple it's fine. If needed it can be always added later.
> 
> There's not really a way to automatically test scarpet rn anyway.
> 
> With this in mind, would you say this is ready?

I will just test it quickly again then

ekardnam:
Id say it works

altrisi (approved):
LGTM, thanks for contributing!

PR #1470, opened 2022-06-22
add zeroTickPlants rule

silverisntgold:
"Restores the ability to force plants (Bamboo [NOT SHOOTS], Cactus, Chorus Flower, and Sugar Cane) to advance their growth stage with zero-ticking"
I'm not sure what carpets policy on @ Overwrite is, but in this case I believe it's a better approach than @ Inject, since we need the else branch in order to ensure destroyBlock() and randomTick() remain mutually exclusive. This ensures that growing plants with zero-ticking will work exactly as it did before being patched, since this mirrors the vanilla implementation that caused the behavior.


Ghoulboy78:
AFAIK carpet's policy is to not reintroduce bugs which were fixed, especially ones which gnembon himself has called cheaty.
That said. it may be that he or altrisi allow this, but I think it won't be cos it was a bug report which got fixed, same as how gnembon didn't bring back piston translocation.

PR #1471, opened 2022-06-23
new: fakePlayersLoadChunks carpet rule

Simba98:
enabled by default: behavior of fake players like vanilla players
disabled: fake players do not load chunks like spectators with gamerule spectatorsGenerateChunks false

This rule allows the server to save performance while they are using fake players.
Server owners can use /forceload to load only a few selected chunks.
It would be useful for farms. (Neither creative nor spectator mode cannot be used for this purpose)

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -1068,6 +1068,15 @@ public class CarpetSettings
     public static boolean creativePlayersLoadChunks = true;
 
     @Rule(
+            desc = "Fake players load chunks, or they don't! Just like spectators!",
+            extra = {"Like creativePlayersLoadChunks",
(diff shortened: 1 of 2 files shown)


Fallen-Breath:
Notes that players who returning `true` might behave differently in mob spawning logics, even if it's not a spectator. Have you really tested if your rule really helps your mob farming thing?

And anyway it's much more better to just make a (void) perimeter for the mob farm for less lag, either in cmp or smp.

Simba98:
> Notes that players who returning `true` might behave differently in mob spawning logics, even if it's not a spectator. Have you really tested if your rule really helps your mob farming thing?
> 
> And anyway it's much more better to just make a (void) perimeter for the mob farm for less lag, either in cmp or smp.

Oh, I tested with a farm. Wheat is growing like normal.

PR #1475, opened 2022-06-26
Adding system_info('world_min_spawning_light')

Ghoulboy78:
Title says it all. Note that it takes into account modifications fro mdatapacks.

### DIFF
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -710,7 +710,7 @@ system calls. In all circumstances, these are only provided as read-only.
   * `world_center` - Returns coordinates of the center of the world with respect of the world border
   * `world_size` - Returns radius of world border for current dimension.
   * `world_max_size` - Returns maximum possible radius of world border for current dimension.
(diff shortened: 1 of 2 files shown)


ch-yx:
Sounds like it's referring to the spawning cap. . . Is it possible to have the word light in the rule name?

Ghoulboy78:
True that

Ghoulboy78:
Should I backport this into 1.18.x? Cos that's the version where you were first able to change this with a datapack, and it may be useful info

altrisi:
Carpet doesn't usually do backports, I don't think it should be tbh, it'd likely not even be released.

Ghoulboy78:
fair enough.

Ghoulboy78:
@altrisi ok, ig this is good to go

altrisi:
Always felt like it has a very long name, that's why I didn't merge it, but given idk how to call it didn't propose anything but wait to see if maybe someone came up with something better.

Feature itself looks good, though don't remember if I tested it.

altrisi (approved):
Looks good!

What would you think about calling it `world_min_spawning_light` or something like that?

Ghoulboy78:
Yes. I didn't really like this name, thought it was too clunky.

PR #1476, opened 2022-06-27
Making portal_timer and portal_cooldown return portal timer and portal cooldown in that order

### LINKED ISSUE #1465: `query(e, 'portal_timer')` and `query(e, 'portal_cooldown')` are reversed
`query(e, 'portal_cooldown')` results in how long the player has been standing in a nether portal, while `query(e, 'portal_timer')` results in the player's portal cooldown, counting down as they exit a portal. This is backwards to what the docs indicate.

Note, the inverse `modify()` code appears to be correct. This issue only applies to `query()`.

Tested with 1.4.79
It appears to have been this way since they were added to the API.

Relevant code:
https://github.com/gnembon/fabric-carpet/blob/81d3b9e04f2f3acde2ab4fd27252af41227f383f/src/main/java/carpet/script/value/EntityValue.java#L526-L527

Ghoulboy78:
Fixes #1465 
A silly mistake


altrisi:
Note that there's apps using it that will suddenly start malfunctioning because of the change. Is it worth it? I think it is tbh, but you should also change `ai_tracker` to use the proper one and potentially open a PR to the scarpet repo to change the `world_map` app (assuming it's using it on purpose and knowing it was wrong).

Ghoulboy78:
true that

Ghoulboy78:
As for world_map.sc, it appears that it was actually broken without realising, cos it used portal_timer as if it actually returned the portal_timer.

PR #1480, opened 2022-07-01
Change the default gamemode for fake players

disymayufei:
Many minecraft server administrators choose to give their server's players a 3rd-part way（like a website, or a bot for social media apps）to spawn the fake player, and these 3rd-part way always execute the spawn command from the server source, and that makes the spawned fake player is creative mode, that's not safe for a server! So I try to make the deafult gamemode to survival.

### DIFF
--- a/src/main/java/carpet/commands/PlayerCommand.java
+++ b/src/main/java/carpet/commands/PlayerCommand.java
@@ -278,7 +278,7 @@ public class PlayerCommand
                 () -> DimensionArgument.getDimension(context, "dimension").dimension(),
                 () -> source.getLevel().dimension() // dimension.getType()
         );
-        GameType mode = GameType.CREATIVE;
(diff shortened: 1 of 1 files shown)


CommandLeo:
You can just add `in survival` to the /player command to make the bots spawn in survival

altrisi:
If you are making an automated tool you can always just `/gamemode` them right after spawning, I don't really see why the default should be changed for something that can change it easily.

Ghoulboy78:
> But I found that the option "in" will conflict with the option “at”，so if I want to spawn the fake player to the specific position, that method will not work

I'm fairly certain the extended version of the command allows to put in both...
image

Ghoulboy78:
You can hardly get more specific than that now...

altrisi:
Going to close this for now for the above reasons, though feel free to give more arguments at Discord then this can be reopened. Changing the default for no good reason is likely to break other stuff.

PR #1493, opened 2022-07-14
Fix #1492

### LINKED ISSUE #1492: Placing redstone wire does not always update connections when `fastRedstoneDust` is enabled
The recent update to the `fastRedstoneDust` setting introduced a bug where placing redstone wire in a powered position would see indirect shape updates omitted entirely, meaning any neighboring wires diagonally above or below would not update their connections and remain in an invalid state.

SpaceWalkerRS:
This PR fixes #1492 by re-introducing indirect shape updates.


altrisi (approved):
Fixes the issue

PR #1497, opened 2022-07-16
[Scarpet] Added ability to cancel events

### LINKED ISSUE #455: [Suggestion] Events termination statuses
Be able to send termination statuses of events returning either 'pass', 'fail', 'success' or 'consume'  to cause the call chain to stop and that status returned. If your callback doesn't return anything (null) then is not cancellable.

replaceitem:
Adds the ability to cancel events by returning `'cancel'` in the event.
This will stop minecraft from processing the event.

* ❌ Are not (yet) cancellable
* ✅ Can be cancelled
* ⛔ Not applicable (Cannot be cancelled with current implementation, or just doesn't make sense)

### Cancellable events include (for now):

⛔ `server_starts`
⛔ `server_shuts_down`
⛔ `tick`
⛔ `chunk_generated`
⛔ `chunk_loaded`
⛔ `chunk_unloaded`
⛔ `player_jumps`
❌ `player_deploys_elytra` (could in theory be cancelled, but causes client to get stuck in fly mode - too buggy)
⛔ `player_wakes_up`
⛔ `player_escapes_sleep`
⛔ `player_rides`
✅ `player_uses_item`
✅ `player_clicks_block` (when breaking blocks that are instaminable, blocks disappear client side)
✅ `player_right_clicks_block`
⛔ `player_interacts_with_block` (happens after minecraft handles it -> not cancellable)
✅ `player_placing_block`
⛔ `player_places_block` (happens after minecraft handles it -> not cancellable)
✅ `player_breaks_block`
✅ `player_interacts_with_entity`
⛔ `player_trades`
⛔ `player_picks_up_item`
✅ `player_attacks_entity`
⛔ `player_starts_sneaking`
⛔ `player_stops_sneaking`
⛔ `player_starts_sprinting`
⛔ `player_stops_sprinting`
⛔ `player_released_item`
✅ `player_finished_using_item` (creates ghost items)
✅ `player_drops_item` (creates ghost items)
✅ `player_drops_stack` (creates ghost items)
✅ `player_chooses_recipe`
⛔ `player_switches_slot`
✅ `player_swaps_hands`
⛔ `player_swings_hands`
✅ `player_takes_damage`
✅ `player_deals_damage`
⛔ `player_collides_with_entity`
⛔ `player_dies`
⛔ `player_respawns`
⛔ `player_changes_dimension`
⛔ `player_connects`
⛔ `player_disconnects`
✅ `player_message`
⛔ `statistic`
⛔ `lightning`
⛔ `carpet_rule_changes`
⛔ `explosion_outcome`
❌ `explosion` (Can be cancelled, but requires a return value for vanilla, not sure if `null` just works, or if that creates errors and an expty explosion needs to be returned)

#### Todo

- [x] Adding support for cancelling events
- [x] Making events cancellable
- [x] Allowing to cancel `__on_player_places_block`
- [x] Docs

Resolves #455


James103:
If `player_takes_damage` is cancelled, but the player's HP is set to 0 in the event handler during that event, does the player die normally?

To test:
1. Run: `/script run __on_player_takes_damage(p,a,s,se) -> ( modify(p, 'health', 0); return('cancel'); );`
2. Run: `/script load event_test`
3. Run: `/script load stats_test`
4. If the player is not in Survival mode, run `/gamemode survival`
5. Take damage.
6. See whether the player dies normally (death screen if `/gamerule doImmediateRespawn false`, `custom.deaths` statistic incrementing, and chat output showing `__on_player_dies` was invoked)?

replaceitem:
Looks like the player does die normally. I get a normal death screen
grafik

James103:
Therefore, cancelling `player_takes_damage` while setting player HP to 0 in that event handler results in a proper death. However, the death message is incorrect -- it is always "[player] died" instead of what the death message would have been.

replaceitem:
If you run `modify(p, 'health', 0)` on its own, the player also dies (supposedly on the next tick).
Therefore i think the death is not actually associated with the original reason for the damage.

replaceitem:
Added docs now, should be ready i hope.

gnembon:
exciting :)
rebase :(

src/main/java/carpet/mixins/Player_scarpetEventsMixin.java
```diff
@@ -59,21 +63,26 @@ private void onEntityCollision(Entity entity, CallbackInfo ci)
...
 
-    @Inject(method = "interactOn", at = @At("HEAD"))
+    @Inject(method = "interactOn", cancellable = true, at = @At("HEAD"))
     private void doInteract(Entity entity, InteractionHand hand, CallbackInfoReturnable<InteractionResult> cir)
     {
         if (!level.isClientSide && PLAYER_INTERACTS_WITH_ENTITY.isNeeded())
         {
-            PLAYER_INTERACTS_WITH_ENTITY.onEntityHandAction((ServerPlayer) (Object)this, entity, hand);
+            if(PLAYER_INTERACTS_WITH_ENTITY.onEntityHandAction((ServerPlayer) (Object)this, entity, hand)) {
+                cir.setReturnValue(InteractionResult.PASS);
```
gnembon, line 72: note for the future - there is a difference between PASS / SUCCESS / CONSUME etc. I think events should allow for more than just 'cancel', but that could be specified later.

replaceitem: Yeah, might do those too at some point in a seperate PR

gnembon (approved):

Ghoulboy78:
Just as a hilarious POC, we should write an app which cancels all the events it possibly can and see what you can actually do in minecraft with all those events cancelled.

altrisi:
This broke Carpet Extra :(. Super simple fix though, but still needs a rebuild. Hopefully it didn't affect many other mods adding events.

PR #1500, opened 2022-07-18
Fixed typos in README

ZeroKun265:


altrisi:
I'd say `Hwat` is intended. Always thought it was at least.

ZeroKun265:
Now that you said it i can see it, both how and what start with an h, but as a non native speaker it just took me a while to understand and i didn't get the joke

PR #1503, opened 2022-07-21
Fix handling files with uppercase letters in case-sensitive systems

### LINKED ISSUE #1502: [scarpet] Failure to read contents of files containig uppercase characters
Steps to reproduce:

1. Create a script `test.sc` and a corresponding data directory `test.data` in a world's scripts folder.
2. Place a txt file in the data directory with some random contents, `testing`, for example. Crucially: name the file `Test.txt`, mind the uppercase T.
3. Edit `test.sc` to read from the file like so: `print(read_file('Test', 'text'))`
4. When loading the app with `/script load test`, the app prints `null`.

Expected behaviour:
When loading the app, it should print `testing`.

MC version: 1.18.2
Carpet version: 1.4.66

altrisi:
Should fix #1502.

Windows file system is case-insensitive, so doing a `toLowerCase` on the file names doesn't matter. However, others, such as Linux, do not ignore upper/lower case, so this was preventing scarpet from reading or modifying files that weren't fully lower-case. Luckily I believe files created by scarpet should have gotten lower case given they're handled by this path so they didn't break, but if you put the files by yourself (or through some other code path, if there is) they do.

### DIFF
--- a/src/main/java/carpet/script/argument/FileArgument.java
+++ b/src/main/java/carpet/script/argument/FileArgument.java
@@ -140,7 +140,7 @@ public class FileArgument
 
     public static Pair<String,String> recognizeResource(String origfile, boolean isFolder, Type type)
     {
-        String[] pathElements = origfile.toLowerCase(Locale.ROOT).split("[/\\\\]+");
+        String[] pathElements = origfile.split("[/\\\\]+");
         List<String> path = new ArrayList<>();
         String zipPath = null;
         for (int i =0; i < pathElements.length; i++ )


Firigion:
I'd add a line in the docs in this PR to make it explicit that the file handling system in scarpet is case insensitive.

altrisi:
This is actually making it case-sensitive.

altrisi:
There's a small chance that apps that were relying on this behaviour of file names getting lowercase will break or have data that can no longer be read on linux until it's recreated with proper case, but I think it's worth it, plus this'll go out for 1.19.1, so it's not that much of a breaking change inside a same release.

Firigion:
I think making the apps work the same independently of the OS is the preferred behaviour. Particularly since there is no way from within scarpet (as far as I can tell) to find the server's OS. Enforcing case-insensitibity, or wahtever you call it sounds lik ehte better idea to me.

altrisi:
That'd cause issues in those systems though, given you could have two files that when lowercased are the same but aren't normally. And then we'd probably also have to iterate through everything in the folder(s) in order to make sure to find one that matches, given we can't just give every case combination to the file system (else files with upper case on systems that care wouldn't be found, like happens right now, we can't predict what characters will have different case).

Plus, this way it should work pretty much the same in both systems, just on linux you'll be able to save/read two different files that only differ in case (saved files will keep the casing), but that's a thing of the filesystem that already happens with other stuff.

gnembon:
Well, I thought initially that for simplicity sake it should only handle lowercased files, like variable names were case insensitive. Afaik, that changed. 

With this, I am not opposed to add case sensitivity to files read by file API. I don't see a value in that (i think it potentially could create issues with people that start with programming), but I am not opposed to it either.

Would it make sense to support casiness? I assume altrisi is already for it.

gnembon:
Given the issues with non-win systems, I would agree to add casiness for consistent behaviour, but I recall there is quite a bit of that mentioned in the docs and that needs to be removed.

Edit - docs don't mention that at all. It makes sense to add that immediately without changing of the docs.

PR #1505, opened 2022-07-22
Refactor Carpet's sticky blocks code

SpaceWalkerRS:
Carpet's sticky blocks code is a bit of a mess at the moment. All the logic is in one Mixin class, making it hard to read and add new custom behavior. This refactor moves all the logic into two methods in the blocks' classes themselves. This makes it fairly simple to add new custom sticky behavior.


gnembon:
I have been trying to get back to carpet for a few weeks already, and then I get stuck on this one. I agree that's the proper way to do it (via adding methods to block / blockstate to handle behaviour of different blocks), just need to convince myself its the proper way to do it. Gimme a sec.

src/main/java/carpet/mixins/BlockStateBase_customStickyMixin.java
```diff
@@ -0,0 +1,31 @@
...
+    @Shadow private Block getBlock() { return null; }
+    @Shadow private BlockState asState() { return null; }
+
+    @Override
+    public boolean isSticky() {
+        return ((BlockBehaviourInterface)getBlock()).isSticky(asState());
+    }
+
+    @Override
+    public boolean isStickyToNeighbor(Level level, BlockPos pos, BlockPos neighborPos, BlockState neighborState, Direction dir, Direction moveDir) {
```
gnembon, line 28: I know that technically accessing blocks is customary to be done via BlockState, but in this case you only use it as a proxy to get to your block information. Seems you might just use (BlockBehaviourInterface)state.getBlock() to get that info, right?

gnembon:
looks great. One comment though if you really need one extra layers of interfaces and mixins to get to the block directly.

src/main/java/carpet/mixins/BlockStateBase_customStickyMixin.java
SpaceWalkerRS: That's a fair point. It could be removed without issue.

PR #1507, opened 2022-07-22
Add blockpos and chunkpos getters for EntityValue

Crec0:


altrisi:
`blockpos` doesn't sound right to me for the API. Yes, the type that is used internally is a `BlockPos`, but on scarpet that means nothing, it's just giving the position of the entity but as integers. I don't think `block` adds any info for someone who doesn't know what a `BlockPos` is. And the docs don't give add much more info either.

Also please use the functions in `ValueConversions` when possible (`of(BlockPos)`), so it can be all updated easily and all has the same conversion impl.

gnembon:
Not sure if 'chunk_pos' is needed. Chunks are internal concepts that don't (shouldn't) surface to the gameplay. For this reason for instance chunk events, like chunk_loaded or chunk_generated provides block positions of the lowest blockpos, not the chunk pos so we don't introduce new concepts that seem redundant.

Having said that - I think its actually not a good idea to have it in entity API.

What would be viable is a more generic helper function in the form of `chunk_pos(x, y, z), chunk_pos(block), chunk_pos([x, y, z]), chunk_pos(entity)` etc. returning consistent output for various viable inputs.

But this boils down to the age old question is that needed in the language, or should be supplied with libraries (since it doesn't require the java code to do).

But I think its needed, at least for the performance improvements, same as we need other helper functions like `distance(..)` etc. and that would be good to propose at once.

maybe...

I suggest yeet chunk_pos for now and maybe propose the new helper category and start it with `chunk_pos()` method.

docs/scarpet/api/Entities.md
```diff
@@ -125,15 +125,26 @@ accessed via UUID, but should be accessed with their name instead.
...
 Triple of the entity's position
 
+`block_pos` is rounded down / floored position of the entity.
+
+<pre>
+query(p, 'pos') => [420.19801928, 69.3928301, 420.4098232]
+query(p, 'block_pos') => [420, 69, 420]
+</pre>
+
+### `query(e, 'chunk_pos')`
```
gnembon, line 139: yeet that - info in PR comments

gnembon (changes requested):

Crec0:
Removed `chunk_pos`. 
What you said is correct, it should be generic for all types. As for now, its not really needed (as in no one has asked for it or requested it), but if someone does ask for it in the future, I can PR the generic helper method for all types.

PR #1513, opened 2022-07-25
Don't add `Mod` to CF release to prevent cutting game version

altrisi:
Multiple people have reported crashes because this is what CurseForge shows for the latest versions for snapshots:
imagen

It was just that they were trying to run it on 1.19 release.

This removes the `Mod` from there so more stuff fits and hopefully helps people not get lost.


gnembon:
Ahh, gotcha - took a moment to realize. 
Sure

PR #1519, opened 2022-07-29
Refactor spawn reporting

altrisi:
WIP

More specifically, stuff around the `SpawnReporter` class. It's not my plan to refactor the entirety of how spawn reporting works.

Done:
- Cache MobCategory.values() for performance (that method makes a copy)
- Use primitive-specific collections for performance
- Improve performance by reducing lots of hashing and hash operations
  - This is helped further by the primitive-specific collections given they have `addTo` methods that do so in one operation, and help readability
- Rename `ServerWorldInferface` to its mojmap name, given I'll likely be using it
- Revise naming
- Made `SpawnReporter` not self reset in class init given it's loaded by a call to reset anyway, gets rid of `null` `MinecraftServer` handling

Planned:
- Store reporting info in the `World` (`Level`) instance itself, or else at least change the pairs to a record `CategoryInLevel`


gnembon:
All fine so far, except from maybe storing the info in the Level instance. Mods do not own this and I 'feels better' to me to store it separately. Wonder why you like to store it there?

altrisi:
I thought about it basically to prevent another few map operations, given there's a lot of them going on here that could be saved by just attaching them to the Level. IIRC it could be reduced to just a simple array access. And it'd also remove having to instantiate these pairs for the queries.

Not really needed, already removed a bunch of those operations and it's not super hot code, but thought about it. I'll instead just change the maps to have records as keys then.

altrisi:
Going to leave further changes for later to get this in, already fixed what wasn't right.

PR #1525, opened 2022-08-05
Add event for commands and make `__on_player_message` not trigger for them

### LINKED ISSUE #1511: Event "__on_player_message" isn't triggered when runs command
https://github.com/gnembon/fabric-carpet/blob/master/docs/scarpet/api/Events.md#__on_player_messageplayer-message
> ### `__on_player_message(player, message)`
> Triggered when a player sends a chat message or runs a command.

isn't triggered when run command.
test on MC 1.19 + carpet 1.4.79 and MC 1.19.1-rc2 + carpet 1.4.82

Ghoulboy78:
Fixes #1511


replaceitem:
I think it would be better to split them into two events. Vanilla handles them seperately, and most of the time, scripts would want to check whether the message is a command or not anyway. This is impossible here, since (iirc) the commands get sent without the `/`

Ghoulboy78:
Yes they do

altrisi:
Then missing docs + docs update + pr title update.

Though it's worth considering given it'll break apps relying on the other event triggering for commands.

gnembon:
vanilla splits it into two for reasons that are irrelevant at this level. But it would make sense to split them to two events - I am ok with either. 

but as altrisi says: missing docs, update editor keywords, update event_test.sc, and that would be it.

Ghoulboy78:
Yep

altrisi:
Build failed :(

Ghoulboy78:
oh bruh

Ghoulboy78:
OOh I see the issue. Somehow the latest version of the code got lost somewhere. Lemme fix rq

Ghoulboy78:
Actually, it appears that the events API was updated, and I didn't notice when resolving conflicts. fixed now

altrisi (approved):

altrisi:
idk why this wasn't merged already

altrisi:
Oh it's not cancellable. Well I don't think that's that bad tbh

Issue #1238, opened 2021-12-25
[Scarpet Bug] Caches are never cleared


altrisi:
For example, `selectorCache` in `EntityValue` is never cleared, meaning that with enough unique selectors the game will run out of memory.

While it's true that many times entity selectors will be constant, there's apps (source: support request in carpetcord) that use entity selectors with multiple, variable selectors, usually depending on a position. Those could fill the allocated memory and make the game run out of memory.

In singleplayer, this problem is aggravated because the cache isn't cleared even when switching worlds, since the cache is static in `EntityValue`.

This problem could be made less bad by at least moving it to somewhere that is discarded on world change, like `CarpetScriptServer`, but original problem is still present given enough time and executions.

Other caches that may be affected, but likely to a smaller degree, are `particleCache` in `ParticleDisplay` (also in `ShapeDispatcher`), `itemCache` in `NBTSerializableValue` or `bvCache` in `BlockValue`.

gnembon:
all caches should have some limit and be cleared when reaching it. I (or someone else) can go over these and make sure they have a limited capacity

altrisi:
Would it be a good idea to use a WeakHashMap for those that use the String in the StringValue directly? Would ensure that the constant ones are kept around for the lifetime of the app, but the rest would be garbage collected at the gc's discretion.

PR #1528, opened 2022-08-09
Merge particle caches

altrisi:
Until this we had two particle caches, one in ParticleDisplay and one in ShapeDispatcher.

With this ShapeDispatcher uses the cache from ParticleDisplay.

Can this cause issues with concurrent access to both?

Note that caches are still never cleared, relates to #1238.

Pending testing.


gnembon:
your suggestion about WeakHashMap is great I think

gnembon:
Do you want to add WeakHashMap use here? or is it something you would want to add later in a separate pr?

altrisi:
I wasn't planning to do it in this PR, given there's also some other caches that would need looking, and something I didn't check back then was the extra overhead that handling weak keys + reference queues (for eviction) may cause (or whether it isn't really measurable).

PR #1529, opened 2022-08-09
Get rid of remaining log4j references

altrisi:
Minor breaking API change, but it's been deprecated for (IMO) long enough ago given it was mentioned it'd be removed earlier than the rest of the API.

Pending Actions build to make sure the full build doesn't break and maybe investigating remaining usages of the API in extensions out there.


altrisi:
Sooo many extensions depend on this methods still

altrisi:
I'll reopen it when it's safer.

PR #1530, opened 2022-08-09
Migrate Carpet's rules to the new settings system

altrisi:
Does that.

While doing this I was wondering if the language file should maybe be datagen-ed from some annotation closer to the rule.

Maybe something like
```java
@RuleData(desc = "...", extra = {"...", "..."})
@Rule(...)
```
?

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -1,11 +1,11 @@
 package carpet;
 
 import carpet.api.settings.CarpetRule;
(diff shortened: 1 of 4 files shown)


gnembon:
Yes, I found it annoying in extra that rule definition and what it does is done somewhere else. 
If you could pull off that via datagen, that would be ideal.
That file could very well be in the repo and commited as such, just like Full.md is

gnembon:
it should be possible to use the existing @Rule fields, am I correct?

PR #1531, opened 2022-08-10
trying fix #1515. seems that it works fine.

### LINKED ISSUE #1515: if player drops the item he received from inventory_set fast enough, it will not trigger vanilla inventory_changed advancement.
use a repeating command block :
/player chenyuxuan dropStack all

give me many a apple via inventory_set:
/script run inventory_set(player('chenyuxuan'),0,3,'apple')

the advancement is not triggerd at all. neither getting nor dropping triggers the advancement.

but if i give myself apples by /give or picking up item or some thing:
the advancement will be triggerd twice (get&drop). 

_found when testing a bug in my script fakeplayer_inv_editor. that player could use it to clone items._

ch-yx:
that pr is to fix #1515.

i tested it . seems that it works fine.

But I'm not familiar with this part of code. 

 Please correct me if there are any mistakes. please leave your comment.

### DIFF
--- a/src/main/java/carpet/script/api/Inventories.java
+++ b/src/main/java/carpet/script/api/Inventories.java
@@ -258,6 +258,9 @@ public class Inventories {
                 // clear slot
                 ItemStack removedStack = inventoryLocator.inventory().removeItemNoUpdate(slot);
                 syncPlayerInventory(inventoryLocator, slot);
+                if (inventoryLocator.owner() instanceof ServerPlayer sp){
+                    sp.containerMenu.broadcastChanges();
+                };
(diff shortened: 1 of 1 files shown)


altrisi:
From a very quick look (also don't know a lot about this code) I think this is supposed to be happening in the `syncPlayerInventory` call, maybe it's that what isn't doing enough stuff or it's filtering what it shouldn't.

gnembon:
yeah, seems like syncPlayerInventory is where that could potentially be done. Seems like current behaviour of syncPlayerInventory is too leaninent.

PR #1539, opened 2022-08-23
Adding arguments to events in Notepad++ docs

Ghoulboy78:
Basically instead of suggesting `__command` or `__on_player_breaks_block` it will suggest `__command() ->` and `__on_player_breaks_block(player, block) ->`.
Will probs be modified after #1525 is merged (Or I'll modify that, based on which gets merged first)

### DIFF
--- a/docs/scarpet/resources/editors/npp/scarpet.xml
+++ b/docs/scarpet/resources/editors/npp/scarpet.xml
@@ -101,55 +101,55 @@
             <Keywords name="Keywords5">_ _i _a _x _y _z _trace</Keywords>
             <Keywords name="Keywords6">global_</Keywords>
             <Keywords name="Keywords7">
-                __command
-                __config
-                __on_chunk_generated
-                __on_chunk_loaded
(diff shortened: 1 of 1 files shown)


gnembon:
this is.. unorthodox, but...

Yeah, that is overall a good idea. 

You got me on the fence with that since these are NOT keywords of the language, but having the arguments filled is very helpful. Not sure about the arrow though - its like you are automatically generating code. 

Little context - in the past, I wanted to add all functional equivalents to all operators, same as for sum() or then(), and then -> would be weird when using defun() for example.

Thoughts on -> ?

Ghoulboy78:
well how many times are you gonna be using these without sticking a -> in there yourself?
tbf I did also only understand about half of what you said about the functional equivalents thing, but then again the reason this is only a pr to notepad ++ docs is cos you already allowed it for IDEA

gnembon:
fair enough - we can always yeet the arrows when non-operatoral defun is in.

PR #1547, opened 2022-09-17
Store unwrapped `Value` instances in `Context`

altrisi:
This PR changes `Context` to store `Value` instances instead of `LazyValue`s.

Help wanted to change the impl of `Expression.getOrSetAnyVariables`, else this PR doesn't do much more than some code cleanup, given I don't know how to properly handle the code around compilation.

Doing this:
- Reduces memory usage, given we're saving multiple objects for what actually is just on
- Should (not sure if noticeably) improve performance if `getOrSetAnyVariable` is migrated, given it could reduce a bunch of wrapping+unwrapping
- Improves code readability given we can pass non-effectively-final variables to the Value constructors

Risk to be checked:
- This means that in places like loops, the `NumericValue` instances are instantiated regardless of whether they're used. This may slightly lower performance, though it's very unlikely given its instantiation is really cheap (for long and int), and the current code has to instantiate the `LazyValue`s (they're just a regular though autogenerated class) already, that I'd assume has a similar overhead. And in case the value is used it also has to generate the value.

Also pending more testing.


altrisi:
Just realized there's some bounding issues here where I bind some Value constants, I'll fix that later.

gnembon:
lazyness of the variables has been something that have been here from day one. Not sure what use there was for it - probably to keep the lambda equality between actual representation of original reprezentation where anything is a LazyValue - a variable, constant or any other language construct, but maybe for variables there is no need to maintain that. The PR looks not as scary as I thought

altrisi:
Will rebase this on top of 1581, maybe after that's merged, and will refactor setAnyVar and try to remove undef as a `Value` at the same time.

altrisi:
Rebased on top of #1581, diff that matters for this PR is against that.

PR #1550, opened 2022-09-19
Add rule to make the World Border update with ticks instead of real time

senseiwells:
Reworks WorldBorder to update based on game ticks instead of relying on real time.

Changing tick rate (and tick warping) now affects the speed at which the world border moves.


senseiwells:
I was going to implement it behind a rule but I couldn't find a situation where you would want the world border to be in real time when you are tick warping or increasing tick rate. I suppose if your server was lagging and you had a world border it would be inconsistent with vanilla... 
Making it a rule adds a little more complexity but if you insist I don't mind implementing it.

altrisi:
The world border movement can be used to detect the game being slowed down, not sure if in survival but at least with commands I remember some item counters using it for counting `items/realtime` when the game didn't reach 20TPS.

senseiwells:
Added the rule 👍

src/main/java/carpet/CarpetSettings.java
```diff
@@ -1077,4 +1079,41 @@ public String description() {
...
+    private static class WorldBorderValidator extends Validator<Boolean>
+    {
+        @Override
+        public Boolean validate(CommandSourceStack source, CarpetRule<Boolean> changingRule, Boolean newValue, String userInput)
+        {
+            if (changingRule.value() ^ newValue)
+            {
+                // Needed for the update
+                tickSyncedWorldBorders = newValue;
+                MinecraftServer server = source.getServer();
```
altrisi, line 1092: `source` is nullable in a validator (in rule sync)

src/main/java/carpet/patches/TickSyncedBorderExtent.java
```diff
@@ -0,0 +1,160 @@
...
+import net.minecraft.world.phys.shapes.BooleanOp;
+import net.minecraft.world.phys.shapes.Shapes;
+import net.minecraft.world.phys.shapes.VoxelShape;
+import java.util.Arrays;
+import java.util.OptionalDouble;
+
+/**
+ * This class is essentially a copy of {@link net.minecraft.world.level.border.WorldBorder.MovingBorderExtent}
+ * but instead of using real time to lerp the border
+ * this class uses the in game ticks.
```
altrisi, line 19: Wouldn't it be better to extend/mixin to it then?

src/main/java/carpet/patches/TickSyncedBorderExtent.java
```diff
@@ -0,0 +1,160 @@
...
+	public double getSize()
+	{
+		double progress = this.ticks / this.tickDuration;
+		return progress < 1.0 ? Mth.lerp(progress, this.from, this.to) : this.to;
+	}
+
+	@Override
+	public double getLerpSpeed()
+	{
+		return Math.abs(this.from - this.to) / this.tickDuration;
```
altrisi, line 80: Isn't this in seconds? (or blocks/second or ms or whatever)

src/main/java/carpet/patches/TickSyncedBorderExtent.java
```diff
@@ -0,0 +1,160 @@
...
+		double ms;
+		if (server == null) 
+		{
+		    ms = TickSpeed.mspt;
+		} 
+		else 
+		{
+		     OptionalDouble optional = Arrays.stream(server.tickTimes).average();
+		     // Optional should never be empty but just in case...
+		     ms = optional.isEmpty() ? TickSpeed.mspt : optional.getAsDouble() * 1.0E-6D;
```
altrisi, line 97: Could use `orElseThrow` then

src/main/java/carpet/patches/TickSyncedBorderExtent.java
```diff
@@ -0,0 +1,160 @@
...
+	{
+		if (this.ticks++ % 20 == 0)
+		{
+			for (BorderChangeListener listener : this.border.getListeners())
+			{
+				if (!(listener instanceof BorderChangeListener.DelegateBorderChangeListener))
+				{
+					listener.onBorderSizeLerping(this.border, this.from, this.to, this.realDuration);
+				}
+			}
```
altrisi, line 138: Can't find this in 1.19.3, may have changed since it was written. Saves an AW :)

altrisi (commented):
Some comments, quite some duplicated code.

altrisi:
Also you could save an AW by calling setSize() when finishing moving instead of manually constructing the static one, though idk if it'd be an issue with the listeners.

About the Mth.average thing, I believe it's because actions is trying to build against the snapshot already.

src/main/java/carpet/patches/TickSyncedBorderExtent.java
senseiwells: No, this is not part of the vanilla code. I added this to periodically update the client. This is needed as the client relies on the WB updating in real time. Real time doesn't change but the tick rate can change - for this reason we just update the lerping state every second.

You could probably implement this better by having it only update when the tick rate actually changes but I don't think that's worth while.

src/main/java/carpet/patches/TickSyncedBorderExtent.java
senseiwells: Sure.

src/main/java/carpet/patches/TickSyncedBorderExtent.java
senseiwells: Yes, you are correct, this is a mistake - I fixed this earlier for my UHC mod impelmentation but forgot to update it here.

src/main/java/carpet/patches/TickSyncedBorderExtent.java
senseiwells: Mixing into the class I would essentially be overwriting all the methods - slashing compatability. My original implementation did actually do this, but I decided against this.
I also tried extended the MovingBorder class, but I don't think that is a good idea; there will be unused fields which will not be accurate for the behaviour of the class: `lerpDuration` and `lerpDuration`, this would also require another AW.

senseiwells:
> Also you could save an AW by calling setSize() when finishing moving instead of manually constructing the static one, though idk if it'd be an issue with the listeners.
> 
> About the Mth.average thing, I believe it's because actions is trying to build against the snapshot already.

I did try this at some point - no this wouldn't work, the `update` method of `BorderExtent` is assigned to `WorldBorder.extent`. If you called `setSize` it would be overriden by this. You can also not just `setSize` then return `WorldBorder.extent` because that field is private.

src/main/java/carpet/patches/TickSyncedBorderExtent.java
senseiwells: It's also valid to update all listeners as we are assuming that they do not know about the `tick` update behaviour of this custom implementation so they would also need to be updated. The reason I exclude `DelegateBorderChangeListener` is because that shadows all changes to other world borders in different worlds, that is not wanted.

senseiwells:
I just tried again to extend `MovingExtent`, I forgot - you cannot extend an inner class, this was probably the reason why I ended up not doing this xD.

Modifying the rending, as you said you'd need to also send packets to the client about what the current tick is as that is not done in vanilla, maybe a future PR...

altrisi:
Oh true, didn't realize it's not static, nvm that then.

PR #1559, opened 2022-10-02
Make Carpet's Network Handlers Run On Main Threads

senseiwells:
This PR ensures that Carpet's network handlers are being run on the main thread:

## Client
- The client network handler now mixins after the `handleCustomPayload` method has been pushed to the main thread.
- This wasn't direcly causing any concurrency issues but any Carpet Extensions with validators on the client side would be running off-thread and may lead to race conditions, this also would cause any other mods that mixin into Carpet's network handling to have the same issue.
- Since all the shape code is now running on the main thread the maps that store the shapes no longer needed to be synchronized.

## Server
- The server network handler now pushes the method to the main thread.
- This could have been the cause for a concurrency issue as `ServerNetworkHandler.remoteCarpetPlayers` gets updated on the Network thread. This also gets updated/iterated on the main thread. Also this caused Carpet's clientCommands to be executed off-thread (I didn't check but I suppose this could possibly cause issues).

I've tested outside of a dev environment on both the client and server in both on a dedicated and integrated server.

### DIFF
--- a/src/main/java/carpet/mixins/ServerGamePacketListenerImplMixin.java
+++ b/src/main/java/carpet/mixins/ServerGamePacketListenerImplMixin.java
@@ -3,6 +3,8 @@ package carpet.mixins;
 import carpet.network.CarpetClient;
 import carpet.network.ServerNetworkHandler;
 import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.network.protocol.PacketUtils;
+import net.minecraft.network.protocol.game.ServerGamePacketListener;
 import net.minecraft.network.protocol.game.ServerboundCustomPayloadPacket;
 import net.minecraft.resources.ResourceLocation;
 import net.minecraft.server.level.ServerPlayer;
@@ -24,6 +26,10 @@ public class ServerGamePacketListenerImplMixin
         ResourceLocation channel = packet.getIdentifier();
         if (CarpetClient.CARPET_CHANNEL.equals(channel))
         {
+            // We should force onto the main thread here
+            // ServerNetworkHandler.handleData can possibly mutate data that isn't
+            // thread safe, and also allows for client commands to be executed
+            PacketUtils.ensureRunningOnSameThread(packet, (ServerGamePacketListener) this, player.getLevel());
             ServerNetworkHandler.handleData(new FriendlyByteBuf(packet.getData().copy()), player);
             ci.cancel();
         }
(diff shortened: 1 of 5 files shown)


src/main/java/carpet/mixins/ServerGamePacketListenerImplMixin.java
```diff
@@ -24,6 +26,10 @@ private void onCustomCarpetPayload(ServerboundCustomPayloadPacket packet, Callba
         ResourceLocation channel = packet.getIdentifier();
         if (CarpetClient.CARPET_CHANNEL.equals(channel))
         {
+            // We should force onto the main thread here
+            // ServerNetworkHandler.handleData can possibly mutate data that isn't
+            // thread safe, and also allows for client commands to be executed
+            PacketUtils.ensureRunningOnSameThread(packet, (ServerGamePacketListener) this, player.getLevel());
```
gnembon, line 32: @altrisi do you know why you needed to .copy() here? seems only carpet is using these packets

altrisi: I don't know, this is what was previously implemented in the mixin with this behavior, I just inlined it when the mixin was no longer needed.

It was previously here:
https://github.com/gnembon/fabric-carpet/blob/19ad8cfbf3a0d92eb0d133eaf4fb15bf808034d7/src/main/java/carpet/mixins/CustomPayloadC2SPacketMixin.java#L25

senseiwells: I assume it might be a result of the packet buf being released at some point and causing issues?
I'm pretty sure that it's redundant now.

src/main/java/carpet/mixins/ServerGamePacketListenerImplMixin.java
```diff
@@ -24,6 +26,10 @@ private void onCustomCarpetPayload(ServerboundCustomPayloadPacket packet, Callba
         ResourceLocation channel = packet.getIdentifier();
         if (CarpetClient.CARPET_CHANNEL.equals(channel))
         {
+            // We should force onto the main thread here
+            // ServerNetworkHandler.handleData can possibly mutate data that isn't
+            // thread safe, and also allows for client commands to be executed
```
gnembon, line 31:
```suggestion
            // thread safe, and also allows for client commands to be executed
            // [TODO] need to investigate if that is actually an issue since carpet doesn't fiddle with its packets.
```

gnembon (approved):

PR #1574, opened 2022-10-16
Fixed NPE when events are triggered within events

### LINKED ISSUE #1573: Null Pointer Exception when running veinminer on latest master
It has to do with events, so it's likely to be caused by the event cancellation API. Though I haven't investigated it, so it may not be.
[message.txt](https://github.com/gnembon/fabric-carpet/files/9794240/new.23.txt)

replaceitem:
Fixes #1573

This would happen if an event is triggered inside an event (In the case of veinminer, by calling `harvest` inside `__on_player_breaks_block`, which would trigger the event again). This makes `runIfEnabled` return null, which errors when converting `Boolean` to the primitive


altrisi (approved):
Seems to fix veinminer

PR #1575, opened 2022-10-17
small change. i want to add a function that returns item s rarity

ch-yx:
it s a small change. 
please accept it as soon as possible. XD


Ghoulboy78 (approved):
Now all that's missing is the notepad++ suggestions, but looks good otherwise

altrisi:
Question: What is rarity used for?

Ghoulboy78:
Map-making afaik

ch-yx:
> Question: What is rarity used for?

to decide items name 's color in map making.   i sometime hope to print the received item in chat like in vanilla.

PR #1577, opened 2022-10-20
fix #1576

### LINKED ISSUE #1576: when x is NaN, it equals to any number.
that is because of https://github.com/gnembon/fabric-carpet/blob/master/src/main/java/carpet/script/value/NumericValue.java#L197

the method `equals` of numbers returns  !(a-b> epsilon)
if any of a or b is NaN, it is true.

there for,
`x == x`
`x == 3`
`x == 4`
`3 == x == 4`
are all true🙃😄😅

ch-yx:


src/main/java/carpet/script/value/NumericValue.java
```diff
@@ -194,6 +194,8 @@ public boolean equals(Object o)
             if (o instanceof NumericValue no) {
                 if (longValue != null && no.longValue != null)
                     return longValue.equals(no.longValue);
+                if (Double.isNaN(this.value) || Double.isNaN(no.value))
+                    return false;
```
Ghoulboy78, line 198: Can you provide a detailed explanation of the bug and how this fix works?

2No2Name: The bug is that the code doesn't expect any NaN values. The fix adds proper handling for NaN. The proper handling for NaN is that NaN is not equal to anything, even if the other thing is also NaN (This is standardized).
The code below (subtract) returns NaN when receiving NaN as one argument, and getBoolean will return false, which in turn makes the whole function return true, which is not matching the standardized behavior of NaN.
This change fixes it by handling the NaN as a special case beforehand

PR #1582, opened 2022-11-01
Added `pt_br` translation

FITFC:


altrisi:
Thanks a lot for this!

If you want to give a try to translating the rules, you can find them here: https://github.com/altrisi/fabric-carpet/blob/migrate-rules/src/main/resources/assets/carpet/lang/en_us.json (they go directly in the same file)

FITFC:
> Thanks a lot for this!
> 
> If you want to give a try to translating the rules, you can find them here: https://github.com/altrisi/fabric-carpet/blob/migrate-rules/src/main/resources/assets/carpet/lang/en_us.json (they go directly in the same file)

Thank you so much for reporting this! I will definitely translate that part too!

altrisi:
Thanks for contributing!

PR #1584, opened 2022-11-02
Fixing `/player botname stop` with relation to fake player movement

### LINKED ISSUE #1072: `/player bot stop` command doesn't immediately stop the bot's movement
As said in the title, doing the
`/player botname stop` command
will not immediately stop the bot's movement(done using the command `/player botname move forward`). It will slow down eventually, but not immediately.
Showcase of the bug?: [This](https://youtu.be/Y9ePkqr1zzA)

Ghoulboy78:
Fixes #1072
This basically removed two if clauses that made no sense to me. If they were there for a reason, then I'll find another solution, but right now they are the cause of this bug.
Now `/player botname stop` will immediately stop the player's movement.
Todo proper testing.

### DIFF
--- a/src/main/java/carpet/helpers/EntityPlayerActionPack.java
+++ b/src/main/java/carpet/helpers/EntityPlayerActionPack.java
@@ -235,14 +235,9 @@ public class EntityPlayerActionPack
                 }
             }
         }
-        if (forward != 0.0F)
-        {
-            player.zza = forward*(sneaking?0.3F:1.0F);
-        }
-        if (strafing != 0.0F)
-        {
-            player.xxa = strafing*(sneaking?0.3F:1.0F);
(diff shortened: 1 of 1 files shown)


epifeny:
I can confirm that this works in single player. It'll take another day before I can test it in a multiplayer server.

epifeny:
I can also confirm now that I've tested it on a multiplayer server and it works as expected.

Ghoulboy78:
@gnembon
Was there a specific reason for those if statements? Cos I feel like there was, except they are the cause of this bug.

altrisi:
Minecarts

Ghoulboy78:
damn

Ghoulboy78:
I'll open a new one ig

PR #1585, opened 2022-11-04
More Ender Dragon queries

BisUmTo:
`query(e, 'dragon_nodes');`  returns Ender Dragon nodes 
`query(e, 'dragon_node_adjacency');` returns Ender Dragon node adjacency map 
`query(e, 'dragon_target');` returns Ender Dragon target pos 
`query(e, 'path');`  works with Ender Dragon too


BisUmTo:
Do you have suggestions how to rename queries?

PR #1591, opened 2022-11-07
Add `compute_density_function` to replace removed `sample_noise

Crec0:
Sorry about accidentally closing #1590


gnembon:
Thanks for contribution. First question before I get started - why name change? looks like the call does what the previous one was doing?

Crec0:
Hey,
It's not really same as before. Previously it was sampling actual noise values, howerver, now it's computing density functions which are used to sample noises either directly or after applying some operation on them. Datapacks and game internally also refers to them distinctively. Since there's a clear distinction between the two terms, in order to eleminate confusion between the two, the name is was changed to `compute_density_function`

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -1627,23 +1631,69 @@ else if (ticket == TicketType.POST_TELEPORT) // post teleport
...
+            var randomState = RandomState
+                    .create(generator.generatorSettings().value(), level.registryAccess()
+                    .lookupOrThrow(Registries.NOISE), level.getSeed());
+
+            var densityFunctionRegistry = level.registryAccess().registryOrThrow(Registries.DENSITY_FUNCTION);
+
+            var visitor = ((RandomStateVisitorAccessor) (Object) randomState).getVisitor();
+
+            var singlePointContext = new DensityFunction.SinglePointContext(pos.getX(), pos.getY(), pos.getZ());
+            var router = generator.generatorSettings().value().noiseRouter();
```
altrisi, line 1655: I believe gnembon doesn't like `var` usage.

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -1627,23 +1631,69 @@ else if (ticket == TicketType.POST_TELEPORT) // post teleport
...
+
+                if (densityFunction == null) {
+                    ResourceLocation densityFunctionKey = ResourceLocation.tryParse(densityFunctionQuery);
+                    if (densityFunctionKey == null) {
+                        throw new InternalExpressionException("Density function name '" + densityFunctionQuery + "' is invalid. Please follow the correct [a-z0-9._-]:[a-z0-9/._-] format.");
+                    }
+
+                    if (densityFunctionRegistry.containsKey(densityFunctionKey)) {
+                        densityFunctionQuery = densityFunctionKey.toString();
+                        densityFunction = densityFunctionRegistry.get(densityFunctionKey);
```
altrisi, line 1685: Given the `densityFunctionQuery` is used for mapping the results, why change it here? That means it's possible these won't be the same and it may be confusing. Also I believe ResLoc.toString returns with the `minecraft` namespace, while scarpet usually skips it (see `ValueConversions`)

altrisi (commented):
Genuinely forgot about this PR, sorry!

Some minor comments, looks good otherwise.

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -1626,23 +1634,72 @@ else if (ticket == TicketType.POST_TELEPORT) // post teleport
...
-            noiseQueries = new String[]{"continentalness", "erosion", "weirdness", "temperature", "humidity", "depth"};
+    public Value compute_density_function(Context c, @Locator.Block BlockPos pos, String... densityFunctionQueries) {
+
+        Map<Value, Value> result = new HashMap<>();
+
+        ServerLevel level = ((CarpetContext) c).level();
+        Registry<DensityFunction> densityFunctionRegistry = level.registryAccess().registryOrThrow(Registries.DENSITY_FUNCTION);
+
+        if (densityFunctionQueries.length == 0) {
+            return ListValue.wrap(densityFunctionRegistry.keySet().stream().map(ResourceLocation::toString).map(StringValue::of).toList());
```
altrisi, line 1645: `Stream.toList()` creates an immutable list, while Scarpet lists should be mutable (and `wrap` doesn't copy the list).
```suggestion
            return ListValue.wrap(densityFunctionRegistry.keySet().stream().map(ResourceLocation::toString).map(StringValue::of));
```

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -1626,23 +1634,72 @@ else if (ticket == TicketType.POST_TELEPORT) // post teleport
...
+                    );
+                }
+            };
+
+            if (densityFunction == null) {
+                throw new InternalExpressionException(
+                    "Somehow density function was not found in the whole registry and error checking didn't trigger." +
+                    "Ideally this should never reach here."
+                );
+            }
```
altrisi, line 1698: Can this actually ever happen? If it shouldn't I'd remove it or throw a non-scarpet exception (like IllegalState/Argument)

altrisi (commented):
Some comments. Are there many density functions usually in a world? Wondering if that's the case if it may be more useful to return a set? Though I don't know as I don't have a use case. Not blocking.

src/main/java/carpet/script/api/WorldAccess.java
Crec0: No. This should never happen. I did it just to make one less warning "densityFuction could be null"

Crec0:
> Some comments. Are there many density functions usually in a world? Wondering if that's the case if it may be more useful to return a set? Though I don't know as I don't have a use case. Not blocking.

Yes. Many. You can boot the game and run command `script run compute_density_function(p~'pos')` to get the list of currently available density function. World generation datapacks add a lot of their own density functions too.

altrisi:
A set I mean when returning the available noises (length 0), was wondering if checking "do available noises contain x" was a common thing.

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -1626,23 +1634,72 @@ else if (ticket == TicketType.POST_TELEPORT) // post teleport
...
-            noiseQueries = new String[]{"continentalness", "erosion", "weirdness", "temperature", "humidity", "depth"};
+    public Value compute_density_function(Context c, @Locator.Block BlockPos pos, String... densityFunctionQueries) {
+
+        Map<Value, Value> result = new HashMap<>();
+
+        ServerLevel level = ((CarpetContext) c).level();
+        Registry<DensityFunction> densityFunctionRegistry = level.registryAccess().registryOrThrow(Registries.DENSITY_FUNCTION);
+
+        if (densityFunctionQueries.length == 0) {
+            return ListValue.wrap(densityFunctionRegistry.keySet().stream().map(ResourceLocation::toString).map(StringValue::of));
```
altrisi, line 1645: Didn't realize: Removes the `minecraft` namespace, like other functions do, and removes a map call
```suggestion
            return ListValue.wrap(densityFunctionRegistry.keySet().stream().map(ValueConversions::of));
```

altrisi (approved):
Will merge after snapshot, just got to test it one last time after the merge

Crec0:
> A set I mean when returning the available noises (length 0), was wondering if checking "do available noises contain x" was a common thing.

Oh i see. 0 names passed in is more of a debug thing instead of actual usage thing. Using a set or list is pretty much same effect. However I do see how set would be kinda better in case someone actually uses it in a loop with filters and stuff. I will make this change.

altrisi:
If it's going to be debug only I'd keep it as a list (simpler to use), was wondering if an app may want to check that a data pack function is present for example.

Though maybe it should just check for the data pack to be present in that case...

src/main/java/carpet/script/api/WorldAccess.java
altrisi:
```suggestion
            if (densityFunction == null) // IDE thinks may be null as Registry.get is Nullable, but we checked it
                throw new IllegalStateException("density function can't be null here"); 
```

PR #1594, opened 2022-11-12
just typos

lucifiel1618:


altrisi:
Full.md is autogenerated from the other two, it doesn't need to be changed.
Also having commas or not is more of a design decision, it works just fine without them.
Only one I see meaningful is disteance, will push that. The one space in a title isn't the only one so I don't think it should be fixed individually.

Thanks for finding it though!

lucifiel1618:
in the example the intended output is `'player at: 567, -2423, 124'`.
You can find commas in it, which is not consistent with the input.
I think you need to either have or remove the commas in both.

altrisi:
Oh true, completely missed that!
Sorry then

altrisi:
Fixed now!

PR #1612, opened 2022-12-26
Fixed drop decay gamerules not working properly.

LagPixelLOL:
Fixed drop decay gamerules work exactly the opposite when using optimized tnt.


altrisi:
Seems to fix it from a quick test, thanks for finding it!

PR #1617, opened 2023-01-02
Add `quasiConnectivityRange` setting

SpaceWalkerRS:
The `quasiConnectivity` setting could be made redundant, but I opted to keep it for the sake of backwards compatibility. The `quasiConnectivityRange` setting can be configured to a value between 1 and 383 (no point in going higher due to the world height limit).

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -347,6 +347,21 @@ public class CarpetSettings
     @Rule(desc = "Pistons, droppers and dispensers react if block above them is powered", category = CREATIVE)
     public static boolean quasiConnectivity = true;
 
+    private static class QuasiConnectivityRangeValidator extends Validator<Integer> {
+
+        @Override
+        public Integer validate(CommandSourceStack source, CarpetRule<Integer> changingRule, Integer newValue, String userInput) {
+            return newValue > 0 && newValue < 384 ? newValue : null;
+        }
+    }
+
+    @Rule(
+        desc = "The range at which pistons, droppers, and dispensers check for 'quasi power'",
+        category = CREATIVE,
+        validate = QuasiConnectivityRangeValidator.class
+    )
+    public static int quasiConnectivityRange = 1;
+
     @Rule(
             desc = "Players can flip and rotate blocks when holding cactus",
             extra = {
--- a/src/main/java/carpet/helpers/QuasiConnectivity.java
+++ b/src/main/java/carpet/helpers/QuasiConnectivity.java
@@ -1,0 +1,21 @@
+package carpet.helpers;
+
+import carpet.CarpetSettings;
+
+import net.minecraft.core.BlockPos;
+import net.minecraft.world.level.Level;
+
+public class QuasiConnectivity {
+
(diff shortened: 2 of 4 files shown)


JacktheDevel0per:
Thoughts on changing the validator to get the highest world height out of Al dimensions (I'm not sure where that is possible it should not matter with the latter change) and use a min call on the current dimension's max height and the setting? And ofc you can do -1 because there needs to be a piston but that does not seem to matter. Hope this makes sense and I'm not saying something very dumb

SpaceWalkerRS:
Thanks for the good suggestions! The setting now respects the current world's height (so it won't check for redstone power above the world height unnecessarily) and the validator now allows values up to the largest world height of any dimension.

src/main/java/carpet/CarpetSettings.java
```diff
@@ -347,6 +347,28 @@ public String description() {
...
     public static boolean quasiConnectivity = true;
 
+    private static class QuasiConnectivityRangeValidator extends Validator<Integer> {
+
+        @Override
+        public Integer validate(CommandSourceStack source, CarpetRule<Integer> changingRule, Integer newValue, String userInput) {
+            int minRange = 1;
+            int maxRange = 1;
+
+            for (Level level : source.getServer().getAllLevels()) {
```
altrisi, line 357: Note that `source` is nullable in a Validator (for example during rule sync)

altrisi (commented):
Works, code looks good.

Wondering about changing making the qc rule accept the number directly (maybe as a string for a fallback for `false`? Though idk if many really have it set to false)

altrisi:
No, for compat with the current boolean it'd need to be a String (validators aren't ran if it can't be parsed into the type), and then parse the number in the validator (and save it to a field to not have str->int conversions at runtime).

SpaceWalkerRS:
Hmm, I'm not sure that's necessarily better than the split settings. I'll leave up to you whether or not I change it.

altrisi:
The correct answer is "we need a ~better~ way for migrating rules", but that's not a thing.
I personally don't think that many people using carpet have qc disabled and changing it could break something, plus I believe it's not the first time some rule has been upgraded.

I'd go with changing qc rule to int tbh. Mention it in the patch notes (including for the first release to stable) and be done with it.

SpaceWalkerRS:
Alright, I merged the two settings into one. So now there's just one int setting that controls the range.

altrisi (approved):

src/main/java/carpet/CarpetSettings.java
```diff
@@ -344,8 +344,32 @@ public String description() {
     )
     public static boolean antiCheatDisabled = false;
 
-    @Rule(desc = "Pistons, droppers and dispensers react if block above them is powered", category = CREATIVE)
-    public static boolean quasiConnectivity = true;
+    private static class QuasiConnectivityValidator extends Validator<Integer> {
+
+        @Override
+        public Integer validate(CommandSourceStack source, CarpetRule<Integer> changingRule, Integer newValue, String userInput) {
+            int minRange = 1;
```
altrisi, line 351: Given the rule now replaces qc false the minimum should be zero.
```suggestion
            int minRange = 0;
```

PR #1620, opened 2023-01-06
allow function `block` accept the same format as function `set` & `block_state`

ch-yx:
i copy some code from function `set` so that it can use the same format as set too.
the format is used in return value of `block_state()` too.

eg:
`block('iron_trapdoor',{'half'->'top'},nbt)`

that is to avoid some unnecessary changing maps into string. (then it need to turn back inside the function)

### DIFF
--- a/src/main/java/carpet/script/api/WorldAccess.java
+++ b/src/main/java/carpet/script/api/WorldAccess.java
@@ -264,7 +264,14 @@ public class WorldAccess
             {
                 throw new InternalExpressionException("Block requires at least one parameter");
             }
-            BlockValue retval = BlockArgument.findIn(cc, lv, 0, true).block;
+
+            BlockArgument locator = BlockArgument.findIn(cc, lv, 0, true);
+            BlockValue retval = locator.block;
+            if (lv.size() > locator.offset)
+            {
+                var rl = find_blockstate_extra_args(lv, locator.offset, retval.getBlockState(), retval.getString());
+                return new BlockValue((BlockState) rl[1], (CompoundTag) (rl[0] == null ? retval.getData():rl[0]));
+            }
             // fixing block state and data
             retval.getBlockState();
             retval.getData();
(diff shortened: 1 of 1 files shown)


ch-yx:
emmmm.....git conflit.......😟

ch-yx:
😀 I finally fix the git conflit...... now i need to check if the "set" function works as it used to be.

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -1797,4 +1760,62 @@ public static double sampleNoise(NoiseRouter router, ServerLevel level, String w
         }
         return DensityFunctions.zero();
     });
+
+    private static Object[] find_blockstate_extra_args(List<Value> lv, int start,BlockState sourceBlockState,String name_in_exception){
```
gnembon, line 1764: name (camelCase), whitespaces, and return a typed Pair (at least)

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -1797,4 +1760,62 @@ public static double sampleNoise(NoiseRouter router, ServerLevel level, String w
...
+        CompoundTag data = null;
+        List<Value> args = new ArrayList<>();
+        for (int i = start, m = lv.size(); i < m; i++)
```
gnembon, line 1767: no need to copy array here.

gnembon: I guess, that's copy-pasted code, so its not your fault. nvm

src/main/java/carpet/script/api/WorldAccess.java
```diff
@@ -750,52 +757,8 @@ else if (!("any".equals(statusString)))
...
-                    Property<?> property = states.getProperty(paramString);
-                    if (property == null)
-                    {
-                        throw new InternalExpressionException("Property " + paramString + " doesn't apply to " + sourceLocator.block.getString());
-                    }
-                    String paramValue = args.get(i + 1).getString();
-                    sourceBlockState = setProperty(property, paramString, paramValue, sourceBlockState);
-                }
+                var rl = find_blockstate_extra_args(lv,sourceLocator.offset,sourceBlockState,sourceLocator.block.getString());
+                data=(CompoundTag) rl[0];sourceBlockState=(BlockState) rl[1];
```
gnembon, line 761: press Enter here

src/main/java/carpet/script/api/WorldAccess.java
ch-yx: solved

src/main/java/carpet/script/api/WorldAccess.java
ch-yx: ok

ch-yx:
ok. i have fixed all the things you mentioned above.
 and i have added that usage in the document.

PR #1635, opened 2023-01-21
Fix unchecked removal of commands by app name

### LINKED ISSUE #1634: Unloading an app destroys any command with its name, regardless of ownership
Unloading an app will currently remove a command with the app's name, regardless if it's that app's command or a vanilla/mod's one.

That includes failing to load an app.

altrisi:
Fixes #1634.

Should this make it into a 1.19.3 release? This isn't directly on top of 1.19.3, but it doesn't include any 23w03a commits, so it could be merged into 1.19.3 without many issues by just changing the target branch. Update: Actually it's sitting on top of a commit with an NPE.

The bug is present in all 1.19.3 versions, and on the latest 1.19.2 version. Older versions are unaffected.

### DIFF
--- a/src/main/java/carpet/script/CarpetScriptHost.java
+++ b/src/main/java/carpet/script/CarpetScriptHost.java
@@ -83,6 +83,7 @@ public class CarpetScriptHost extends ScriptHost
     Predicate<CommandSourceStack> commandValidator;
     boolean isRuleApp;
     public AppStoreManager.StoreNode storeSource;
+    boolean hasCommand;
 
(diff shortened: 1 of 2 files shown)


src/main/java/carpet/script/CarpetScriptHost.java
```diff
@@ -401,7 +408,7 @@ public Boolean addAppCommands(Consumer<Component> notifier)
                 LiteralArgumentBuilder<CommandSourceStack> command = readCommands(commandValidator);
                 if (command != null)
                 {
-                    scriptServer().server.getCommands().getDispatcher().register(command);
+                	registerCommand(command);
```
altrisi, line 411:
```suggestion
                    registerCommand(command);
```

PR #1642, opened 2023-01-23
Do not load clientside-only classes in serverside ShapeDispatcher code

vlad2305m:
Built and tested:
Screenshot_20230123_212508

### DIFF
--- a/src/main/java/carpet/script/utils/ShapesRenderer.java
+++ b/src/main/java/carpet/script/utils/ShapesRenderer.java
@@ -563,7 +563,7 @@ public class ShapesRenderer
             else
                 RenderSystem.enableCull();
             
-            bufferBuilder.begin(shape.mode, DefaultVertexFormat.POSITION_COLOR);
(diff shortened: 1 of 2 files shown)


vlad2305m:
WTF

src/main/java/carpet/script/utils/ShapesRenderer.java
```diff
@@ -563,7 +563,7 @@ public void renderFaces(Tesselator tessellator, BufferBuilder bufferBuilder, dou
             else
                 RenderSystem.enableCull();
             
-            bufferBuilder.begin(shape.mode, DefaultVertexFormat.POSITION_COLOR);
+            bufferBuilder.begin(VertexFormat.Mode.values()[shape.mode], DefaultVertexFormat.POSITION_COLOR);
```
altrisi, line 566: Not reviewed yet, but mentioning that a call to Enum.values() makes a copy of the array. It should probably be cached somewhere instead.

gnembon: I would also not count on enum to stay correct forever, just list them up here in statics. Will do that after merge

PR #1644, opened 2023-01-25
Do not issue deprecation warning when correctly using `scoreboard_add` but an objective already exists

altrisi:
According to the docs, it is supported to call `scoreboard_add` when a scoreboard with that name already exists as long as that didn't change anything.

https://github.com/gnembon/fabric-carpet/blob/6adccfd46ace8e9d0e103fdd453cdf1480167e29/docs/scarpet/api/Scoreboard.md?plain=1#L14

However, that was causing a deprecation warning to be issued as that was made instantly once it was found that an objective with that name exists. This PR delays the warning to after checking the criterion doesn't match.

This introduces a compatibility risk though (with the old deprecated behaviour): People querying a criterion with a single argument will get `null` if it's `dummy`, as it's the default.

I wonder if it'd be a good idea to deprecate the default argument and requiring the criterion to always be passed as a parameter. Would allow preserving the old behaviour in that case, and also fixing the amount of parameters of the function (after deprecation removal). What do you think?

CC @replaceitem

### DIFF
--- a/src/main/java/carpet/script/api/Scoreboards.java
+++ b/src/main/java/carpet/script/api/Scoreboards.java
@@ -132,18 +132,15 @@ public class Scoreboards {
             else
             {
                 String critetionName = lv.get(1).getString();
-                criterion = ObjectiveCriteria.byName(critetionName).orElse(null);
-                if (criterion==null)
-                {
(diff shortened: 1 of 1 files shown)


replaceitem:
I think i would rather leave the criterion argument optional. Not sure how to handle the compat issue in that case though.

PR #1654, opened 2023-02-05
removed static reference to server in annotation parser

gnembon:


altrisi:
Looking into this, the Context is actually already passed to this system, the only thing that really has to change is that the entity and dimension converters need to use a converter that uses the other method, `checkAndConvert`, instead of the simple `convert`. I only used it because at the time it was the simplest and I think there were other places doing this.

I'll revert the changes here and implement those instead.

altrisi:
Though it does make sense to have that for some things looking at it again... Issue is it kind-of renders `checkAndConvert` useless other than for multiparam stuff.

altrisi:
Kept it, added docs and an exception fix instead.

PR #1665, opened 2023-03-02
fake-players-teleporting-fix:

### LINKED ISSUE #1190: [Bug] Player bots don't load chunks in 1.18
Showcase: https://media.discordapp.net/attachments/894946828980723762/913542111150866442/2021-11-25_16-23-34-converted2.mp4

Through testing I've found that this can partially be fixed uncommenting the 3 lines in EntityPlayerMPFake that are this: https://github.com/gnembon/fabric-carpet/blob/6150e1f29ca84b52529f3fd055e420ff355bb249/src/main/java/carpet/patches/EntityPlayerMPFake.java#L135
 and updating them to `this.getWorld().getChunkManager().updatePosition(this)`, however this doesn't fully fix it. If a bot is teleported to unloaded chunks (through /tp, /spreadplayers, etc), then it doesn't load those chunks until a real player loads them.

Steps to reproduce (tp-specific issue):
* `/player Steve spawn`
* `/tp Steve 69420 100 1337`
* `/script run loaded(pos(player('Steve')))` (= false)

silnarm:
* Fix fake players not being moved on the chunkmap when teleporting
* Fixes #1190 (Player bots cannot be teleported to unloaded chunks)

### DIFF
--- a/src/main/java/carpet/patches/NetHandlerPlayServerFake.java
+++ b/src/main/java/carpet/patches/NetHandlerPlayServerFake.java
@@ -4,8 +4,11 @@ import net.minecraft.network.Connection;
 import net.minecraft.network.chat.Component;
 import net.minecraft.network.chat.contents.TranslatableContents;
 import net.minecraft.network.protocol.Packet;
+import net.minecraft.network.protocol.game.*;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.network.ServerGamePacketListenerImpl;
+import net.minecraft.world.entity.RelativeMovement;
+import java.util.Set;
 
 public class NetHandlerPlayServerFake extends ServerGamePacketListenerImpl
 {
@@ -27,6 +30,26 @@ public class NetHandlerPlayServerFake extends ServerGamePacketListenerImpl
             ((EntityPlayerMPFake) player).kill(message);
         }
     }
+    private boolean hasSpawned = false;
+
+    @Override
+    public void teleport(double d, double e, double f, float g, float h, Set<RelativeMovement> set)
+    {
+        super.teleport(d, e, f, g, h, set);
+
+        handleAcceptTeleportPacket(
+            new ServerboundAcceptTeleportationPacket(this.awaitingTeleport)
+        );
+
+        if (!hasSpawned) {
+            hasSpawned = true;
+        } else {
+            handleMovePlayer(
+                new ServerboundMovePlayerPacket.PosRot(player.getX(), player.getY(), player.getZ(), player.getYRot(), player.getXRot(), false)
+            );
+        }
+    }
+
 }
 
 


src/main/java/carpet/patches/NetHandlerPlayServerFake.java
```diff
@@ -4,8 +4,11 @@
 import net.minecraft.network.chat.Component;
 import net.minecraft.network.chat.contents.TranslatableContents;
 import net.minecraft.network.protocol.Packet;
+import net.minecraft.network.protocol.game.*;
```
gnembon, line 7: please not star imports.

src/main/java/carpet/patches/NetHandlerPlayServerFake.java
```diff
@@ -27,6 +30,26 @@ public void disconnect(Component message)
...
+        super.teleport(d, e, f, g, h, set);
+
+        handleAcceptTeleportPacket(
+            new ServerboundAcceptTeleportationPacket(this.awaitingTeleport)
+        );
+
+        if (!hasSpawned) {
+            hasSpawned = true;
+        } else {
+            handleMovePlayer(
```
gnembon, line 47: what stops you from calling that every time (for simplicity sake)?

src/main/resources/carpet.accesswidener
```diff
@@ -12,3 +12,4 @@ accessible class net/minecraft/world/level/biome/Biome$ClimateSettings
 accessible method net/minecraft/world/level/border/WorldBorder getListeners ()Ljava/util/List;
 
 accessible field net/minecraft/world/level/block/state/BlockBehaviour UPDATE_SHAPE_ORDER [Lnet/minecraft/core/Direction;
+accessible field net/minecraft/server/network/ServerGamePacketListenerImpl awaitingTeleport I
```
gnembon, line 15: hmm, please resolve access issues with an interface / accessor rather than access widener.

src/main/java/carpet/patches/NetHandlerPlayServerFake.java
silnarm: Oops. IntelliJ did that, I had imported many things from there whilst investigating. My commit though, so my bad.

src/main/java/carpet/patches/NetHandlerPlayServerFake.java
silnarm: Yeah this is a kludge; the spawn command fails though, I didn't look into it properly (hard slog, had figured the bug out, got lazy with the loose end), I'll take a another look and see if I can figure it out.

silnarm: Didn't figure out exactly what was happening here, but the teleport() indirectly called in createFake() is not playing nice, perhaps because things aren't entirely constructed and we start using the instance as if it were ('handling' packets for it).
Also suspect getting rid of it may mean special casing somewhere else.

altrisi:
Thanks for contributing!

PR #1670, opened 2023-03-08
Add arithmetic shift, use logic shift in current right shift

### LINKED ISSUE #1666: Bit shift operations work incorrectly
`bitwise_shift_left(num, amount)` and `bitwise_shift_right(num, amount)` returns `num << (amount & 63)` or `num >> (amount & 63)` when `amount` is negative or greater than 63.
```
bitwise_shift_left(1, 64) => 1, should be 0
bitwise_shift_right(1, 64) => 1, should be 0
bitwise_shift_left(1, -1) => -9223372036854775808, should be 0
bitwise_shift_right(1, -1) => 0, shuould be 2
```
maybe `bitwise_shift_right` should be implement with `>>>` instead of `>>`.
`bitwise_roll_left` and `bitwise_roll_right` can be implement with `Long.rotateLeft` and `Long.rotateRight`.

altrisi:
Partially resolves #1666.

This PR changes the current `bitwise_shift_right` to do a logic shift instead of arithmetic, and adds `bitwise_arithmetic_shift_right`.

RFC for whether it should be the other way around, logic being the new one added, but I personally think default should be logic.

Pending is either updating the docs to reflect that only the least significant bits of the amount is used (as specified by the JLS and therefore how the operator works), or if we should work around it and allow amounts greater than that by wrapping the operator (passing a number greater than those is useless as it'd just return 0, or 1 if arithmetic and with the MSB being 1).

### DIFF
--- a/docs/scarpet/language/Operators.md
+++ b/docs/scarpet/language/Operators.md
@@ -310,6 +310,9 @@ only take integer values, so if the input has a decimal part, it will be discard
 	than 63 positions will result in a 0 (cos you shift out all the bits of the number)
  - `bitwise_shift_right(num, amount)` -> Shifts all the bits of the first number `amount` spots to the right. Like with the above,
 	shifting more than 63 bits results in a 0.
+ - `bitwise_arithmetic_shift_right(num, amount)` -> Shifts all the bits of the first number `amount` spots to the right arithmetically.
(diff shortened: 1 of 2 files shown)


gnembon:
I agree and support. I think a mention about lshr would be nice, similarly to what you put in arithmetic.

Approve and feel free to merge as you please.

gnembon (approved):

PR #1678, opened 2023-03-16
Add `player <name> use perTick <amount>`

Crec0:


gnembon:
Please explain how this would be useful? 

Also - seems like too many unrelated changes to what you want to have done.

Crec0:
Reason for the PR is because this has been a frequent question/request on the carpet discord. This is a carpet bot/server sided alternative to tweakeroo's fastBlockPlacement and useful for testing things like concrete converters, block converters and other contraptions which require block placement faster than `interval 1`. 
Unsure about unrelated changes, can you please elaborate? Tick function is same behavior, just adjusted so the return value is accounting for the loop

PR #1681, opened 2023-03-19
Revert "Fixing `/player botname stop` with relation to fake player movement"

altrisi:
Reverts gnembon/fabric-carpet#1584. Opens #1072.

Closes #1680.

PR #1683, opened 2023-03-19
Fixing `/player botname stop` with relation to *fake* player movement

### LINKED ISSUE #1680: Can't move minecarts using movement keys
Nothing happens when hold W in a minecart. It should start moving but it doesn't. All carpet rules are disabled.

Steps to reproduce:

1. Join singleplayer world with carpet mod installed
2. Place rails and install minecart on it
3. Sit in minecart and hold W

Ghoulboy78:
As opposed to real player movement.
Closes #1680.
Supersedes and therefore closes #1681.

### DIFF
--- a/src/main/java/carpet/helpers/EntityPlayerActionPack.java
+++ b/src/main/java/carpet/helpers/EntityPlayerActionPack.java
@@ -6,6 +6,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 
+import carpet.patches.EntityPlayerMPFake;
 import carpet.script.utils.Tracer;
 import net.minecraft.commands.arguments.EntityAnchorArgument;
 import net.minecraft.core.BlockPos;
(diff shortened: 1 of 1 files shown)


src/main/java/carpet/helpers/EntityPlayerActionPack.java
```diff
@@ -238,8 +239,12 @@ public void onUpdate()
             }
         }
         float vel = sneaking?0.3F:1.0F;
-        player.zza = forward*vel;
-        player.xxa = strafing*vel;
+        if (forward != 0.0F || player instanceof EntityPlayerMPFake) {
```
altrisi, line 242:
```suggestion
        // The != 0.0F check are needed given else real players can't control minecarts, however it works with fakes and else they don't stop immediately
        if (forward != 0.0F || player instanceof EntityPlayerMPFake) {
```

altrisi (approved):
Works, no regressions

PR #1686, opened 2023-03-25
shadow-player-reconnect-fix

### LINKED ISSUE #1468: Advancements not working after logging in from /player shadow
## The Issue

Logging in after an AFK session using `/player _____ shadow` fails to load advancements for the player.
It also means that advancements cannot be earned (eg. entering a nether fortress for the first time).

The issue only lasts until a relog.

image

## Version

Minecraft 1.19 with Fabric 0.14.6
fabric-carpet-1.19-1.4.79+v220607

This issue was also present in 1.18.2 versions and potentially before that

silnarm:
If a bot is `kill()`ed for a duplicate login, handle the disconnect immediately, Fixes #1468

Fixes player advancements and statistics being unloaded after the player logs back in.

### DIFF
--- a/src/main/java/carpet/patches/EntityPlayerMPFake.java
+++ b/src/main/java/carpet/patches/EntityPlayerMPFake.java
@@ -5,6 +5,7 @@ import com.mojang.authlib.GameProfile;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.UUIDUtil;
 import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.contents.TranslatableContents;
 import net.minecraft.network.protocol.PacketFlow;
 import net.minecraft.network.protocol.game.ClientboundPlayerInfoUpdatePacket;
 import net.minecraft.network.protocol.game.ClientboundRotateHeadPacket;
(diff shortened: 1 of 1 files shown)


altrisi:
Fix makes sense, now wondering if there's a reason it's usually scheduled (until next tick I'd assume?) to kill, or if would make sense to make it always immediate.

silnarm:
Yes, my first thought was is this delay really needed, and it _may_ not be, but it makes sense to do so as bots are disconnected on death which can happen anywhere, while with normal players the disconnects happen at a particular time in the tick. 
Minecraft in general is pretty sloppy and leaves dangling references in situations like this anyway, so I don't actually think it would matter to handle the disconnect immediately; but there is probably some edge case that would bite us, so I went with caution.

gnembon:
The reason for the delay was that in the past that was causing some hard crashes that I didn't care to investigate too deep. There is a chance it will not happen with specific death causes like duplicate login, there is a chance other death causes are fine too, and there is a chance that entity handling has improved to the point its not needed at all. Never had a chance to investigate that.

gnembon (approved):

L-Lawlite:
Can this be back ported to 1.19.2?

L-Lawlite:
> 

There should not be any code change so probably u can paste this directly.

PR #1687, opened 2023-03-25
shadow-single-player-owner-fix:

### LINKED ISSUE #1442: problem of command shadow(1.16.5)
It's known that `/player [player's name] shadow` doesn't work in single player mode and the single player server will stop working if you try to make yourself shadow. I tried to use this command in my single player save and I found that the fake player seems to join the game not with the latest playerdata but the playerdata saved last time. When I opened the save again, I found that the playerdata was not saved but the world was updated. For example, I can duplicate any item by following the process below:

1. get the item you want to duplicate in your inventory
2. save and quit to title
3. open the save again
4. drop the item you want to duplicate on the ground
5. type'/player [your name] shadow' and enter
6. you will see 'you are logged in from another location', click 'back to server list' and then back to title screen
7. open the save again, you'll see the item you want to duplicate both in your inventory and on the ground

silnarm:
* Disallow shadowing a single player server owner. Fixes #1442

### DIFF
--- a/src/main/java/carpet/commands/PlayerCommand.java
+++ b/src/main/java/carpet/commands/PlayerCommand.java
@@ -35,7 +35,6 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.LinkedHashSet;
 import java.util.List;
-import java.util.Objects;
 import java.util.Set;
 import java.util.function.Consumer;
 
@@ -366,8 +365,7 @@ public class PlayerCommand
             Messenger.m(context.getSource(), "r Cannot shadow fake players");
             return 0;
         }
-        MinecraftServer server = Objects.requireNonNull(player.getServer());
-        if (server.isSingleplayer() && server.isSingleplayerOwner(player.getGameProfile())) {
+        if (player.getServer().isSingleplayerOwner(player.getGameProfile())) {
             Messenger.m(context.getSource(), "r Cannot shadow single-player server owner");
             return 0;
         }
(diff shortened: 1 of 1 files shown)


src/main/java/carpet/commands/PlayerCommand.java
```diff
@@ -365,6 +366,11 @@ private static int shadow(CommandContext<CommandSourceStack> context)
             Messenger.m(context.getSource(), "r Cannot shadow fake players");
             return 0;
         }
+        MinecraftServer server = Objects.requireNonNull(player.getServer());
```
altrisi, line 369: Don't think the server can ever be null for a ServerPlayer.

src/main/java/carpet/commands/PlayerCommand.java
```diff
@@ -365,6 +366,11 @@ private static int shadow(CommandContext<CommandSourceStack> context)
...
+        if (server.isSingleplayer() && server.isSingleplayerOwner(player.getGameProfile())) {
```
altrisi, line 370: No need to check `isSinglePlayer`, dedicated server will always return `false` for `isSinglePlayerOwner`.

altrisi (commented):
Some tiny comments

src/main/java/carpet/commands/PlayerCommand.java
silnarm: Yeah, I've even seen this in action plenty already I think. Too trusting of what IntelliJ is telling me, time to disable some inspections.

altrisi (approved):

altrisi:
I believe this crashes if the given player isn't online (`getPlayer` returns `null`), will fix it in a small refactor I'm doing to the player command.

silnarm:
Oops, missed a test case there I guess.

PR #1689, opened 2023-04-01
Fix `/player` `look` and `turn` commands reversing yaw and pitch

### LINKED ISSUE #1688: Yaw/pitch applying is reversed in player command
Currently we pass `rot.x`, `rot.y` to a method that applies it in the order `setY()`, `setX()`.

This makes it so it's reversed vs those in the F3 hud, and I'd assume with other vanilla commands that use the rotation argument.

Originally reported on Discord: https://discord.com/channels/882822986795716608/1091533813399683083

altrisi:
Fixes #1688.

Also uses a switch expression in `EntityPlayerActionPack::look(Direction direction)`.


Fallen-Breath:
This is a breaking changes. It breaks existing contraptions using the `/player look` command to control player

I'll suggest to keep the existing behavior for `/player` `look` / `turn` at least for a while, and notice people for this change. It does be a bug, but it keeps a consistently behavior since even the very beginning of fabric-carpet (50d1d74579917ea3d3d5af802185fdd7e6ef9622), and there might be quite a lot of stuffs that depend on the wrong behavior

Please give people enough time to adjust, instead of simply find-and-fix

altrisi:
How do you give the notice though? We can tell players when they use it that it's going to change, but we're just saying "it's going to", they can't do anything to remove the warning and they'll need to keep using the old (also broken) way and keep being used to it, until one day this exact same thing happens: it stops working.

I doubt this'll be in stable until 1.20, I see no point in delaying the change. Wouldn't be against a warning to uses of the command mentioning it has changed though given I do see the point in this breaking some stuff.

Fallen-Breath:
That's why it's hard to changing a existing command syntax. imo as long as users are informed, it is a success.

For best compatibility you can make a new subcommand to use the correct order, though that might be a bit strange

A version bump for major / minor version might be good for the breaking change, but users might just ignore anything and just download-and-use the mod

altrisi:
Yeah, it's a difficult thing to do, but IMO it being consistent with the rest of the game is important. I don't think player look and turn is super used either, and also Carpet has had more breaking changes in other minor versions.

Fallen-Breath:
Being consistent with the game is nice, but if you release something inconsistent and keep it untouched for years, it's better to not touch that, and give users enough time to adjust if changes will happen

> I don't think player look and turn is super used either

Example: Initializing player position in farms, in command blocks in wdl. This is a very common usage with this command. Changing the syntax will make people complaining the farm being broken

> Carpet has had more breaking changes in other minor versions

Examples? And I don't think this is a good excuse

altrisi:
I know it is used, just saying it's not super used vs other commands and features.
Once again, changing it later after a period of warning will do nothing, just let time to create more world downloads with the old, at some point broken commands, given if you're going to make a command to setup a player you're going to use the one that's working at the time, even if it's giving you a warning, because the future version doesn't work.

> if you release something inconsistent and keep it untouched for years

I've never seen it reported before, else I doubt it'd have been untouched for years. I'd also think it wasn't inconsistent intentionally, though I can't know.

altrisi:
Going to revert this.

Fallen-Breath:
> I know it is used, just saying it's not super used vs other commands and features.

It's used, and used commonly, that's enough. You don't to compare it with other commands and features, cuz you are not changing the behavior of other stuffs 

I know it a notice cannot help anything and cannot stop the existing usage from being broken, but once again, as long as users are informed, users get awared, it is a success, other than just make the change and break stuffs without noticing

Also, the current behavior is introduced in 2019 (metioned in https://github.com/gnembon/fabric-carpet/pull/1689#issuecomment-1493215972), which is 4 years ago from now

PR #1696, opened 2023-04-08
Add utility method to easily run code on all `SettingsManager`s

altrisi:
It's a relatively common pattern in Carpet, and you have to loop on all extensions, get SM, check not null, then run your code. And then do it with ours too.

This adds a method that runs an action for all managers, doing all the null checking and running it for carpet too.

~Other usage of this pattern is for `attachServer`, but haven't changed it here given it also calls an extension handler, though tbh it could be changed.~

WIP because I'm not convinced on putting it in CarpetServer, but I don't know where to put it (and don't want to expose it).


gnembon:
good idea

PR #1702, opened 2023-04-15
Replace deprecated `world_time()` with `system_info('world_time')`

Ghoulboy78:


altrisi (approved):
Nice find! I feel like it should've been caught way earlier though, we need something to find these at least in the built-in apps...

PR #1709, opened 2023-04-22
Use `DyeColor`'s `textColor` for hopper counters

altrisi:
Uses `DyeColor`'s own `textColor` for creating the colored name for hopper counters.

RFC: These are _slightly_ different.

Though they still look good, are slightly more consistent with similar minecraft colors and lets use get rid of a 16 entry manual map.

Not tested at all, ported from 1.19.4 in my IDE to current master using the Github web editor.


gnembon:
reason before was that original Messenger didn't support hex colors. That was added way later. Now there is plenty of palletes to choose from.

gnembon (approved):

PR #1715, opened 2023-05-01
fix long values being converted to doubles when negated

manyrandomthings:
(also fixes a game crash caused by #1709)

This affects 1.19.4 as well, so a new release for that version should probably be made that includes this fix. but there doesn't seem to be a 1.19.4 branch at the moment

Example: 
image

### DIFF
--- a/src/main/java/carpet/helpers/HopperCounter.java
+++ b/src/main/java/carpet/helpers/HopperCounter.java
@@ -102,7 +102,7 @@ public class HopperCounter
         this.color = color;
         String hexColor = Integer.toHexString(color.getTextColor());
         if (hexColor.length() < 6)
-            hexColor = "0".repeat(hexColor.length() - 4) + hexColor;
+            hexColor = "0".repeat(6 - hexColor.length()) + hexColor;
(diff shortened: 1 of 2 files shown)


src/main/java/carpet/helpers/HopperCounter.java
```diff
@@ -102,7 +102,7 @@ private HopperCounter(DyeColor color)
         this.color = color;
         String hexColor = Integer.toHexString(color.getTextColor());
         if (hexColor.length() < 6)
-            hexColor = "0".repeat(hexColor.length() - 4) + hexColor;
+            hexColor = "0".repeat(6 - hexColor.length()) + hexColor;
```
altrisi, line 105: Idk how did this happen, thanks for fixing it!

altrisi (approved):

altrisi:
Introduced in 8f6127160666d6c2688a145a8ffdaed3f0cca65b. I'll check if there's more similar ones that happened from that.

altrisi:
Going to put this change into a PR with all the fixes.

altrisi:
Closing in favour of #1719 that has the rest of operations covered, but thanks for and reporting it!

PR #1722, opened 2023-05-07
Improve Modrinth download badge

triphora:
Just replaces the weird XML thing with the proper `logo=modrinth` :)


altrisi:
Didn't know it existed, that's definitely nicer in the code, thanks!

Wondering, is there a similar one for the lighter/less simple one? I feel like there's not a lot of contrast with this one (though that may just be I'm not used to seeing it).

triphora:
You can always specify logoColor like so:
https://img.shields.io/modrinth/dt/carpet?label=Modrinth%20downloads&logo=modrinth&logoColor=blue
Bluerinth
I can change the logo color to the proper `#1bd96a` of the dark mode brand color, if you'd like. It uses the light mode brand color by default.

PR #1724, opened 2023-05-11
Fake player portal fix

### LINKED ISSUE #1712: Server crashed when carpet bot goes through portal
Players on my server experienced an issue where they tried to transport a carpet bot through a nether portal, and it ended up crashing the server. Here is the server log and crash log:

https://mclo.gs/1d3vNTc
https://mclo.gs/vxe72vK

silnarm:
Fixes #1712 by re-implementing the fix for #1190.

The new fix just moves the bot on the chunkmap directly, rather than handling pretend packets. This avoids adding another special case to not do the `handleMovePlayer()`, although we still need to know if the bot is freshly spawned (because bots are teleported before they are added to the chunkmap). The `hasSpawned` got removed anyway, as I found a way to determine this without us having to track it ourselves.


silnarm:
Have some automated tests now, #1712 is a 'sometimes' bug, though a given portal will give consistent behaviour (be fine, or crash the game). When I first got to running them on 1.19.4 the nether portal test passed, but another portal added a little away consistently crashes, so I added a second nether portal test.

<details><summary>1.4.96 (MC 1.19.3):</summary>

image
</details>

<details><summary>1.4.101 (MC 1.19.4):</summary>

image
</details>

<details><summary>#fake-player-portal-fix (MC 23w17a):</summary>

image
</details>

altrisi (approved):
Seems to be working fine, tested both portal types just in case + some actions after using one, thanks again!

PR #1727, opened 2023-05-13
Add modrinth link to Discord release message

LostLuma:
Since the [tag names](https://github.com/gnembon/fabric-carpet/tags) for carpet match the exact URL part needed to create a modrinth version URL I thought this may be useful for those of us preferring to download from there.


altrisi:
This is planned, but I want to do it when I automate releasing too so they always are the correct ones. I just haven't gotten around to it yet, I'm currently uploading them manually.

PR #1740, opened 2023-06-07
add cancel for `__on_player_command`

shenjackyuanjie:


replaceitem:
Do not edit the Full.md file, since that gets automatically generated from the other doc files when they are updated.

src/main/java/carpet/mixins/ServerGamePacketListenerImpl_scarpetEventsMixin.java
```diff
@@ -291,7 +291,7 @@ private void onChatMessage(ServerboundChatPacket serverboundChatPacket, Callback
     private void onChatCommandMessage(ServerboundChatCommandPacket serverboundChatCommandPacket, CallbackInfo ci) {
```
Harveykang, line 291: Forgot to add "cancellable = true" to the annotation

PR #1741, opened 2023-06-15
Create fr_fr.json

Calvineries:


altrisi:
Thanks a lot for this!

If you want to give a try to translating the rules, you can find them here: [altrisi/fabric-carpet@migrate-rules/src/main/resources/assets/carpet/lang/en_us.json](https://github.com/altrisi/fabric-carpet/blob/migrate-rules/src/main/resources/assets/carpet/lang/en_us.json) (they go directly in the same file)

src/main/resources/assets/carpet/lang/fr_fr.json
```diff
@@ -0,0 +1,36 @@
...
+  "carpet.category.bugfix": "correction de bugs",
+  "carpet.category.survival": "survie",
+  "carpet.category.creative": "créatif",
+  "carpet.category.experimental": "expérimental",
+  "carpet.category.optimization": "optimisation",
+  "carpet.category.feature": "fonctionnalité",
+  "carpet.category.command": "commande",
+  "carpet.category.tnt": "TNT",
+  "carpet.category.dispenser": "distributeur",
+  "carpet.category.scarpet": "tapis",
```
Ghoulboy78, line 15: This should probably just be `scarpet`, since it's referring to the programming language itself, not to a carpet.

Ghoulboy78:
Otherwise, looks good!

altrisi:
Thanks for contributing! Right now the translation system is not the best so (once it's released) you'll have to set the language with `/carpet language`, but it should have French thanks to this!

PR #1753, opened 2023-06-30
Fix MC-88967: Mobs when converting don't keep NBT tags

Ghoulboy78:
Fix for MC-88967. Not a major bug, but not a simple fix either, and one which will likely take a while to get into the game.


altrisi:
Looks like a lot of stuff and complexity for this, and also if mojang ever adds a new field the rule will stop working properly (and also this may not work if mods do that too). Wondering if there's a better way, though likely not.

Ghoulboy78:
Ye, I know, but as you said, I can't think of a better solution.

PR #1755, opened 2023-07-04
Reset fake player's delta movement when spawning

### LINKED ISSUE #1058: Fake Player knockback on death is applied at respawn
As of Carpet mod 1.4.45, when you kill a fake player using an attack that deals knockback, the knockback is applied when the fake player respawns. This causes `/player ... spawn at x y z` to end up positioning the fake player 2-6 blocks away, depending on the knockback dealt.

I would suggest that either:
* The fake player stays online for 20 ticks after death (same as real players), during which the knockback is applied.
* `/player ... spawn at x y z` should cancel the fake player's momentum on respawn.

This is caused by the Motion tag saved into the fake player's data being the motion vector with the knockback applied instead of without, for example [0.7, 0.4, 0.0] with knockback, [0.0, ~-0.1, 0.0] without knockback.

Modlist (MC 1.17.1):
```
Mods:
  [✔️] advancementinfo-1.17.1-fabric0.36.1-1.2.1
  [✔️] carpet-extra-1.17.1-1.4.43
  [✔️] fabric-api-0.37.2+1.17
  [✔️] fabric-carpet-1.17.1-1.4.45+v210811
  [✔️] itemscroller-fabric-1.17.1-0.15.0-dev.20210707.005506
  [✔️] litematica-fabric-1.17.1-0.0.0-dev.20210713.103711
  [✔️] malilib-fabric-1.17.1-0.10.0-dev.24
  [✔️] minihud-fabric-1.17.1-0.19.0-dev.20210707.150016
  [✔️] notenoughcrashes-3.4.0+1.17-fabric
  [✔️] tweakeroo-fabric-1.17.1-0.10.0-dev.20210710.155746
```

filips123:
With this PR, fake player's delta movement (momentum) is reset when the player is spawned. This is done to prevent knockback from the previous death from being applied. This should fix #1058.


Ghoulboy78 (approved):
This is good.

altrisi:
Is this the right solution though? Likely should instead fix them keeping the speed when being fully killed. Though this fix also makes sense.

filips123:
Well, it's probably the easiest way to fix this...

Ghoulboy78:
> Is this the right solution though? Likely should instead fix them keeping the speed when being fully killed.

I'm not so sure, because spawning fake players is typically done either in creative for testing, or in survival for farms.
In survival, you want to spawn the bot in the afk location and not have it move around due to residual knockback, and in creative, you might be trying to align the bot in a specific location, in which case having it move around is also a bad idea.

altrisi:
Actually yeah it makes sense that when you are spawning a fake player you always want it to stay there, although if it was a real player that had actual movement that spawning them would reset it for when they log back in. Though there's a lot more like that, so I don't think it should be the thing to care about here.

Ghoulboy78:
> so I don't think it should be the thing to care about here.

Wdym by that?

filips123:
Can this be merged please? Even if it is not the most correct solution, it should work and is quite simple.

PR #1756, opened 2023-07-04
Add option to configure listing fake players on the multiplayer screen

### LINKED ISSUE #1749: Players spawned with /player are now labeled as 'Anonymous Player' in server list.
In the most recent editions of carpet, when players spawn using /player <name> spawn and /player <name> shadow, they are consistently shown as 'Anonymous Player' in the server list.

player spawned from /player command

Before, these players would be listed with their respective usernames. I have been unable to locate any documentation regarding this alteration or the reason behind it.

filips123:
This adds a new Carpet option (`allowListingFakePlayers`) that configures whether fake/bot players are displayed on a player list on the multiplayer screen. When this option is disabled, all fake players will be listed as "Anonymous Player". This fixes #1749.

Previously, this option was always enabled, so fake players were also always displayed on the list. However, recently (I think in `1.20-pre1`), Minecraft changed the internal option to be `false` by default, so fake players are now listed as anonymous. With this PR, the behaviour can now be configured by users.

~~I set `true` as default because this was the old behaviour (until recently). However, I can change it to `false` by default if needed.~~ Now it's `false` by default.

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -597,6 +597,9 @@ public class CarpetSettings
     @Rule(desc = "Spawn offline players in online mode if online-mode player with specified name does not exist", category = COMMAND)
     public static boolean allowSpawningOfflinePlayers = true;
 
+    @Rule(desc = "Allows listing fake players on the multiplayer screen", category = COMMAND)
+    public static boolean allowListingFakePlayers = true;
+
     @Rule(desc = "Allows to track mobs AI via /track command", category = COMMAND)
     public static String commandTrackAI = "ops";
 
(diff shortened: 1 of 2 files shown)


Ghoulboy78:
As altrisi mentioned in the issue, it might make more sense to set default as `false`, since it is a fake player, and not really on the server. Also because that is current behaviour.

altrisi:
I don't know why this should be configurable. What's the use for it? If you're trying to check if a fake player is logged in there's likely better ways to do it. We have so many rules already, and to me this seems like one that won't be used much.

Code looks good though.

filips123:
What are those better ways of checking for fake players?

This option can be good if you want to quickly check fake players without the need to log in, which I think is useful. Also, it's to preserve the old behaviour.

altrisi (approved):
Makes sense, thought you originally meant for some program to read it in which case I was thinking of having something else server-side.

PR #1764, opened 2023-07-16
Improve docs for `create_explosion()`

Ghoulboy78:


altrisi:
Probably would make sense to instead of making the default parameters the important part, making the descriptions be, then mention their defaults. What do you think?

altrisi (approved):

Issue #1546, opened 2022-09-17
Piglin Brutes spawn in air when piglinsSpawningInBastions is turned on


Ace-tyl:
After cleaned the bastion to make a piglin brute farm (it can produce ancient debris in my server), I found the piglin brutes can spawn in air and start falling down immediatly after they are spawned.

James103:
What Carpet mod version is the server running?

Ace-tyl:
the newest version 1.4.83 for Minecraft 1.19.1

Ace-tyl:
> What Carpet mod version is the server running?

The newest version 1.4.83 in Minecraft 1.19.1

aria1th:
This is caused because SpawnPlacements does not have Piglin Brutes registry.

PR #1769, opened 2023-07-24
Fix piglin brutes spawning in air

aria1th:
Fixes https://github.com/gnembon/fabric-carpet/issues/1546

The core cause was Piglin Brutes not having spawn condition, allowing it to pass every check except 'damage'.

This patch registers spawn condition check for piglin brutes.

This does not mean they will spawn in overworld / elsewhere, since spawn requires mob list entry first.

### DIFF
--- a/src/main/java/carpet/mixins/PiglinBrute_spawnPlacementsMixin.java
+++ b/src/main/java/carpet/mixins/PiglinBrute_spawnPlacementsMixin.java
@@ -1,0 +1,44 @@
+package carpet.mixins;
+
+import net.minecraft.core.BlockPos;
+import net.minecraft.util.RandomSource;
+import net.minecraft.world.entity.EntityType;
+import net.minecraft.world.entity.Mob;
+import net.minecraft.world.entity.MobSpawnType;
+import net.minecraft.world.entity.SpawnPlacements;
+import net.minecraft.world.entity.monster.piglin.Piglin;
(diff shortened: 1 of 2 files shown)


altrisi:
No rule checks there, does this affect vanilla behaviour? Or does this only run with the rule on?

Also throw something (like `new AssertionError`) on static shadows please.

aria1th:
@altrisi for normal cases, it should not be affected, but there might be some cases that could be inconsistent 

1. If some mods are already trying to make piglin brutes naturally spawnable then 😨 
2. Initial Structure mob spawns could be affected by this. (if it uses same SpawnPlacements, it means it was already spawning in air which can make newly generated piglin brutes far less than before).

I'll do some check in 24 hours.

aria1th:
Some inspections:
to NOT register static value and prevent some exceptions, instead of current code, the code should be modified as:

inject getHeightmapType, checkSpawnRules function instead.

aria1th:
This was the minimal change required to prevent piglin brutes spawning in air - but it may prevent some new chunk piglin brute generation... but rule itself is already changing the behavior so it shouldn't matter, right?
@altrisi

altrisi:
As long as it only affects behaviour when the rule is on, and with it on it doesn't like severely like reduce base structure spawns (changes significantly "how it feels" (?) going normally through a bastion) or something it's fine for me.

altrisi (approved):

PR #1770, opened 2023-07-26
Extract all logic in `modifiedRunLoop` to small injections

### LINKED ISSUE #1747: Refactor `modifiedRunLoop` mixin to not wrap vanilla logic
It's incredibly common for people see `carpet$modifiedRunLoop` in stack traces and blame Carpet. Some of these get into issues here, some in the discord, some in other discords (like Fabric's), where they'll be told Carpet looks to be the cause and should be removed.

That piece of code should be refactored so it isn't a block wrapping vanilla, but specific changes where it's needed, possibly also helping compatibility and preventing unwilling behaviour changes if that code changes and it gets unnoticed.

altrisi:
Reviewing wanted! I'm not good at all with mixins, and although this has taken me a while and I've tried to check and the output is similar, it may not be the same! And this is quite a core part of Carpet (and vanilla!) that I'd prefer to not have messed up.

This PR gets rid of `modifiedRunLoop` and instead separates it into smaller mixins to the specific places where Carpet was changing its behaviour. This has the following benefits:
- Removes the possibility for it to get out of sync with vanilla (it actually already was, though only vs a JFR profiling event) and having to check for that
- Improves mod compatibility with mods that change other stuff in the run loop. Currently any mod that injected any logic there would just be silently skipped, IMO the worst of compatibility issues
- Removes it from logs in a few places where we wrapped vanilla logic reducing reports and confusion with people thinking the crashes were caused by Carpet
- Removes the workaround of `TimeProfiler` being private that meant we had to copy all the logic that used it with our own type

As a compromise, the local variable `msThisTick` is now global, but shouldn't be an issue given that should always accessed by the same thread and in the same section.

Missing (only ideally, likely not really needed for it to just work):
- Replacing `l = Util.getMillis() - this.nextTickTime;` with our `l` in the `preTick` (currently lost)

Resolves #1747.


altrisi:
Currently crashes with g4mespeed.

altrisi:
It got extracted further it seems!

PR #1771, opened 2023-07-26
Remove `lightEngineMaxBatchSize` rule

### LINKED ISSUE #1711: `lightEngineMaxBatchSize` is disabled since 23w17a
Tracking issue.

The actual call to set it was removed in 07033af7dd95efffffbe3af53e0fe61561704fc9.

Wondering if it still has uses with the new engine, though I've never known what it's used for exactly so I can't really judge.

altrisi:
Resolves #1711.

The rule has been soft-disabled (did nothing) since the new light engine, and the main reason this was added from what I can see is performance, that is better in the new engine.

Not only that, the new default batch size is `1000`, and the main values this rule used to be set were high values, making it seem more pointless.

This PR removes the rule. However, it wouldn't be difficult to make it usable with the new engine if still useful to someone.

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -863,43 +863,6 @@ public class CarpetSettings
     )
     public static int spawnChunksSize = MinecraftServer.START_CHUNK_RADIUS;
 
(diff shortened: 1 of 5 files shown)


Andrews54757:
yes it can be removed

PR #1774, opened 2023-07-26
Fix registry lists being randomized

manyrandomthings:
In 1.18.1 and earlier `item_list()` and `block_list()` would return the list of all items/blocks in registry order, but in 1.18.2 it started to be randomized due to getting the values from the keys of a HashMap. This fix keeps them consistent in the order that they are in the item/block registry.

### DIFF
--- a/src/main/java/carpet/script/api/Auxiliary.java
+++ b/src/main/java/carpet/script/api/Auxiliary.java
@@ -133,7 +133,7 @@ public class Auxiliary
             CarpetContext cc = (CarpetContext) c;
             if (lv.isEmpty())
             {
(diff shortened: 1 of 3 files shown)


manyrandomthings:
Being in registry order is pretty nice for both consistency and historic reasons I guess. It's just a fairly standard way to list all the items/blocks in minecraft, and it's much nicer than being random or alphabetical.
I suppose this could also have a use by being able to find the item/block id by querying the index of a value even though numeric ids aren't really used anymore.

altrisi (approved):
Makes sense! At some point we should maybe specify some ordering, given registry order doesn't really help much and could change at any point.

PR #1775, opened 2023-07-27
Refactor PistionStructureResolver API for better Compatiblity

Dragon-Seeker:
The point of this PR is to make the changes to `PistionStructureResolver` opt-in to prevent entirely hijacking the `PistionStructureResolver` disallowing any other modders from creating custom interactions not thru carpet API. It also defaults to using vanilla code when a given block is not found to be of `BlockPistonBehaviourInterface`

### Changes:
- Switch to a opt in system for pistion interaction methods by moving to a dedicated interface and removing mixin that injects such into every block
- Default to vanilla code for slime blocks and honey blocks behavior.
- Keep `BlockBehaviourInterface` and `BlockBehaviour_customStickyMixin` for compatibility reasons
- Minor code refactors


src/main/java/carpet/fakes/BlockBehaviourInterface.java
altrisi, line 1: Can be removed if no longer used, `fakes` are not really public api but stuff to work with mixins (mostly accessors but some other stuff too).

Note that it's still used in `BlockBehaviour_customStickyMixin` (was the base so everything implemented it by default).

src/main/java/carpet/mixins/ChestBlock_customStickyMixin.java
altrisi, line 1: Unnecessary style changes in this file, only the iface change should be there.

src/main/java/carpet/mixins/PistonStructureResolver_customStickyMixin.java
```diff
@@ -89,6 +96,8 @@ private void captureNeighborPositions(BlockPos pos, CallbackInfoReturnable<Boole
         )
     )
     private boolean onAddBranchingBlocksCanStickToEachOther(BlockState neighborState, BlockState state, BlockPos pos) {
-        return ((BlockBehaviourInterface)state.getBlock()).isStickyToNeighbor(level, pos, state, neighborPos_addBranchingBlocks, neighborState, dir_addBranchingBlocks, pushDirection);
+        if(!(state.getBlock() instanceof BlockPistonBehaviourInterface behaviourInterface)) return canStickToEachOther(neighborState, state);
```
altrisi, line 99:
```suggestion
        if (!(state.getBlock() instanceof BlockPistonBehaviourInterface behaviourInterface)) return canStickToEachOther(neighborState, state);
```

Minor thing. Same in two other places in this file.

Also wondering if it'd be more readable with the opposite condition, though not blocking the PR with that.

src/main/java/carpet/fakes/BlockPistonBehaviourInterface.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Opt-in Interface that allows for more control on a blocks interaction within the {@link PistonStructureResolver} via {@link PistonStructureResolver_customStickyMixin}
+ */
+public interface BlockPistonBehaviourInterface {
+
+    /**
+     * @return whether this block is sticky in any way when moved by pistons
+     */
+    default boolean isSticky(BlockState state){
```
altrisi, line 18:
```suggestion
    default boolean isSticky(BlockState state) {
```
Ik it was like this in the old iface, but good opportunity to fix it given you're changing it.

altrisi (commented):
Makes sense! Also tested and working fine. Some comments.

src/main/java/carpet/mixins/ChestBlock_customStickyMixin.java
Dragon-Seeker: Reverted that changes of the Code Style

src/main/java/carpet/fakes/BlockBehaviourInterface.java
Dragon-Seeker: Got it, did not know if such was used as a public API of sorts so left it there until told otherwise

src/main/java/carpet/mixins/PistonStructureResolver_customStickyMixin.java
Dragon-Seeker: Did as you asked and inverted the logic for the method

src/main/java/carpet/fakes/BlockPistonBehaviourInterface.java
Dragon-Seeker: Done and did such for the bottom method also

altrisi (approved):

PR #1778, opened 2023-07-28
Fix chat signatures desyncing when cancelling `__on_player_message`

### LINKED ISSUE #1777: Cancelling `__on_player_message` desyncs signatures, kicking the player on the next signed message/command
This is why you test stuff in prod.

Looking into it.

altrisi:
Attempt to do that.

At the very least, this doesn't kick the _same_ player if he is the only one chatting, but I haven't been able to test if _other_ players get them desynced (and therefore kicked) with this injection, and testing that is difficult because I only have one account to test.

Seems to work in an environment without signing enabled (tested vanilla client into offline server, not sure if that still signs, more testing may be needed).

Fixes #1777.


altrisi:
Tested and seems to work.

PR #1788, opened 2023-08-03
Fix typos in Auxiliary.md

Arian04:
I had to guess a little on what was intended for the second typo, but I think it's good.


Ghoulboy78:
I do love the jokes gnembon dots throughout the scarpet docs, but ye this looks good

PR #1789, opened 2023-08-03
Added 'item_cooldown' to query and modify item cooldowns

### LINKED ISSUE #597: Scarpet suggestion: Allow use of Scarpet API to set cooldown timers on items for players
Currently, there is no way to interact with the `player.getItemCooldownManager().set(item, ticks)` method via the Scarpet API. It would be great if there was, as that would allow Scarpet scripts to set cooldowns on items when they are used for custom abilities in similar ways that other Fabric mods such as [Identity](https://github.com/Draylar/identity) ([example use](https://github.com/Draylar/identity/blob/b85877d52da55c10ee55f9189f40fa382fe8341b/src/main/java/draylar/identity/api/ability/IdentityAbilities.java#L51)) do it.

Ghoulboy78:
Fixes #597
Simple fix, but querying item cooldowns only returns a percentage. I'd like to make it return the number of ticks, but there are private classes involved (specifically `net.minecraft.world.item.ItemCooldowns.CooldownInstance`), which idk how to mix into.

### DIFF
--- a/docs/scarpet/api/Entities.md
+++ b/docs/scarpet/api/Entities.md
@@ -302,6 +302,10 @@ Number of ticks remaining until an entity can use a portal again.
 
 Number of ticks an entity sits in a portal.
 
+### `query(e, 'item_cooldown', item)`
+
+Percentage remaining of item cooldown, e.g., from pearl throwing or horn blowing.
+
 ### `query(e, 'item')`
 
 The item triple (name, count, nbt) if its an item or item frame entity, `null` otherwise.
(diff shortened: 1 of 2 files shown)


altrisi:
To mixin you can use a string `target` for the class.

Ghoulboy78:
How so?

Ghoulboy78:
Wait, I might have found a solution

altrisi:
`@Mixin(target="net/minecraft/...")` iirc

altrisi:
If that's not enough then there's accesswidener

Ghoulboy78:
I've tried that and I've tried accesswidener, neither seems to be able to work. `net.minecraft.world.item.ItemCooldowns.CooldownInstance` is a private class, and idk how to get to it. But that is the only way to fix this issue.

altrisi:
Note that after accesswidener you have to regenerate/reimport/sync/refresh (or whatever it's called on intellij gradle) the minecraft sources/the project given it modifies directly the game jar.

Ghoulboy78:
Done!

PR #1791, opened 2023-08-04
Add food properties

### LINKED ISSUE #1701: Scarpet: Add functions to get a food item's hunger value and saturation modifier
Currently, there is no way I know of to get the hunger value of a food item, nor is there a way to query its saturation modifier.
A current workaround is to hardcode all hunger and saturation values of all edible foods, but that does not scale well with modded food items.

To fix this, I would like to suggest the following two functions (names subject to change):

### `item_hunger(item)`
Returns the hunger value of the given item, or `null` if it is not a food item.

### `item_saturation_mod(item)`
Returns the saturation modifier of the given item, or `null` if it is not a food item.

(The return value `null` can be substituted with `0` if necessary).

Ghoulboy78:
Fixes #1701 
I have added what was requested in the issue, but I realised that there is currently also no way to look at a food item's food properties, so I decided to add a system_info() style thing where you can either query it for a single property, or query all of them.
This way it's simple, and with no unnecessary overload.

### DIFF
--- a/src/main/java/carpet/script/api/Auxiliary.java
+++ b/src/main/java/carpet/script/api/Auxiliary.java
@@ -71,6 +71,8 @@ import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.EntityType;
 import net.minecraft.world.entity.EquipmentSlot;
 import net.minecraft.world.entity.decoration.ArmorStand;
+import net.minecraft.world.food.FoodProperties;
+import net.minecraft.world.item.Item;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.level.ChunkPos;
 import net.minecraft.world.level.Level;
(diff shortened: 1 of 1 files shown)


Ghoulboy78:
On an unrelated note, altrisi, do you think it would be a good idea to have pull request checks not run on drafts?

src/main/java/carpet/script/api/Auxiliary.java
```diff
@@ -1240,6 +1244,62 @@ else if (fdesc.type == FileArgument.Type.JSON)
...
         });
+
+        expression.addUnaryFunction("item_nutrition", v-> NumericValue.of(NBTSerializableValue.parseItem(v.getString(), CarpetServer.minecraft_server.registryAccess()).getItem().getFoodProperties().getNutrition()));
+        expression.addUnaryFunction("item_saturation", v-> NumericValue.of(NBTSerializableValue.parseItem(v.getString(), CarpetServer.minecraft_server.registryAccess()).getItem().getFoodProperties().getSaturationModifier()));
+
+        expression.addContextFunction("food_properties", -1, (c, t, lv)->{
+            if(lv.size()==0) throw new InternalExpressionException("'food_properties' requires at least an item name");
+            if(lv.size()>2) throw new InternalExpressionException("'food_properties' takes at most an item name and a property");
+
+            Item item = NBTSerializableValue.parseItem(lv.get(0).getString(), CarpetServer.minecraft_server.registryAccess()).getItem();
```
Ghoulboy78, line 1255: I didn't want to use `CarpetServer.minecraft_server.registryAccess()` here, but the function requires a `RegistryAccess` to be passed in.
Is there some better way?

altrisi: CarpetContext has it.

altrisi:
> On an unrelated note, altrisi, do you think it would be a good idea to have pull request checks not run on drafts?

Why not? They don't cause issues, and they still provide all the benefits of checks normally, and you can ignore them if you don't want them yet.

altrisi:
About the PR, wondering if it'd make sense to have a more general thing for item/block properties that you can then query for this stuff, though I don't know what else it'd have or how would be the best way to make handle it.

Ghoulboy78:
The thing is, that food properties are kinda unique.

I agree that maybe we could refactor the item/block properties thing to be easier, but tbh it's pretty understandable rn, and besides it would deprecate a bunch of apps which use these functions, especially in ways we don't expect.

Unless, of course, we were to add some new functions and deprecate the old ones, but that's separate from this cos, again, food is kinda unique. But that kind of a general querying thing could work for block properties and items both. That way if an item is inputted, then you get the default values for that item, but if a block value is inputted, then you get the properties of that block specifically, if that makes sense?
If it helps I could draft it, but tbh this sounds pretty similar to the system we already have.

Ghoulboy78:
> > On an unrelated note, altrisi, do you think it would be a good idea to have pull request checks not run on drafts?
> 
> Why not? They don't cause issues, and they still provide all the benefits of checks normally, and you can ignore them if you don't want them yet.

Well I figured cos that way you don't run unnecessary actions cos drafts can't be merged, but ig ye y not.

Ghoulboy78:
That should be it.

PR #1800, opened 2023-08-12
fix-fake-player-xp-not-resetting-on-death

### LINKED ISSUE #1797: Fake Players do not lose their XP when dying and still drop XP orbs
Just as the title implies, whenever a fake player dies while containing experience points, it will drop the experience (as expected) but will still have that experience when respawned via /player. To put it another way, fake players dupe XP when killed repeatedly.

https://www.youtube.com/watch?v=K0Dikci5CZk (example recording)

Keep inventory was turned off, and there weren't other datapacks/command-blocks running which could affect player experience levels. 
I am in minecraft version 1.20.1 using Carpet v1.4.112

silnarm:
* remove XP on bot death
* fixes #1797

### DIFF
--- a/src/main/java/carpet/patches/EntityPlayerMPFake.java
+++ b/src/main/java/carpet/patches/EntityPlayerMPFake.java
@@ -187,6 +187,7 @@ public class EntityPlayerMPFake extends ServerPlayer
         shakeOff();
         super.die(cause);
         setHealth(20);
+        giveExperienceLevels(-(experienceLevel + 1));
         this.foodData = new FoodData();
         kill(this.getCombatTracker().getDeathMessage());
     }


src/main/java/carpet/patches/EntityPlayerMPFake.java
```diff
@@ -187,6 +187,7 @@ public void die(DamageSource cause)
         shakeOff();
         super.die(cause);
         setHealth(20);
+        giveExperienceLevels(-(experienceLevel + 1));
```
James103, line 190: This line, which removes all levels and XP, should be wrapped in a check that `/gamerule keepInventory` is false.

PR #1817, opened 2023-09-22
Fix movableBlockEntities crashing in 1.20.2

manyrandomthings:
The `PistonBaseBlock#moveBlocks()` method was modified slightly in 1.20.2 causing the mixin target to shift slightly, which causes a crash when a piston moves any block.


altrisi (approved):
BEs now move without crashes

altrisi:
Thanks!

PR #1826, opened 2023-10-06
Fix hopper counters command error when guessing item color

### LINKED ISSUE #1825: /counter command seems to stop working if hopper gets certain type of items
So i tried the discord Player help forum made a post there, 0 response ig il try my luck here.

Hopper counters command **/counter** seems to stop working if the hoppers get a certain type of item into them, idk why but i have made a video of the issue

https://streamable.com/entvy1

i could not upload the video directly here it was too big in size, as you can see inputting bottles in the hopper will cause the /counter command to no longer work same with mob eggs, and prob other items too. i couldn't test every single one. 

For this test i have removed all other mods with only carpet in my mods folder even fabric api and my game version is 1.20.2 if any other information is needed i will glady provide :)

altrisi:
Fixes #1825 

I'd say warrants a 1.20.2 release.


altrisi:
This is also used by the `recipe_data` Scarpet function, so it also would cause an error there.

PR #1828, opened 2023-10-07
Fix #1081 and improve OptimizedExplosion readability

### LINKED ISSUE #1081: OptimizedTNT breaks TNT pushing other TNT
In vanilla, there is an exception for explosions pushing TNT, where the direction of the velocity vector is from the explosion to the TNT's feet position, as opposed to from the explosion to the TNT's eyes like every other entity.

In the vanilla code, there is an if statement to make this exception, which changes the direction of velocity:
`double aa = (entity instanceof TntEntity ? entity.getY() : entity.getEyeY()) - this.y;`
Here is the equivalent in carpet:
`double d7 = (entity instanceof TntEntity ? entity.getY() : entity.getEyeY()) - eAccess.getY();`
However, there is an extra, unnecessary exception which breaks tnt when they are both in the same place:
```
if (entity instanceof TntEntity &&
  entity.getX() == eAccess.getEntity().getX() &&
  entity.getY() == eAccess.getEntity().getY() &&
  entity.getZ() == eAccess.getEntity().getZ()) {
    if (eLogger != null) {
      eLogger.onEntityImpacted(entity, new Vec3d(0,-0.9923437498509884d, 0));
    }
    continue;
}
```
Removing the second check I mentioned would fix this behavior.

How to recreate:
summon tnt 0 10 0 {NoGravity:1b,Fuse:80}
summon tnt 0 10 0 {NoGravity:1b,Fuse:160}

Expected behavior: The second TNT shoots down at approximately 0.992342749851 blocks per tick.
Actual behavior: The second TNT does not move.

lntricate1:
- Moved the check for equal position into the if statement above, to only run when new entities are fetched (optimization)
- Added the condition of being onGround to that check, to fix #1081
- Renamed some variables in doExplosionA to be more readable
- Removed an unnecessary import for `java.util.Collections`

### DIFF
--- a/src/main/java/carpet/helpers/OptimizedExplosion.java
+++ b/src/main/java/carpet/helpers/OptimizedExplosion.java
@@ -5,8 +5,10 @@ import java.math.RoundingMode;
 import java.text.NumberFormat;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Set;
 
 import net.minecraft.Util;
 import net.minecraft.core.BlockPos;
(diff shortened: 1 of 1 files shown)


altrisi:
Seems to work testing it on top of 23w44a, though there's conflicts now and I have near zero familiarity with this part of the codebase.

lntricate1:
Hello, I've realized this has an edge case where the behavior is non-vanilla (if an entity moves between 2 explosions happening, this code uses the cached values instead of recalculating for the new position). I'll try to fix it within a few days, please don't merge yet!

PR #1833, opened 2023-10-16
some planned screen anvancement

ch-yx:
- [ ] get_current_screen()
- [ ] event of player opening GUI

### DIFF
--- a/src/main/java/carpet/script/value/ScreenValue.java
+++ b/src/main/java/carpet/script/value/ScreenValue.java
@@ -17,6 +17,7 @@ import java.util.Map;
 import java.util.OptionalInt;
 
 import net.minecraft.commands.CommandSourceStack;
+import net.minecraft.core.registries.BuiltInRegistries;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.nbt.ListTag;
 import net.minecraft.nbt.Tag;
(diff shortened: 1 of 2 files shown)


src/main/java/carpet/script/value/ScreenValue.java
```diff
@@ -130,6 +131,26 @@ public ScreenValue(ServerPlayer player, String type, Component name, @Nullable F
         this.inventory = new ScreenHandlerInventory(this.screenHandler);
     }
 
+    public ScreenValue(ServerPlayer player, @Nullable FunctionValue callback, Context c)
+    {
+        this.screenHandler = player.containerMenu;
+        
+        this.name = Component.literal("seems that the game forgot that");//should i make something like a weak map to remember it?
```
replaceitem, line 138: I would not use a map, since you aren't gonna know whether a screen will ever be retrieved from `get_current_screen` at all. It just creates additional overhead, that would be required even if no scripts are running. Setting it to null and returning a null value when querying name is probably the best option IMO.

replaceitem (changes requested):

src/main/java/carpet/script/value/ScreenValue.java
ch-yx: ok. i let it be null now.    should i use mixin to add a field in vanilla screen objects to store it?

PR #1834, opened 2023-10-21
Remove check for sound presence from Scarpet's sound function

### LINKED ISSUE #1832: [scarpet] `unknown_sound` error when loading a custom sound in resource pack
I added sound `minecraft:block.note_block.piano` in a custom resource pack, and the sound works ok with `/playsound` command, but I got `unknown_sound` error when playing it using `sound()` function in Carpet Script.

ch-yx:
to fix #1832


docs/scarpet/api/Auxiliary.md
```diff
@@ -14,7 +14,6 @@ and `voice`. `pos` can be either a block, triple of coords, or a list of three n
  
 Used with no arguments, return the list of available sound names.
  
-Throws `unknown_sound` if sound doesn't exist.
```
altrisi, line 17: Instead this should (in some way) mention that it won't throw if a sound isn't a vanilla one (or isn't present on the server or whatever used to trigger this), and also mention that the list of available sounds above won't include those.

src/main/java/carpet/script/api/Auxiliary.java
```diff
@@ -138,10 +138,7 @@ public static void apply(Expression expression)
             String rawString = lv.get(0).getString();
             ResourceLocation soundName = InputValidator.identifierOf(rawString);
             Vector3Argument locator = Vector3Argument.findIn(lv, 1);
-            if (cc.registry(Registries.SOUND_EVENT).get(soundName) == null)
-            {
-                throw new ThrowStatement(rawString, Throwables.UNKNOWN_SOUND);
-            }
+
             Holder<SoundEvent> soundHolder = Holder.direct(SoundEvent.createVariableRangeEvent(soundName));
```
altrisi, line 142: Wondering what does this return now, given holder stuff sounds like registry stuff and now the sound may not be in the registry. Does it create some temporary one? (mostly concerned about the intrusive holder stuff on freeze)

altrisi (commented):
Makes sense that sounds wouldn't be on the server.

Do you know if there is any way to add those to the server so they appear in the list? Or does the server just not load those (I assume the only way would be from datapacks, that idk if they can do sound).

docs/scarpet/api/Auxiliary.md
ch-yx: done

src/main/java/carpet/script/exception/Throwables.java
```diff
@@ -21,7 +21,7 @@ public class Throwables
     public static final Throwables UNKNOWN_ITEM = register("unknown_item", VALUE_EXCEPTION);
     public static final Throwables UNKNOWN_BLOCK = register("unknown_block", VALUE_EXCEPTION);
     public static final Throwables UNKNOWN_BIOME = register("unknown_biome", VALUE_EXCEPTION);
-    public static final Throwables UNKNOWN_SOUND = register("unknown_sound", VALUE_EXCEPTION);
+    //public static final Throwables UNKNOWN_SOUND = register("unknown_sound", VALUE_EXCEPTION);
```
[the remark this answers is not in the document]
ch-yx: ok

altrisi (approved):
Thanks! Sorry for the delay!

PR #1835, opened 2023-10-22
Remove `fillLimit` rule and its migration code

altrisi:
Maybe we should add some more generic code for carpet rule -> gamerule migration now that there's so many rules getting going this way.


Ghoulboy78:
That would be nice, but it may not always work, cos it would depend on how mojang actually implemented it, which may be significantly different to carpet implementation.

Issue #1841, opened 2023-11-07
[Enhancement] updates to `flippinCactus`


end-user:
The list of items `flippinCactus` can affect should be updated to include: rotate beehives, beenests, chiseled bookshelves, calibrated sculk sensor, decorated pots, and flip lightning rods like end rods.

PR #1842, opened 2023-11-07
Add new rotable blocks to `flippinCactus`

end-user:
Added convenience method and additional block comparisons to improve rotatable block options; #1841.


altrisi (approved):
Looks good and works, thanks! Sorry for the long delay!

PR #1844, opened 2023-11-08
to let it support infinity effect

ch-yx:


src/main/java/carpet/script/value/EntityValue.java
```diff
@@ -1483,7 +1483,7 @@ else if (v instanceof final ListValue lv)
                         return;
                     }
                     int duration = (int) NumericValue.asNumber(list.get(1)).getLong();
-                    if (duration <= 0)
+                    if (duration <= 0 && duration != -1)
```
gnembon, line 1486: since docs didn't specify all negative values, check only for duration == 0 to clear. Any duration < 0 should set it to infinite.

Also you should take care of the query to return -1 on infinite.

ch-yx: 🆗

gnembon (approved):

PR #1847, opened 2023-12-01
Tick fake players in network updating phase

zly2006:
In this PR, I moved fake player logic from EN to NU, so they can be ticked like real players.

There is also a new rule called `fakePlayerTicksInEU` for players who want the old fake player mechanics, it is false by default.


Fallen-Breath:
See this discussion on scicraft discord in 2020, for gnembon's opinion on the player action pack timing: https://discord.com/channels/211786369951989762/573613501164159016/771405054632394782

> for that to work properly would require a lot of chances and code - essentially attaching client code to the server part. Right now fake players don't differ that much from pigs. Is it bad that they behave like other server side entities?

zly2006:
> Also, "EN" and "NU" are not commonly used terms in the non-Chinese community. I would suggest using more precise words to describe them

Could you provide more suggestions? what word should i use for tick phases?

> for that to work properly would require a lot of chances and code - essentially attaching client code to the server part. Right now fake players don't differ that much from pigs. Is it bad that they behave like other server side entities?

In fact, in most situations we just expect fake players to be ticked and interact with other things in the game (actions) like real players, so that most farms can work both for real and fake players such as raid farms.

Btw, there is an option to configure--if you want pig-like fake players you can just turn it on.

src/main/java/carpet/mixins/ServerConnectionListener_fakePlayersMixin.java
```diff
@@ -0,0 +1,38 @@
...
+        if (!CarpetSettings.fakePlayerTicksInEU) {
+            for (ServerGamePacketListenerImpl connection : FakePlayerManager.connections) {
+                // from ServerGamePacketListenerImpl#tick
+
+                connection.player.xo = connection.player.getX();
+                connection.player.yo = connection.player.getY();
+                connection.player.zo = connection.player.getZ();
+
+                // action packet update
+                ((ServerPlayerInterface) connection.player).getActionPack().onUpdate();
```
Fallen-Breath, line 29: Real player actions (from c2s packet) do not tick in this phase

src/main/java/carpet/mixins/ServerConnectionListener_fakePlayersMixin.java
```diff
@@ -0,0 +1,38 @@
...
+
+                connection.player.doTick();
+                //  connection.player.absMoveTo(connection.firstGoodX, connection.firstGoodY, connection.firstGoodZ, connection.player.getYRot(), connection.player.getXRot());
+
+                // todo: vehicle?
```
Fallen-Breath, line 34: I'll suggest to mark the PR as a draft, if it's not completed yet

src/main/java/carpet/mixins/PlayerList_fakePlayersMixin.java
```diff
@@ -41,7 +42,9 @@ private ServerGamePacketListenerImpl replaceNetworkHandler(MinecraftServer serve
     {
         if (playerIn instanceof EntityPlayerMPFake fake)
         {
-            return new NetHandlerPlayServerFake(this.server, clientConnection, fake, cookie);
+            var handler = new NetHandlerPlayServerFake(this.server, clientConnection, fake, cookie);
+            FakePlayerManager.connections.add(handler);
```
Fallen-Breath, line 46: where's the corresponding code for removing the handler? Adding without deleting might cause a slow memory leak in servers that run for a long time.

zly2006: https://github.com/gnembon/fabric-carpet/pull/1847/files/303e099b5885fd4c77285e0c4ea87a41a45ca50b#diff-c59c3e90cb5d221f2a8191ae3e51d58741a8b3a4c48d48773859d4fa7b4f25d8R11

Fallen-Breath: handlers from removed fake players will also be ticked here, which does not seem correct https://github.com/gnembon/fabric-carpet/pull/1847/files#diff-b8fad85a8571843b58c03460fb7bf66949628f5a369d0ae14d2b97834199e0adR21

src/main/java/carpet/patches/FakePlayerManager.java
```diff
@@ -0,0 +1,14 @@
+package carpet.patches;
+
+import net.minecraft.server.network.ServerGamePacketListenerImpl;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class FakePlayerManager {
+    public static List<ServerGamePacketListenerImpl> connections = new ArrayList<>();
```
Fallen-Breath, line 9: Suggestion: use concurrency-proof containers to handle potential concurrent access in certain situations (e.g. using mods)

src/main/java/carpet/mixins/PlayerList_fakePlayersMixin.java
Fallen-Breath: > https://github.com/gnembon/fabric-carpet/pull/1847/files/303e099b5885fd4c77285e0c4ea87a41a45ca50b#diff-c59c3e90cb5d221f2a8191ae3e51d58741a8b3a4c48d48773859d4fa7b4f25d8R11

that's not a correct implentation since `onServerClosed` only gets invoked once per server lifecycle

Fallen-Breath:
> Could you provide more suggestions? what word should i use for tick phases?

- EU: entity phase
- Network (when `ServerConnectionListener#tick` gets invoked): just network phase
- NU (when c2s packets are handled, and async tasks are executed): player phase, or async task phase

> In fact, in most situations we just expect fake players to be ticked and interact with other things in the game (actions) like real players, so that most farms can work both for real and fake players such as raid farms.

that's gnembon's opinion, I'll suggest to ask him directly

> Btw, there is an option to configure--if you want pig-like fake players you can just turn it on.

See https://github.com/gnembon/fabric-carpet/pull/1847#discussion_r1412217146

zly2006:
> Is there any downside to _only_ ticking them in the right/more correct phase? If there isn't I'd remove the old behaviour completely.

Some contraptions may be breoken after we changed this, so I think it is a right choice not to cchange the existing behavior and add an option to tick them in the correct phases.

altrisi:
How many contraptions depend on fake players working differently than real ones? (outside of ones built specifically to detect them). I'd expect fake players would be used to test things where real players would be the actual target, so having them work for real-like players would be the idea (that's been the idea of fake players from the start, just wasn't completely accurate).

Having two code paths seems redundant when one is clearly "better" (more accurate). We'd also release this on a snapshot/major (mojang versioning wise) version anyway and we don't usually backport, so the chances of breaking being unnoticed are lower.

zly2006:
> How many contraptions depend on fake players working differently than real ones? (outside of ones built specifically to detect them).

afaik, some raid farms are specially designed for fake players for the tick phase order, this makes them 1gt faster then real players

> Having two code paths seems redundant when one is clearly "better" (more accurate). We'd also release this on a snapshot/major (mojang versioning wise) version anyway and we don't usually backport, so the chances of breaking being unnoticed are lower.

Yes, it is a breaking change so I just suggest we shoulld be more careful, I totally agree that a contraption that real players cannot use it not "useful".

PR #1849, opened 2023-12-19
Fix Race Condition Crash Exploit

senseiwells:
I was getting some off-thread warnings in another mod I'm working on when spawning fake players. 
This lead me to try spamming `/player foobar spawn` in chat, which crashed the game.

I have just changed the `thenAccept` into a `thenAcceptAsync` on the main server thread. This now creates the player and spawns them on the main thread, so there are no more possible race conditions that can happen.


altrisi (approved):
Makes sense, thanks!

PR #1854, opened 2024-01-05
idea: let draw_shape can use clip coordinate, so that it can be used to create HUD easilier.

ch-yx:
https://discord.com/channels/882822986795716608/894946828980723762/1192811784445374525

this is to add a optional flag, 
with this, you can draw some thing as usual, in [-1,1]^3, and add this optional flag, it will be drawn before users sight. 
which is helpful to create some HUD like things. for example, to display some info on users screen.

Uploading screenhud[1].mp4…

### DIFF
--- a/src/main/java/carpet/script/utils/ShapesRenderer.java
+++ b/src/main/java/carpet/script/utils/ShapesRenderer.java
@@ -159,7 +159,7 @@ public class ShapesRenderer
             // lines
             RenderSystem.lineWidth(0.5F);
             shapes.get(dimensionType).values().forEach(s -> {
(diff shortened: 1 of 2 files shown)


src/main/java/carpet/script/utils/ShapesRenderer.java
```diff
@@ -53,10 +55,13 @@
 
 public class ShapesRenderer
 {
+    private static final Matrix4f ORTHOMAT = new Matrix4f().m22(-1);
```
ch-yx, line 58: not sure if this line of code is written correctly

gnembon:
this looks very useful. Is that still WIP?

PR #1864, opened 2024-01-19
Permissions API support concept 2

### LINKED ISSUE #1850: Support fabric permissions apis for better customizing carpet commands
Currently, carpet commands's usage limit is hard-encoded in carpet config, adding support to lucko/fabric-permissions-api will allow more accurate access control.
This may be optional or disabled by default, but I do think it would be helpful to someone like me.

altrisi:
Prototype on supporting the [fabric permissions API](https://github.com/lucko/fabric-permissions-api) for Carpet commands.
It'd only be used if the permission manager is present, so it's not required/included otherwise.

TBD:
- Is this the way to do it? Or allow choosing the node in the rule/app? ([like this old prototype](https://github.com/altrisi/fabric-carpet/commit/693f37619bff1987eaf962a550da5f85ca7bffb2)) The other way is more intrusive IMO, it changes behaviour of a lot of things depending on whether another mod is present
- Should we disable/hide the command rules if a permission api is present? It could make sense: if it's there, who installed Carpet will probably want to use permissions for commands. We could use a simple `@Rule.Condition` for that
- (if no to the above) Is it fine to put the hint message in the general command validator? Extensions may not support permissions yet (unlike the above method, this doesn't just automatically work for them, they have to call a different method)

TODO:
- Hook it with Scarpet apps, their logic is more complex so I didn't hook it yet for the concept
- Check that we're doing fallbacks correctly, we may need to pass a fallback to the api.
- Add actual api call
- Ensure the api can't be present without a permission handler mod, or see if we can check that to disable this system in that case

Maybe future enhancements:
- Check if there's a way we can hook to give a different permissions group/whatever it's called to scarpet `run`, so owners can configure what apps can run
- Maybe add subcommand specific permissions for scarpet apps? Maybe for a followup, it'd also be nice to have without perms

Note that I don't know a lot about the permissions api stuff, barely used it a similar one many years ago.

Fixes #1850.

### DIFF
--- a/src/main/java/carpet/api/settings/SettingsManager.java
+++ b/src/main/java/carpet/api/settings/SettingsManager.java
@@ -602,7 +602,7 @@ public class SettingsManager {
         }
 
         LiteralArgumentBuilder<CommandSourceStack> literalargumentbuilder = literal(identifier).requires((player) ->
-                CommandHelper.canUseCommand(player, CarpetSettings.carpetCommandPermissionLevel) && !locked());
+                CommandHelper.hasPermission(player, identifier + ".commands.settings", CarpetSettings.carpetCommandPermissionLevel) && !locked());
 
         literalargumentbuilder.executes((context)-> listAllSettings(context.getSource())).
                 then(literal("list").


sakura-ryoko:
Hey, nice work.  I just wanted to mention that there are additional versions of fabric-permissions-api that are fairly common to use.  Take a look at implementing a check for Drex's fork, as he is implementing it in a bunch of mods that he contributes code for.
https://github.com/DrexHD/fabric-permissions-api

altrisi:
As far as I know that 0.3 is compatible with mods depending on the original, so given Carpet wouldn't actually include the API (just use it if it's there), the idea is it'd be compatible with either out of the box.

Issue #1882, opened 2024-03-11
Carpet fake players in survival mode have 4.5 block reach when attacking mobs


TheobaldTheBird:
Carpet bots have significantly higher reach than real players when attacking entities. Normal player attack reach is 3 blocks. This happens because EntityPlayerActionPack.getTarget() uses the same reach value for block interaction and entity interaction

Issue #1881, opened 2024-03-11
Carpet fake players don't slow down while using shields, consumables, bows, etc.


TheobaldTheBird:
Carpet fake players don't slow down realistically when using shields, consumables, brushes, or other items that slow players down

PR #1883, opened 2024-03-11
Fix fake players using wrong reach values

TheobaldTheBird:
The bots still have a 4.5 block range while interacting with blocks, but entity interaction range has been reduced to 3 blocks, fixing #1882

Player attacking is more realistic now. Using /player <player> attack once will swing even while not targeting an entity, and will properly reset attack cooldown

#1881 is also fixed, as players will properly lower their movement speed while using slow items

### DIFF
--- a/src/main/java/carpet/helpers/EntityPlayerActionPack.java
+++ b/src/main/java/carpet/helpers/EntityPlayerActionPack.java
@@ -6,6 +6,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import carpet.logging.Logger;
 import carpet.patches.EntityPlayerMPFake;
 import carpet.script.utils.Tracer;
 import net.minecraft.commands.arguments.EntityAnchorArgument;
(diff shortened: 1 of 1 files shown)


src/main/java/carpet/helpers/EntityPlayerActionPack.java
```diff
@@ -238,7 +250,16 @@ public void onUpdate()
                 }
             }
         }
-        float vel = sneaking?0.3F:1.0F;
+
+        if (player.isUsingItem())
+        {
+            ItemStack item = player.getUseItem();
+            ((ServerPlayerInterface) player).getActionPack().setSlowed(item.isEdible() || item.is(Items.SHIELD) || item.is(Items.BRUSH) || item.is(Items.BOW) || item.is(Items.CROSSBOW) || item.is(Items.TRIDENT) || item.is(Items.GOAT_HORN) || item.is(Items.SPYGLASS));
```
altrisi, line 257: Isn't this equivalent to just `isUsingItem`, or am I missing any?

Also given nothing else is using `slowed`, it could be a local.

Haven't checked more of the code yet, sorry.

TheobaldTheBird: Yep, looks like `player.isUsingItem()` is only true while using any of those listed items, most of that was completely unneeded. I'll commit a cleaned up version

PR #1890, opened 2024-03-24
Fixed a bug in the ai_tracker script which caused the golem detection range to be incorrect.

Math2128:
Changed height to villager_height in the ai_tracker.sc file which caused the golem detection range to be 2 blocks smaller towards the positive Z.

Consist of one line of code, but I believe this should resolve the issue.


Math2128:
### Before the fix:



### After the fix:




Notice the white line in the background which now correspond to the 16 blocks detection zone (instead of 14 before)

PR #1896, opened 2024-04-14
Increase Empty Shulker Boxes Stack Size for new ItemStack tolerances. (Minor change request)

sakura-ryoko:
Allow shulker boxes to stack up to 99 with a semi-hidden value for the new "max_stack_size" Component value under ItemStack allowed rules, but the default remains at 64 for (true).


sakura-ryoko:
I changed my mind, I think this is a terrible idea.

PR #1900, opened 2024-04-24
Cleanup and improve

extclp:


altrisi:
Many of those here are just a different style, and some are simply wrong (replacing Collectors.toList() with Stream.toList(), the ones where that could be done I already did, Stream.toList() is immutable, and others).

`getLast` is nice, tho I don't think `getFirst` is necessarily that much better. Much noise for little reward, actually longer.

I may take a closer look when I have some more time and take the parts that make sense.

PR #1919, opened 2024-05-31
Update to 1.21-pre2

axialeaa:
Requires changing `LivingEntity_creativeFlyMixin#canChangeDimensions` target, as `canChangeDimensions()` is no longer overridden in `LivingEntity.class`. I have changed it to target `canUsePortal()` instead which has a different purpose to `canChangeDimensions()`, so please let me know if this is not a good fix.


voidpointer0x00 (approved):
The original method was just renamed (implementations match), can confirm that `creativeNoClip` still works

PR #1937, opened 2024-06-23
fix incompatibility with lithium explosion exposure optimization

2No2Name:
Lithium redirects the same call to replace the computation with a cached value lookup. 

I don't see a way to avoid lithium's redirect, without getting rid of the optimization, which is why I am suggesting to change it in carpet.

This PR uses the `@Local` instead of CAPTURE_FAILHARD to get the local variable, because it is easier.

I don't know how to use scarpet, so I couldn't test it, but it at least didn't crash the game when I exploded a tnt in a flat world

### DIFF
--- a/src/main/java/carpet/mixins/Explosion_scarpetEventMixin.java
+++ b/src/main/java/carpet/mixins/Explosion_scarpetEventMixin.java
@@ -1,5 +1,6 @@
 package carpet.mixins;
 
+import com.llamalad7.mixinextras.sugar.Local;
 import it.unimi.dsi.fastutil.objects.ObjectArrayList;
 import net.minecraft.core.Holder;
 import net.minecraft.core.particles.ParticleOptions;
@@ -9,7 +10,6 @@ import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
-import org.spongepowered.asm.mixin.injection.Redirect;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
 import java.util.ArrayList;
@@ -22,7 +22,6 @@ import net.minecraft.world.entity.LivingEntity;
 import net.minecraft.world.level.Explosion;
 import net.minecraft.world.level.ExplosionDamageCalculator;
 import net.minecraft.world.level.Level;
-import net.minecraft.world.phys.Vec3;
 
 import static carpet.script.CarpetEventServer.Event.EXPLOSION_OUTCOME;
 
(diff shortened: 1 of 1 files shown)


altrisi:
You can test events by loading the built-in `event_test` (`/script load event_test`), which should print the info received by them.

I'll try to take a look and merge later.

manuelgrabowski:
Any chance this could get merged? The current latest versions of Lithium and Carpet are now crashing when used together.

djmrFunnyMan:
Can this be merged ASAP?

tajemniktv:
This really needs to be merged on Carpet side. Both mods are essential and crucial for many players, and further optimization of explosions is always a good thing to see. 
If anything breaks with that, we will report immediately, don't worry lmao

src/main/java/carpet/mixins/Explosion_scarpetEventMixin.java
```diff
@@ -1,5 +1,6 @@
 package carpet.mixins;
 
+import com.llamalad7.mixinextras.sugar.Local;
```
supersaiyansubtlety, line 3:
```suggestion
import org.spongepowered.asm.mixin.injection.ModifyArg;
```

src/main/java/carpet/mixins/Explosion_scarpetEventMixin.java
```diff
@@ -55,17 +52,16 @@ private void onExplosionCreated(Level world, Entity entity, DamageSource damageS
         }
     }
 
-    @Redirect(method = "explode", at=@At(
+    @Inject(method = "explode", at=@At(
```
supersaiyansubtlety, line 55:
```suggestion
    @ModifyArg(method = "explode", at=@At(
```

src/main/java/carpet/mixins/Explosion_scarpetEventMixin.java
```diff
@@ -55,17 +52,16 @@ private void onExplosionCreated(Level world, Entity entity, DamageSource damageS
...
             value = "INVOKE",
             target = "Lnet/minecraft/world/level/Explosion;getSeenPercent(Lnet/minecraft/world/phys/Vec3;Lnet/minecraft/world/entity/Entity;)F")
     )
-    private float onExplosion(Vec3 source, Entity entity)
+    private void onExplosion(CallbackInfo ci, @Local Entity entity)
```
supersaiyansubtlety, line 59:
```suggestion
    private Entity onExplosion(Entity entity)
```

src/main/java/carpet/mixins/Explosion_scarpetEventMixin.java
```diff
@@ -55,17 +52,16 @@ private void onExplosionCreated(Level world, Entity entity, DamageSource damageS
...
     {
         if (affectedEntities != null)
         {
             affectedEntities.add(entity);
         }
-        return getSeenPercent(source, entity);
     }
```
supersaiyansubtlety, line 65:
```suggestion
        return entity;
    }
```

supersaiyansubtlety (commented):
`@ModifyArg` would be less brittle than local capture

sternschnaube:
Thank you @Ka1nnn 😄

MeeniMc:
You can use the compiled artifact for the PR instead of relying on builds from unknown sources 
https://github.com/gnembon/fabric-carpet/actions/runs/10369618488

MeeniMc:
Event script reports events normally, (with and without the optimizeTNT carpet rule active). I ran a set of farms from scicraft Blitz world and they all worked as intended (with and without optimizeTNT active), blast chambers, world eater etc..

gnembon:
Sorry, was busy.. explosion code has changed drastically. Is that still an issue?

2No2Name:
I worked around this on lithium's side

djmrFunnyMan:
But wouldn't it be better if carpet implements the fix so you doj't need to dp a workaround?

WinExp:
When is this pr going to be merged, my don't want to have to modify and compile the pr branch myself anymore : (

PR #1996, opened 2024-09-12
Enhance Value#of Method for Improved Item Data Retrieval (Components compatibility)

Gnottero:
### Overview:
This pull request refactors the `Value#of` method in the `ValueConversions` class to introduce a new item-tuple format. The key change is the inclusion of an encoded third element (via `NBTSerializationValue`), which now encapsulates not only modified components of the `ItemStack` but also the default item components.

### Reason:
This modification allows users to retrieve previously inaccessible data, such as the mining speed of a pickaxe when mining specific blocks, enhancing the functionality of the system.
**Key Changes:**

-     The third element in the tuple now contains an encoded value, providing both modified and default item components.
-     The new structure improves data accessibility for complex item behaviors.

### Request for Review:
Please review the code, paying particular attention to the final section of the `getComponentsFromItemStack` method. I would appreciate any feedback on how to further improve code quality, especially regarding performance, readability, or logic in that portion.


src/main/java/carpet/script/value/ValueConversions.java
```diff
@@ -499,9 +505,34 @@ public static ItemStack getItemStackFromValue(Value value, boolean withCount, Re
...
+                                            .encodeStart(
+                                                    regs.createSerializationContext(NbtOps.INSTANCE),
+                                                    cvalue)
+                                            .result().orElse(null);
+                                    yield NBTSerializableValue.of(res);
+                                }
+                            };
+                        }));
+
+        MapValue mapValue = MapValue.wrap(Map.of(StringValue.of("components"), MapValue.wrap(lst), StringValue.of("id"), StringValue.of(stack.getItem().toString()), StringValue.of("count"), NumericValue.of(stack.getCount())));
```
ch-yx, line 532: if you pack all of those k-v pairs to a NBT tag again, why you left the switch there?
you can let Number&Boolean use the default case too.


and you no longer need to use a NBTSerializableValue to warp the `res` tag.

src/main/java/carpet/script/api/Inventories.java
```diff
@@ -19,14 +27,6 @@
...
 import carpet.script.value.ValueConversions;
-
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.OptionalInt;
-import java.util.Set;
-
 import net.minecraft.core.HolderSet;
```
ch-yx, line 30: why you edit this file? i cannot see the reason behind it. 
you dont need to commit all of your edited files

Gnottero: Those file were added by mistake. I'm going to remove them

src/main/java/carpet/script/value/ValueConversions.java
Gnottero: I've changed that to your implementation, removing the part that took the item from the main hand in case of no values passed to the function. I also removed unedited files. It should be fine now, I guess

ch-yx: then you please edit your Overview, Reason and Request for Review......

BisUmTo (approved):

PR #2036, opened 2025-02-11
Fixes bugged `run(expr)` (#2023) by adding `run_cb(expr, callback)` and `run_wait(expr)`

### LINKED ISSUE #2023: Bug: Scarpet `run('command')` function does not return any values
## Description
The `run('command')` function never (except in a syntax exception error) returns the command's values, and always returns `[0, [], null]`

### Test 1: `/script run run('execute if entity @s')`
**Expected output:** `[1, [Test passed, count: 1], null]`
image
*(Ran with minecraft 1.20.1, fabric 0.16.9, carpet v1.4.112+v230608)*

**Observed output:** `[0. [], null]`
image
*(Ran with minecraft 1.21.4, fabric 0.16.9, carpet v1.4.161+v241203)*

### Test 2: `/script run run('execute unless entity @s')`
**Expected output:** `[0, [], Test failed, count: 1]`
image
*(Ran with minecraft 1.20.1, fabric 0.16.9, carpet v1.4.112+v230608)*

**Observed output:** `[0. [], null]`
image
*(Ran with minecraft 1.21.4, fabric 0.16.9, carpet v1.4.161+v241203)*

## Guessed Source

Thank you to @dmunozv04 for finding the source [here](https://github.com/gnembon/fabric-carpet/blob/master/src/main/java/carpet/script/api/Auxiliary.java#L688);

We believe that due to the new way commands are handled, the current implementation of the `run('command')` function returns too early before the command is ever executed, provided it isn't a syntax error, which does give the correct output (see below)
`/script run run ('exec')`
image

Attached is an image of the implementation with an additional oversight highlighted, where the command success value is always set to 0, regardless of the actual command execution
image

Oxtaly:
This PR fixes #2023 by adding `run_cb(expr, callback)` and `run_wait(expr)` functions that cover the previous return value use cases for `run(expr)`
More details and information in the committed Auxiliary.md

This also adds an error callback along with a setter `setErrorCallback` in SnoopyCommandSource to work with errors from the CommandSourceStack sendFailure being called after the resultConsumer from `withCallback` or being called too early with resultConsumer not being triggered

I've tried to follow the code practices and indentation of the project but I'm still new to contributing to open source so apologies if it doesn't match perfectly

### DIFF
--- a/docs/scarpet/api/Auxiliary.md
+++ b/docs/scarpet/api/Auxiliary.md
@@ -406,25 +406,42 @@ read_file('foo', 'shared_text')     => ['one', 'two', 'three', '', 'four', '', '
   
 ### `run(expr)`
 
-Runs a vanilla command from the string result of the `expr` and returns a triple of 0 (unused after success count removal), 
-intercepted list of output messages, and error message if the command resulted in a failure. 
-Successful commands return `null` as their error.
+Runs a vanilla command from the string result of the `expr` and returns a triple of the exit code (usually 0 or 1), 
+intercepted list of output messages, and last synchronous error message if the command resulted in a failure.
+The exit code may be null if a non command caused error occurred while running the command, 
+with the error message field giving the error message.
 
 The command return `null` if the command was not run immediately, but was scheduled for later execution 
-due to being requested while command was executed. This happens most commonly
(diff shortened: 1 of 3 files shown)


gnembon:
I explained what the issue is about. I think there are mitigation measures. Maybe there is still a place for `run_async` with callbacks?

PR #2079, opened 2025-06-21
Fix: Compatibility with polymers moving element holders

tomalbrc:
The current implementation was breaking polymers block attachments when the blocks were moved by pistons.
This change should fix the issue


altrisi:
What's a good way for me to test this?

PR #2148, opened 2026-01-02
endPortalItemVelocityRotationFix

Nokorot:
This is feature for fixing the rotation of the velocity of items thrown though an end portal.
In MC versions  < 1.21, items thrown through an end portal preserved their momentum. This changes in 1.21.

In the SMP I am hosting, we are relying on precise throws though the portal for sorting. I couldn't find any other fixes for this, so I implemented a fix on my own.

The fix I wrote for 1.21.4 is mostly the same, except some of the names has changed.

### DIFF
--- a/src/main/java/carpet/CarpetSettings.java
+++ b/src/main/java/carpet/CarpetSettings.java
@@ -971,6 +971,13 @@ public class CarpetSettings
     public static boolean lightningKillsDropsFix = false;
 
     @Rule(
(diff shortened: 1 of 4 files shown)


Crystal0404:
1. Debug output should not be included in the commit.
2. Carpet is currently only fixing existing bugs, meaning this PR is unlikely to be merged. This feature would be more suitable for your own carpet extension.

PR #2176, opened 2026-03-25
Fixing isLocalInstanceAuthoritative and getKnownMovement for EntityPlayerMPFakes

### LINKED ISSUE #2173: version 1.21.11,after i coded the mace bot it doesnt trigger the mace attack even though i use the carpet mace fix,the bot CANT mace even after i coded it
i coded the mace bot using carpet,it can do stuffs,but when it hits a person,it doesnt work,even when i use carpet mod fix

### LINKED ISSUE #2198: Carpet bots do not take fall damage in 26.1
Carpet bots do not take fall damage in 26.1. To reproduce, spawn a bot in survival mode. Then drop it from a height at which it would take fall damage. Instead of taking damage as intended, no damage is taken.

HerobaneNair:
isLocalInstanceAuthoritative only checked if the passenger controlling an entity was a fake player and marked the ridden entity to be not the authoritative instance, instead of also marking regular fake players. This means it would use the fake player's client side packets for fall damage checks - of which none were sending.

[Maybe there's a better/cleaner way to implement this] getKnownMovement is used for spears, and was always 0 since it uses client side movement packets to work. Even with the localInstanceAuthoritative being false, it wasn't updating - so it's being forced to update correctly.


altrisi:
Super late, but updated it to work on the latest version and did some tweaks.

This also fixes a recent bug where fall damage isn't applied to fake players, which is nice.

Thanks a lot again!

PR #2180, opened 2026-04-02
fix: drowned spawns not spawning in Ocean biomes (fixes #2147)

### LINKED ISSUE #2147: Drowned fail to spawn in Oceans due to the `doesNotCollide` Mixin
https://github.com/gnembon/fabric-carpet/blob/0595f0ced79801a50cd81517b018b25ebb101f59/src/main/java/carpet/mixins/NaturalSpawnerMixin.java#L60

I am not sure why but even with `CarpetSettings.lagFreeSpawning` disabled; this Mixin's existence still breaks Drowned from spawning in Warm, etc Ocean Biomes.

Testing Procedure:
- Write a Test Mod without FAPI with a `Drowned.class` Mixin that uses `@Overwrite` / `@Redirect` for `checkDrownedSpawnRules` to verify that the spawning conditions are correct.
(When `randomSource.nextInt(40) == 0`).  Write Test results to a Test Logger when Results are TRUE that shows the `BlockPos` tested where the Spawn should Occur.
- Install CarpetMod.  Do not enable `CarpetSettings.lagFreeSpawning`.
- Load up a random world, and travel to any Ocean & Move around until the Mixin Reports the spawning location of a Drowned.
- Tick Freeze, and go to the Reported BlockPos, and Verify if a Drowned successfully spawned.

Recommended fix:
- Remove or Refactor the `doesNotCollide` Optimization.

Tested a Custom Build of Carpet with said `@Redirect` commented out and Observed Drowned spawning correctly.

Video Proof:  Watch Xisuma's Coral Nautilus farm-concept stream while CarpetMod is installed; and it spawning Zero Drowned.

sakura-ryoko:
fixes Issue #2147.

Replaced Mixin with `@WrapOperation`.  I recommend making much deeper usage of this type of Mixin, especially with the lag free spawning Mixin.  I also tested with and without `lagFreeSpawning` enabled.

`@WrapOperation` / `@WrapMethod`, etc can generally be used with `original.call()` as the default, to effectively replace any usage of `@Redirect`, basically; especially when using Redirect potentially causes problems.

Proof of functionality:

Jockey-fix

### DIFF
--- a/src/main/java/carpet/mixins/NaturalSpawnerMixin.java
+++ b/src/main/java/carpet/mixins/NaturalSpawnerMixin.java
@@ -4,6 +4,8 @@ import carpet.CarpetSettings;
 import carpet.fakes.LevelInterface;
 import carpet.utils.SpawnReporter;
 import net.minecraft.world.entity.EntitySpawnReason;
+import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
+import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 import org.apache.commons.lang3.tuple.Pair;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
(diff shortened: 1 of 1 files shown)


sakura-ryoko:
[TESTMIXIN.java](https://github.com/user-attachments/files/26727189/TESTMIXIN.java)

TEST MIXIN used if you want to setup the test yourself.

Recommended finding a large Warm Ocean Biome in a random seed using Chunk Base or whatever.

altrisi:
Long time since this, I recall I could kind-of reproduce it but wasn't sure if I was doing it right, seemed to fix something like what the issue described in some of the runs. Weird thing is iirc I tried something similar at some point and it didn't work.

However, regardless of whether I can verify it, at the very least it's a definite improvement to the code by wrapping instead of redirecting so I'll run it on latest master and merge because this has already stalled long enough.

PR #2202, opened 2026-06-19
Adapt for Minecraft 26.1.2

FengguaYa:
- Update minecraft_version and loom_version in gradle.properties
- Upgrade Gradle wrapper to version 9.4 in gradle-wrapper.properties


altrisi:
The 26.1 version works as-is in 26.1.2.

PR #2214, opened 2026-07-26
Fix typo

Bishops-exe:


Bishops-exe:
@gnembon

replaceitem:
This is intended.
See https://github.com/gnembon/fabric-carpet/pull/1500

PR #2221, opened 2026-08-15
Fix the issue where Shadow Player data is not properly saved and loaded due to a race condition

### LINKED ISSUE #2145: /player resets Player Data if you spawn an on the server playing player
Hello,

I’m not sure whether this is intended behavior or a bug, but when a player who has previously played on the server is spawned using /player <name> spawn or /player <name> shadow, their entire inventory, ender chest, etc. are cleared the next time they log in. This seems to happen because the player data file for that player gets reset.

The strange thing is that this has already happened to my account twice, while other players do not seem to be affected. When other players are spawned using the /player command, their inventory is not cleared afterward.

I don’t know if I did something wrong or if this behavior is intentional. I am aware that you can spawn players who have never played on the server (e.g. Alex or Steve) if you just want AFK bots. However, it is very annoying to lose all player progress just because someone spawned your account to be AFK.

### LINKED ISSUE #2166: "/player shadow" can be used to duplicate items or lose items
Versions:
carpet 1.4.194+v251223
Minecraft 1.21.11
Fabric Loader 0.18.4
fabric-api 0.141.3+1.21.11

This is how to reproduce one version of this that accomplishes a dupe:

Log out normally with items on you for server to save your NBT
Log in
Dump items into a container
Run player shadow (this will kick your client)
Log back in

Items will be both on your character and in the container.

You can vary/skip steps above to lose items.

I have replicated this with a new server with only the mods listed under Versions above and created an unlisted video showing dupe: https://youtu.be/sMjSpHP0sx0

I reported this in another thread as a comment but figured it probably should be a separate issue.

I would speculate that logging out saves my player NBT then when logging into a shadowed player you it reloads your player NBT from disk. At no point does it update the NBT from using the shadow command. Other than autosave hitting when you are logged out with no items this seems to work always work.

### LINKED ISSUE #2191: Player data is incorrect(26.2-snapshot-7)
When using the shadow command in /player, it appears that when the player logs in again, their data seems to be rolled back to a short time before the shadow command was used.

### LINKED ISSUE #2216: player shadow dismounts player from minecart from second time onwards
Steps to reproduce:

- Join server
- Ride minecart
-  /player shadow yourself
- rejoin
-  /player shadow yourself again
- rejoin
- You will be dismounted from the minecart

Further notes:
- Every further shadow will result in you being dismounted from the minecart
- Killing and replacing the cart does not stop this
- Restarting the server does not stop this

Crystal0404:
## Reason

When `player.connection.disconnect()` is called, the player does not leave the world immediately; instead, the disconnection is handled asynchronously on another thread at a later time. Consequently, immediately spawning a fake player causes the game to load incorrect player data and leads to conflicts with the still-existing player.

## Fix
I added several interfaces to `ServerPlayer` via `ClassTweaker`. When the command `player xxx shadow` is executed, the `shouldShadow`flag is set to true. Then, through a Mixin, the fake player is created only after the real player has truly left the game. At this point, the game has already completed saving the real player's data, so the fake player's data can be loaded correctly

fix #2216, fix #2145, fix #2166, fix #2191

This is another PR that attempts to address these issues, but it does not handle the race condition between the real player and the fake player. https://github.com/gnembon/fabric-carpet/pull/2183


Crystal0404:
I've made a patch mod for the release version. If anyone's interested, you can grab it from the table below. Requires Fabric Loader 0.18.6 or newer

| Carpet Version      | Patch Version                                                                                                                                   |
|---------------------|-------------------------------------------------------------------------------------------------------------------------------------------------|
| `>=26.1`            | [0.2.1+carpet26.3](https://github.com/Crystal0404/CarpetFakePlayerPatch/releases/download/0.2.1/CarpetFakePlayerPatch-v0.2.1+carpet26.3.jar)    |
| `1.4.188` `1.4.193` | [0.2.1+carpet1.4.193](https://github.com/Crystal0404/CarpetFakePlayerPatch/releases/download/0.2.1/CarpetFakePlayerPatch-v0.2.1+carpet26.3.jar) |

https://github.com/Crystal0404/CarpetFakePlayerPatch

